PR #430, opened 2024-06-26
Edit Client Profile DEV-356

Davit-BetterAngels:
DEV-356

edit client profile. 

This should not be merged until add client is resolved. this branch is behind https://github.com/BetterAngelsLA/monorepo/pull/395

refactored to combined Add and Edit Profile


mikefeldberg (commented):
Is this the best approach to implement an edit screen? Could we modify the existing AddClient components to be used for both adding and editing?

mikefeldberg:
something like this https://github.com/BetterAngelsLA/monorepo/pull/451

Davit-BetterAngels:
> Is this the best approach to implement an edit screen? Could we modify the existing AddClient components to be used for both adding and editing?

take a look. refactored everything

mikefeldberg (changes requested):
It looks like this branch was checked out from `main` before we reset githistory, and then the "new" `main` was merged into this branch. This PR should be closed and reopened from a new branch with just the affected files. 

image

mikefeldberg:
replaced by #454 454

PR #446, opened 2024-07-02
Update development READMEs and make test styles more consistent

lena:
Went through all of the original onboarding instructions since I had to blow away the repo to update the git history. 
Made changes where ever instructions deviated from actual process.

Also updated tests formats to keep class names and assertion syntax consistent across files.

### DIFF
--- a/docs/frontend_readme.md
+++ b/docs/frontend_readme.md
@@ -33,21 +33,23 @@ Run the following on the host machine (not in the container):
    brew install yarn
    ```
 
-1. Follow the instructions in the terminal to configure NVM in your shell profile. It should look similar to:
+1. Follow the instructions in the terminal to configure NVM in your shell profile. If the instructions don't appear, run `brew info nvm`.
 
-   e.g. `~/.profile` or `~/.zshrc`:
+   The instructions will look similar to:
 
-   ```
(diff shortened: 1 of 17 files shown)


docs/frontend_readme.md
```diff
@@ -35,7 +43,7 @@ The betterangels_frontend is built in [React Native](https://reactnative.dev/).
    [ -s "/usr/local/opt/nvm/etc/bash_completion.d/nvm" ] && \. "/usr/local/opt/nvm/etc/bash_completion.d/nvm"
    ```
 
-   If these instructions don't appear, run the following:
+   If these instructions don't appear, run the following: (what do you mean @paul?)
```
lena, line 46: @vecchp what does this line mean?

vecchp: I can't say I wrote this line (though git history may say so since we just rewrote it completely). I believe it originally came from @chris920.

From what I can tell by reading it, the intention is as follows:
* When installing nvm for the first time with brew, certain lines are emitted into the terminal. If these lines don't appear, the user should type brew info nvm to retrieve them.

lena: aw ok thanks :) hi @chris920!!!

PR #447, opened 2024-07-02
Show new client after adding client (DEV-504)

davelindqvist:
[DEV-504](https://betterangels.atlassian.net/browse/DEV-504?atlOrigin=eyJpIjoiYmMxZmQyOTE4MDc2NGQyNWFlMWNiNDFjNTg5MTM1YjYiLCJwIjoiaiJ9)

I tested out:
- [fetchPolicy / nextFetchPolicy](https://github.com/BetterAngelsLA/monorepo/blob/b628516863dae7541d0119dbde7505e55a8d9616/libs/expo/betterangels/src/lib/screens/Clients/index.tsx#L53-L54)
- [refetch](https://github.com/BetterAngelsLA/monorepo/commit/da51462786eb139dc1ccaa9f4813b60e60ff39a6?diff=unified&w=0#)

Tinkering with the `fetchPolicy`/`nextFetchPolicy` didn't do much for me to correctly update the data. It had to do with needing to refetch at an appropriate time, which would be at the point of adding a new client. While the `refetch` function works, I noticed a [refetch query](https://www.apollographql.com/docs/react/data/refetching/) can also be triggered upon mutation. 

What I don't like about this implementation is how there are two spots for the same query:
```javascript
/workspace/libs/expo/betterangels/src/lib/screens/Clients/index.tsx


  const { data, loading } = useClientProfilesQuery({
    variables: {
      pagination: { limit: paginationLimit + 1, offset },
      filters: {
        search: filterSearch,
      },
      order: {
        user_FirstName: Ordering.AscNullsFirst,
      },
    },
    fetchPolicy: 'cache-and-network',
    nextFetchPolicy: 'cache-first',
  });
```
```javascript
/workspace/libs/expo/betterangels/src/lib/screens/AddClient/index.tsx

  const [createClient] = useCreateClientProfileMutation({
    refetchQueries: [
      {
        query: ClientProfilesDocument,
        variables: {
          pagination: { limit: 20 + 1, offset: 0 },
          filters: {
            search: '',
          },
          order: {
            user_FirstName: Ordering.AscNullsFirst,
          },
        },
      },
    ],
  });
```

I was thinking about making a useContext hook, but I felt it would complicate the codebase more.

Ultimately, I decided this implementation for the simple fix and little code to make this work altogether.

show-clients

[DEV-504]: https://betterangels.atlassian.net/browse/DEV-504?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libs/expo/betterangels/src/lib/screens/AddClient/index.tsx
+++ b/libs/expo/betterangels/src/lib/screens/AddClient/index.tsx
@@ -7,9 +7,11 @@ import {
   CreateClientProfileInput,
   GenderEnum,
   LanguageEnum,
+  Ordering,
   YesNoPreferNotToSayEnum,
 } from '../../apollo';
 import { MainScrollContainer } from '../../ui-components';
+import { useClientProfilesQuery } from '../Clients/__generated__/Clients.generated';
 import ContactInfo from './ContactInfo';
 import Dob from './Dob';
 import Gender from './Gender';
@@ -41,6 +43,17 @@ export default function AddClient() {
   const [expanded, setExpanded] = useState<undefined | string | null>();
   const [client, setClient] = useState<CreateClientProfileInput>(INITIAL_STATE);
   const [createClient] = useCreateClientProfileMutation();
+  const { refetch } = useClientProfilesQuery({
+    variables: {
+      pagination: { limit: 20 + 1, offset: 0 },
+      filters: {
+        search: '',
+      },
+      order: {
+        user_FirstName: Ordering.AscNullsFirst,
+      },
+    },
+  });
   const [errorState, setErrorState] = useState<string | null>(null);
   const navigation = useNavigation();
   const scrollRef = useRef<ScrollView>(null);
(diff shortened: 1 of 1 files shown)


davelindqvist:
I tried including `optimisticResponse` (as shown here in the [documentation](https://www.apollographql.com/docs/react/performance/optimistic-ui/#example-adding-a-new-object-to-a-list)) but couldn't make it work. 

show-clients-2

----

I also tried updating the cache after mutation, but I found it more cumbersome. Looking at it from a [bigger picture](https://www.apollographql.com/docs/react/data/mutations#supported-methods), I think the `refetchQueries` is the best way to go since it updates the cache and doesn't require a manual update to refresh the client profiles. I also don't think there's any performance or responsiveness issue, therefore not needing the `optimisticResponse` implemented. 

Hopefully I didn't miss anything else in the Apollo documentation.

vecchp (approved):

shiozakiyuko (approved):
I tested this on iOS, and the new client I added showed up in the All Clients page. Looks good to me!

PR #449, opened 2024-07-03
Make user email nullable (DEV-501)

mikefeldberg:
DEV-501


apps/betterangels-backend/notes/admin.py
```diff
@@ -24,10 +24,10 @@ class MoodAdmin(admin.ModelAdmin):
     readonly_fields = ("created_at",)
 
     def note_client(self, obj: Mood) -> str:
-        return obj.note.client.email if obj.note.client else ""
+        return obj.note.client.email if obj.note.client and obj.note.client.email else ""
```
lena, line 27: can we use a `__str__` on the `client` to return some other info instead?

apps/betterangels-backend/accounts/managers.py
```diff
@@ -9,25 +9,24 @@
...
 
 class UserManager(BaseUserManager["User"]):
-    def create_user(self, email: str, password: str = "", **extra_fields: Any) -> "User":
-        if not email:
-            raise ValueError("The Email field must be set")
+    def create_user(self, **extra_fields: Any) -> "User":
         extra_fields.setdefault("is_staff", False)
         extra_fields.setdefault("is_superuser", False)
-        email = self.normalize_email(email)
+        email = self.normalize_email(extra_fields.pop("email", None))
```
lena, line 15: slight preference for
```
        email = extra_fields.pop("email", None)
        email = self.normalize_email(email) if email else None
```

apps/betterangels-backend/accounts/managers.py
```diff
@@ -9,25 +9,24 @@
...
         user = self.model(email=email, **extra_fields)
-        user.set_password(password)
+        user.set_password(extra_fields.get("password"))
```
lena, line 17: should password get popped too? it's getting sent in as plain text on like 16

apps/betterangels-backend/accounts/admin.py
```diff
@@ -17,6 +21,32 @@
 )
 
 
+class NotesByAdminMixin(object):
+    def notes_by(self, obj: User) -> SafeString:
+        notes = Note.objects.filter(created_by=obj)
+        note_links = [
+            '<a href="{}">{}</a>'.format(
+                reverse("admin:notes_note_change", args=(note.id,)),
+                f"Note {note.id}: {note} (with {note.client.full_name if note.client else None} {note.interacted_at.date()})",
```
lena, line 30: a lil strange to interpolate with `None`, and also this looks like something that can be put on the Note model so that in can be reused. I see you're interpolating `{note}` directly here, what does that evaluate to? the note title? you can move all of this into a `note_label_with_client` property on note, and then reuse it in other places potentially

apps/betterangels-backend/accounts/managers.py
```diff
@@ -36,7 +35,7 @@ def create_superuser(self, email: str, password: str = "", **extra_fields: Any)
             raise ValueError(_("Superuser must have is_staff=True."))
         if extra_fields.get("is_superuser") is not True:
             raise ValueError(_("Superuser must have is_superuser=True."))
-        return self.create_user(email, password, **extra_fields)
```
vecchp, line 39: What is replacing email? Username?


https://github.com/django/django/blob/89557d4c66b469616fc3a16ba11b2999233efa8d/django/contrib/auth/models.py#L159

apps/betterangels-backend/accounts/managers.py
```diff
@@ -9,25 +9,24 @@
...
+    def create_client(self, **extra_fields: Any) -> "User":
         random_id = uuid.uuid4()
-        email = email or self.normalize_email(f"client_{random_id}@example.com")
-        client = self.create_user(email, username=random_id, **extra_fields)
+        client = self.create_user(username=random_id, **extra_fields)
         client.set_unusable_password()
+
         return client
 
-    def create_superuser(self, email: str, password: str = "", **extra_fields: Any) -> "User":
```
vecchp, line 30: Are we not going to conform the the base class's interface anymore? 

https://github.com/django/django/blob/89557d4c66b469616fc3a16ba11b2999233efa8d/django/contrib/auth/models.py#L164

apps/betterangels-backend/accounts/managers.py
```diff
@@ -9,25 +9,24 @@
 
 
 class UserManager(BaseUserManager["User"]):
-    def create_user(self, email: str, password: str = "", **extra_fields: Any) -> "User":
```
vecchp, line 12: I would prefer if we are passing optional args that we explicitly pass them as optional and not rely on extra fields.

lena: realized all these methods weren't necessary since we were just using the wrong base class and then just reimplementing the same functionality that django has for free in `from django.contrib.auth.models import UserManager as DjangoUserManager`

apps/betterangels-backend/accounts/admin.py
```diff
@@ -17,6 +21,28 @@
...
+        notes = Note.objects.filter(client=obj)
+        note_links = [
+            '<a href="{}">{}</a>'.format(
+                reverse("admin:notes_note_change", args=(note.id,)), note.label_with_created_by
+            )
+            for note in notes
+        ]
+        return format_html("<br>".join(note_links))
+
+
```
vecchp, line 45: Is this being reused anywhere else besides here? I'm trying to understand why we went with a mixin instead of just adding it directly to the admin.

apps/betterangels-backend/accounts/models.py
```diff
@@ -79,11 +80,14 @@ class User(AbstractBaseUser, PermissionsMixin):
     servicerequestuserobjectpermission_set: models.QuerySet["ServiceRequestUserObjectPermission"]
 
     def __str__(self: "User") -> str:
-        return self.email
+        return f"{self.full_name} ({self.pk})"
```
vecchp, line 83: Hmm, why do we ahve the PK here?

apps/betterangels-backend/accounts/models.py
```diff
@@ -79,11 +80,14 @@ class User(AbstractBaseUser, PermissionsMixin):
...
 
     @model_property
     def full_name(self: "User") -> str:
-        return f"{self.first_name} {self.last_name}"
+        name_parts = filter(None, [self.first_name, self.middle_name, self.last_name])
+        display_name = " ".join(name_parts).strip()
```
vecchp, line 88: Code Golf not a requirement for change, but useful to know since I've seen us use this pattern in other places.:
This can be simpler if you want to use [compact from funcy](https://funcy.readthedocs.io/en/stable/colls.html?highlight=compact#compact)

```python
from funcy import compact
display_name = " ".join(compact([self.first_name, self.middle_name, self.last_name])).strip()
```

```python
from itertools import filterfalse
display_name = " ".join(filterfalse(bool, [self.first_name, self.middle_name, self.last_name])).strip()
```

apps/betterangels-backend/accounts/models.py
vecchp: If this is for display wouldn't we want something like `{self.pk} - {self.full_name}`.  Let me know if there is a reasoning here.

apps/betterangels-backend/accounts/admin.py
vecchp: Also I don't think this should be in accounts this seems to be note specific.

apps/betterangels-backend/accounts/admin.py
```diff
@@ -27,7 +53,7 @@ class ExtendedOrganizationInvitationAdmin(ModelAdmin[ExtendedOrganizationInvitat
     list_filter = ("organization",)
 
 
-class UserAdmin(BaseUserAdmin):
+class UserAdmin(NotesForAdminMixin, NotesByAdminMixin, BaseUserAdmin):
```
vecchp, line 56: Why are we adding these fields in the user creation form?

vecchp (changes requested):

apps/betterangels-backend/accounts/admin.py
lena: delete this

vecchp: let's just nix this

apps/betterangels-backend/accounts/managers.py
lena: yup `username` will now be required

apps/betterangels-backend/accounts/models.py
lena: funcy is not a dependency, and using itertools.filterfalse causes an `arg-type` typecheck error:
```
 error: Argument 2 to "filterfalse" has incompatible type "List[Optional[str]]"; expected "Iterable[str]"  [arg-type]
```

vecchp (approved):

apps/betterangels-backend/accounts/backends.py
```diff
@@ -29,8 +30,9 @@ def invite_by_email(
         try:
             user = self.user_model.objects.get(email=email)
         except self.user_model.DoesNotExist:
-            user = self.user_model.objects.create(email=email)
+            user = self.user_model.objects.create(username=str(uuid.uuid4()), email=email)
             user.set_unusable_password()
+            # TODO: DEV-554 - Fix the issue that turning the user to inactive disables social login
```
lena, line 35: link for ticket

PR #452, opened 2024-07-03
adjust public note footer action bar (DEV-403)

bojjakaseenadhreddy:
DEV-403

### DIFF
--- a/libs/expo/betterangels/src/lib/screens/PublicNote/index.tsx
+++ b/libs/expo/betterangels/src/lib/screens/PublicNote/index.tsx
@@ -89,92 +89,94 @@ export default function PublicNote({ noteId }: { noteId: string }) {
   }
 
   return (
-    <MainScrollContainer pb={40} bg={Colors.NEUTRAL_EXTRA_LIGHT}>
-      <View
-        style={{
-          gap: Spacings.sm,
-          flex: 1,
-        }}
-      >
-        <TextBold size="lg">Write Public Note</TextBold>
-        {autoNote !== publicNote && (
-          <View
-            style={{
-              padding: Spacings.xs,
-              borderRadius: 8,
-              backgroundColor: Colors.WARNING_EXTRA_LIGHT,
-            }}
-          >
-            <TextRegular size="sm" color={Colors.WARNING_DARK}>
-              You changed the form above. Please review note text for
-              consistency.
-            </TextRegular>
-          </View>
-        )}
-        <TextRegular size="md">
-          Use the generated text below to get started. When finished, click
-          “Save Note.”
-        </TextRegular>
-
-        <View style={{ flexGrow: 1 }}>
-          <TextInput
-            value={publicNote}
-            onChangeText={(text) => onChange(text)}
-            multiline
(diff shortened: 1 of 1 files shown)


mikefeldberg (commented):
I think this is technically correct but probably not what we're going for in terms of look. May need product to weigh in here because the design linked in the ticket is for a differently shaped device. So, while I think this implementation would look good on a screen with tighter corners, it looks cramped and out of place on something with round corners, like an iPhone.

image

I also left a question for product in the ticket comments. The design and the AC don't match, and it's unclear if the AC is missing a requirement or if the design is out of date

bojjakaseenadhreddy:
> I think this is technically correct but probably not what we're going for in terms of look. May need product to weigh in here because the design linked in the ticket is for a differently shaped device. So, while I think this implementation would look good on a screen with tighter corners, it looks cramped and out of place on something with round corners, like an iPhone.
> 
> image
> I also left a question for product in the ticket comments. The design and the AC don't match, and it's unclear if the AC is missing a requirement or if the design is out of date

Fixed this issue based on the conversation in JIRA.

libs/expo/betterangels/src/lib/screens/PublicNote/index.tsx
```diff
@@ -89,92 +92,94 @@ export default function PublicNote({ noteId }: { noteId: string }) {
...
+          alignItems: 'center',
+          backgroundColor: Colors.WHITE,
+          padding: Spacings.sm,
+          paddingBottom: bottomOffset + Spacings.sm,
+          gap: Spacings.xs,
+          elevation: 5,
+          shadowColor: '#CCC',
+          shadowOffset: { width: 0, height: -3 },
+          shadowOpacity: 0.3,
+          shadowRadius: 4,
```
mikefeldberg, line 148:
```suggestion
          shadowRadius: Spacings.xxs,
```

libs/expo/betterangels/src/lib/screens/PublicNote/index.tsx
```diff
@@ -89,92 +92,94 @@ export default function PublicNote({ noteId }: { noteId: string }) {
...
+            gap: Spacings.sm,
+            flex: 1,
+          }}
+        >
+          <TextBold size="lg">Write Public Note</TextBold>
+          {autoNote !== publicNote && (
+            <View
+              style={{
+                padding: Spacings.xs,
+                borderRadius: 8,
```
mikefeldberg, line 108:
```suggestion
                borderRadius: Spacings.xs,
```

mikefeldberg (approved):
I _think_ we should be using constants for formatting wherever possible. If I'm wrong, or if there's a reason we're using hardcoded values here, 👍

libs/expo/betterangels/src/lib/screens/PublicNote/index.tsx
bojjakaseenadhreddy: Thanks for catching it. I just created a subtask that will fix them at the application level. [Refactor the application to use Constant values instead of Hard coded values](https://betterangels.atlassian.net/browse/DEV-536)

PR #453, opened 2024-07-03
Update padding for Interactions screen (DEV-500)

bojjakaseenadhreddy:
DEV-500


mikefeldberg (approved):
lgtm! While we're at it, I think we should reduce the padding around "n interactions". @shiozakiyuko, what do you think?

shiozakiyuko:
Sorry, @mikefeldberg I didn't see your comment/question until now!
Yea, I think this space between "n Interactions" and the first card is coming from the filters we used to have or something.
Let's reduce this space to 16px.

PR #463, opened 2024-07-11
disable autocorrect for name, email, HMIS id fields (DEV-547)

bojjakaseenadhreddy:
DEV-547


libs/expo/betterangels/src/lib/screens/AddEditClient/ContactInfo.tsx
```diff
@@ -76,6 +76,7 @@ export default function ContactInfo(props: IContactInfoProps) {
           required
           placeholder="Enter Email"
           label="Email"
+          autoCorrect={false}
```
mikefeldberg, line 79: could you please also add `autoCapitalize={false}`? Not in the ticket but I just realized that emails shouldn't be autocapped 🙏

mikefeldberg (approved):

PR #466, opened 2024-07-11
Implement Import Aliases to Replace Long Relative Paths (DEV-445)

davelindqvist:
[DEV-445](https://betterangels.atlassian.net/browse/DEV-445?atlOrigin=eyJpIjoiYmM4YmFlZWM3MzAyNGNjOGJhMzY0ZDJhYmQ3YjMwOTYiLCJwIjoiaiJ9)

Currently, the only additional import alias is for `/apps/betterangels/config.ts`.

Excluding these files here:

```
.eslintrc.json, 
.yarnrc.yml, 
package.json, 
project.json, 
tsconfig.app.json, 
tsconfig.json, 
tsconfig.spec.json, 
.detoxrc.json, 
jest.config.json, 
tsconfig.e2e.json, 
jest.config.ts, 
tsconfig.lib.json, 
tsconfig.storybook.json, 
codegen.ts, 
.generated.ts
```

Then there are eleven imports that require three directories up to get to the relevant module:

image

I bring this up because I wonder if these need to have aliases as well?

[DEV-445]: https://betterangels.atlassian.net/browse/DEV-445?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.eslintrc.json
```diff
@@ -11,7 +11,7 @@
           "error",
           {
             "enforceBuildableLibDependency": true,
-            "allow": [],
+            "allow": ["@monorepo/**"],
```
davelindqvist, line 14: Reference: 

https://stackoverflow.com/questions/74296342/nx-error-for-relative-imports-within-the-same-project

vecchp: after reading this it seems like the relative pathing is encouraged within a module / lib and references should be across.  Maybe this is showing that we aren't necessarily splitting up our libs right?  Can you double check the usage and see how much are cross module referencs and how many are within a module?

vecchp: I think this is also indicating that we should be pulling out as much as we can from app and putting it into our libs and passing the config in maybe?

davelindqvist: I've decided to move the `config.ts` file under `libs/expo/shared/utils/src/lib/config.ts` as the `libs` directory is already enabled for sharing code. 

Regarding the cross module references, it should be fine as long as we're sharing and enforcing the scope between the libraries and projects. Here's a more detailed understanding of it through nx [1](https://nx.dev/getting-started/tutorials/react-monorepo-tutorial#imposing-constraints-with-module-boundary-rules) and [2](https://nx.dev/recipes/enforce-module-boundaries/tag-multiple-dimensions), and the reasons for its benefits [here](https://nx.dev/concepts/decisions/code-ownership).

I'm hesitant to tag anything without understanding the relationship between the apps and libraries, but welcome a discussion to flesh out the details. Once we get that achieved, it should be easy to implement.

davelindqvist:
Paul and I spoke, and decided to create a path alias [here](https://github.com/BetterAngelsLA/monorepo/blob/d9d2e97391a7d055f82c6d8762cb491899224dd3/tsconfig.base.json#L27), which will allow autocompletion. Other methods to establish path aliases would create more trouble and delineate from the folder structures already in place.

---

On another note, I noticed redundancy with hexToRGBA in the codebase. Will do a separate PR for that.

vecchp (approved):

PR #468, opened 2024-07-15
adjust clear field icon (DEV-549)

bojjakaseenadhreddy:
DEV-549


bojjakaseenadhreddy:
Screenshot 2024-07-15 at 10 29 45 AM

mikefeldberg (changes requested):
This increases the visual size of the icon. Let's expand the pressable area without changing how the form looks. Something like [this](https://github.com/BetterAngelsLA/monorepo/compare/DEV-549/increase-pressable-area?expand=1#)

mikefeldberg (approved):

PR #474, opened 2024-07-16
Add security, drugs, and program fees to Shelter Admin (DEV-560)

vecchp:
DEV-560

Adds missing fields: security, drugs, and program fees


davelindqvist (approved):
LGTM

PR #476, opened 2024-07-17
Fix date picker Done button display (DEV-521)

Davit-BetterAngels:
DEV-521

fixed date picker button. 

also found a bug when creating a note with the date and time and fixed it


mikefeldberg (approved):
lgtm, but we should confirm with prod if this is the correct button placement. afaict, the [designs](https://www.figma.com/design/waIeX8dknHKedTrVxCQAAG/HIFI---1st-MVPs?node-id=1232-33209&m=dev) don't include date/time edit ui. 

Specifically:
1.  it's not obvious if the "Done" button should be inside or outside the white background of the picker(s) 
2. should the picker(s) have the same gray border as the field cards?

If there _is_ a design for this, let's make sure this PR matches that design. If there _isn't_ a design for this, let's merge this as is but let product know that this is our best guess at how it should look. They can provide updated designs and create a new ticket to update, if necessary.

image

johnr54321:
Checked DOB date field on IOS and Android.  Is completely visible on IOS. (Original AC on Jira ticket is only this.)
Also checked note date fields to verify that the done button is not cut off there.

I got a weird behavior on Android.  The date picker is a calendar.  However, you can't click ok and choose the date.  If you do, it just brings up the same prompt again.

share_1940575398383016198

johnr54321:
tested on ios and andriod. pass.  I was able to pick a date in both.  Android using the calendar, and IOS selecting the date.

PR #478, opened 2024-07-22
Update client search (DEV-567, DEV-568)

mikefeldberg:
DEV-567, DEV-568

This PR:
1. adds nickname and middle name to list of searched fields
2. modifies client search from full string search to parsed term search
    - old: search term = "Mike Feldberg"; return objects where at least one field contains the entire search term
    -  new: search term = "Mike Feldberg"; parse to "Mike" and "Feldberg"; return objects where each term is contained in at least one of the searched fields


apps/betterangels-backend/accounts/types.py
```diff
@@ -61,17 +61,27 @@ def search(
...
+                | Q(hmis_id__icontains=term)
             )
 
-        return queryset, Q()
+            query &= q_search
+
+        return (
+            queryset.filter(query),
+            Q(),
+        )
```
lena, line 84: a lil more performant:
```
from operator import or_, and_
``` 
(up above) then:

```
q_objects = []
combined_word_q_objects = []

searchable_fields = ["user__first_name", "user__last_name", ...]

word_q_objects = [Q(**{f"{field}__icontains": word}) for field in searchable_fields]
combined_word_q_objects.append((reduce(_or, word_q_objects)))
q_objects.append(Q(*combined_word_q_objects))
queryset = model.objects.filter(reduce(and_, q_objects))
```

lena (approved):
👍

PR #481, opened 2024-07-24
Add ClientContact and update ClientProfile (DEV-176)

mikefeldberg:
DEV-176


apps/betterangels-backend/accounts/models.py
```diff
@@ -115,6 +120,16 @@ class ClientProfile(models.Model):
     veteran_status = TextChoicesField(choices_enum=YesNoPreferNotToSayEnum, blank=True, null=True)
 
 
+class ClientContact(models.Model):
+    client_profile = models.ForeignKey(ClientProfile, on_delete=models.CASCADE, related_name="contacts")
+    name = models.CharField(max_length=100, null=True, blank=True)
+    email = models.EmailField(null=True, blank=True)
+    phone_number = models.CharField(max_length=15, null=True, blank=True)
```
vecchp, line 127: Can this be phonenumber field?

https://github.com/BetterAngelsLA/monorepo/blob/d271e7bc47aac526fbf249f156325ce7ca4d6d7e/apps/betterangels-backend/shelters/models.py#L142

apps/betterangels-backend/accounts/models.py
```diff
@@ -115,6 +120,16 @@ class ClientProfile(models.Model):
...
+    mailing_address = models.TextField(null=True, blank=True)
```
vecchp, line 128: Should this be an address?

apps/betterangels-backend/accounts/schema.py
```diff
@@ -73,23 +73,35 @@ def generate_magic_link(self, info: Info, data: MagicLinkInput) -> MagicLinkResp
...
                 info,
                 ClientProfile,
                 {
                     **client_profile_data,
-                    "user": client,
+                    "user": client_user,
                 },
             )
 
+            if contacts_data:
```
vecchp, line 94: Wouldn't this be simpler as it's own mutation to and and remove contacts?

apps/betterangels-backend/notes/types.py
```diff
@@ -41,7 +41,7 @@ class CreateNoteAttachmentInput:
 
 @strawberry_django.type(models.ServiceRequest, pagination=True)
 class ServiceRequestType:
-    id: auto
+    id: ID
```
vecchp, line 44: Can you explain why we're going from auto to ID?

vecchp: Is it because the schema created requires an ID vs making it optional?

vecchp (changes requested):
Good work, here are a few comments.  If you want to rapid fire discuss them before standup or after 11am today I'll make time.

vecchp:
CI is also failing

apps/betterangels-backend/accounts/models.py
mikefeldberg: not as requested/designed

apps/betterangels-backend/accounts/schema.py
mikefeldberg: yes. Working on this, I started thinking it may make a lot of sense to have individual _Save_ buttons for each section. Barring that, having the one _Save_ button trigger multiple mutations is also preferable to `update_client_profile` hitting 100+ lines

apps/betterangels-backend/notes/types.py
mikefeldberg: yep

apps/betterangels-backend/accounts/schema.py
mikefeldberg: the issue with individual mutations is that creating a client profile would require two round trips. Client contact needs a client profile id, so we'd need to (1) create client profile; (2) use the returned id to create client contact. Maybe that's ok because most operations will be updates, not creates? And I assume most clients would be created quick and dirty, not with the entire profile filled out. What do you think?

apps/betterangels-backend/accounts/models.py
mikefeldberg: being worked on in [DEV-631](https://betterangels.atlassian.net/browse/DEV-631), https://github.com/BetterAngelsLA/monorepo/pull/489

vecchp (approved):

PR #486, opened 2024-07-25
Change user self str (DEV-623); fix user admin is_client (DEV-624)

mikefeldberg:
DEV-623, DEV-624

### DIFF
--- a/apps/betterangels-backend/accounts/models.py
+++ b/apps/betterangels-backend/accounts/models.py
@@ -80,7 +80,7 @@ class User(AbstractBaseUser, PermissionsMixin):
     servicerequestuserobjectpermission_set: models.QuerySet["ServiceRequestUserObjectPermission"]
 
     def __str__(self: "User") -> str:
-        return f"{self.pk} - {self.full_name}"
+        return f"{self.full_name}"
 
     @model_property
     def full_name(self: "User") -> str:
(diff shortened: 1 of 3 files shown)


apps/betterangels-backend/accounts/models.py
```diff
@@ -80,7 +80,7 @@ class User(AbstractBaseUser, PermissionsMixin):
     servicerequestuserobjectpermission_set: models.QuerySet["ServiceRequestUserObjectPermission"]
 
     def __str__(self: "User") -> str:
-        return f"{self.pk} - {self.full_name}"
+        return f"{self.full_name}"
```
lena, line 83: maybe:

```suggestion
        return self.full_name if self.full_name else self.pk
```

mikefeldberg: good call. done

apps/betterangels-backend/accounts/models.py
```diff
@@ -80,7 +80,7 @@ class User(AbstractBaseUser, PermissionsMixin):
     servicerequestuserobjectpermission_set: models.QuerySet["ServiceRequestUserObjectPermission"]
 
     def __str__(self: "User") -> str:
-        return f"{self.pk} - {self.full_name}"
+        return f"{self.full_name if self.full_name else self.pk}"
```
lena, line 83:
```suggestion
        return self.full_name if self.full_name else self.pk
```

lena (approved):

PR #482, opened 2024-07-24
Add support for multiple HMIS IDs (DEV-383)

mikefeldberg:
DEV-383

This PR:
1. adds an `HmisProfile` table to store clients' agency-specific hmis ids. 
2. migrates existing client hmis data to the `HmisProfile` table 
3. updates graphql endpoints

Two follow up steps are required:
1. DEV-384: update frontend to use new fields
2. DEV-524: remove old `hmis_id` from `ClientProfile`, types, queries, and mutations

PR #490, opened 2024-07-30
Fix hmis profile data migration (DEV-383)

mikefeldberg:
`ClientProfile.hmis_id` is optional. `HmisProfile.hmis_id` is required. 

The data migration in #482 didn't account for that and tried to create hmis profiles for all client profiles, including those with no hmis id, resulting in a null constraint integrity error

### DIFF
--- a/apps/betterangels-backend/accounts/migrations/0027_hmisprofile_hmisprofile_unique_hmis_id_agency.py
+++ b/apps/betterangels-backend/accounts/migrations/0027_hmisprofile_hmisprofile_unique_hmis_id_agency.py
@@ -10,9 +10,8 @@ def add_hmis_profile(apps, schema_editor):
     ClientProfile = apps.get_model("accounts", "ClientProfile")
     HmisProfile = apps.get_model("accounts", "HmisProfile")
 
-    client_profiles = ClientProfile.objects.all()
-
-    for client_profile in client_profiles:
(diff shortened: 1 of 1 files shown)


apps/betterangels-backend/accounts/migrations/0027_hmisprofile_hmisprofile_unique_hmis_id_agency.py
```diff
@@ -10,9 +10,8 @@ def add_hmis_profile(apps, schema_editor):
     ClientProfile = apps.get_model("accounts", "ClientProfile")
     HmisProfile = apps.get_model("accounts", "HmisProfile")
 
-    client_profiles = ClientProfile.objects.all()
-
-    for client_profile in client_profiles:
+    client_profiles_with_hmis_id = ClientProfile.objects.filter(hmis_id__isnull=False)
```
lena, line 13: non-blocker
```suggestion
    client_profiles_with_hmis_id = ClientProfile.objects.exclude(hmis_id=None)
```
do you happen to know if hmis_id could be `""` (empty string)? if so, might want to filter those out too

mikefeldberg: fortunately, `ClientProfile.hmis_id` has a unique constraint, so there's only one "". Still a good idea; updated

PR #502, opened 2024-08-06
Migrate phone number field to use library (DEV-631)

davelindqvist:
[DEV-631](https://betterangels.atlassian.net/browse/DEV-631?atlOrigin=eyJpIjoiODAxYTAyYzJlYmNiNGQ0NTg0YzY1ODMxZTIyNmFmMWYiLCJwIjoiaiJ9)

---

Had to make a custom scalar to change from `CharField` to `PhoneNumberField`.

[DEV-631]: https://betterangels.atlassian.net/browse/DEV-631?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/betterangels-backend/accounts/migrations/0030_alter_clientcontact_phone_number.py
```diff
@@ -0,0 +1,25 @@
+# Generated by Django 5.0.7 on 2024-08-06 00:29
```
mikefeldberg, line 1: these migrations can be collapsed into one. easiest way to do that is to delete them both and rerun `makemigrations`

apps/betterangels-backend/accounts/models.py
```diff
@@ -118,14 +120,29 @@ class Meta:
         constraints = [models.UniqueConstraint(fields=["hmis_id", "agency"], name="unique_hmis_id_agency")]
 
 
+def phone_number_field_default_value() -> PhoneNumber:
```
mikefeldberg, line 123: is there a reason we need this?

apps/betterangels-backend/accounts/models.py
```diff
@@ -118,14 +120,29 @@ class Meta:
...
+
 class ClientProfile(models.Model):
     user = models.OneToOneField(User, on_delete=models.CASCADE, related_name="client_profile")
     address = models.TextField(blank=True, null=True)
     date_of_birth = models.DateField(blank=True, null=True)
     gender = TextChoicesField(choices_enum=GenderEnum, blank=True, null=True)
     hmis_id = models.CharField(max_length=50, blank=True, null=True, db_index=True, unique=True)
     nickname = models.CharField(max_length=50, blank=True, null=True)
-    phone_number = models.CharField(max_length=15, blank=True, null=True)
+    phone_number = models.JSONField(
```
mikefeldberg, line 143: This should be a `PhoneNumberField` without being wrapped in a `JSONField`. Let's take another look at this later this week

mikefeldberg (changes requested):

apps/betterangels-backend/accounts/migrations/0030_alter_clientcontact_phone_number.py
davelindqvist: Although, I was able to make a single migration, it kept giving me this error when I tried to migrate it:

```
> poetry run python manage.py migrate

Operations to perform:
  Apply all migrations: account, accounts, admin, auth, common, contenttypes, guardian, notes, organizations, pghistory, post_office, sessions, shelters, sites, socialaccount, waffle
Running migrations:
  Applying accounts.0029_alter_clientcontact_phone_number_and_more...Traceback (most recent call last):
  File "/workspace/.venv/lib/python3.12/site-packages/django/db/backends/utils.py", line 103, in _execute
    return self.cursor.execute(sql)
           ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/workspace/.venv/lib/python3.12/site-packages/psycopg/cursor.py", line 97, in execute
    raise ex.with_traceback(None)
psycopg.errors.InvalidTextRepresentation: invalid input syntax for type json
DETAIL:  Token "+" is invalid.
CONTEXT:  JSON data, line 1: +...

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/workspace/apps/betterangels-backend/manage.py", line 30, in <module>
    main()
  File "/workspace/apps/betterangels-backend/manage.py", line 26, in main
    execute_from_command_line(sys.argv)
```

Only when I separated the migrations, did it start working

apps/betterangels-backend/accounts/models.py
davelindqvist: Reflecting on this error here from this [CI/CD run](https://github.com/BetterAngelsLA/monorepo/actions/runs/10170675554/job/28130582042):

```python
  >       raise TypeError(f'Object of type {o.__class__.__name__} '
                          f'is not JSON serializable')
  E       TypeError: Object of type PhoneNumber is not JSON serializable
```

It made me see that putting `PhoneNumberField()` wouldn't allow Django to understand how to put this in a serializable context. Therefore, I had to wrap it around with [JSONField](https://docs.djangoproject.com/en/5.0/ref/forms/fields/#jsonfield).

apps/betterangels-backend/accounts/models.py
davelindqvist: You're right, it looks like we don't!

I had thought we needed a default PhoneNumber dict so the JSONField wrapper could understand how to serialize it. The tests ended up passing without this in it, so it looks like adding this does nothing of value.

Thank you!!!

apps/betterangels-backend/accounts/models.py
davelindqvist: My apologies for not adding much context throughout this PR. I'll be more mindful about it next time so you're not in the dark of what's going on. 

And yes, we shall take another look at this next week!

davelindqvist:
Close pull request. Will revisit when the need comes as there is a type issue with [strawberry](https://github.com/strawberry-graphql/strawberry/issues/3369) when using a custom scalar.

PR #504, opened 2024-08-06
Add pghistory to Shelter model and tracking add, update, & remove

quipcode:
DEV-630


apps/betterangels-backend/shelters/models.py
```diff
@@ -28,69 +29,79 @@
 
 
 # Advanced Info
+@pghistory.track(pghistory.InsertEvent("shelter_type.add"), pghistory.DeleteEvent("shelter_type.remove"))
```
lena, line 32: does this only track the `ShelterType` table itself? or the relationship to the Shelter model? seems like it wouldn't track the later?

quipcode: correct this only tracks the events on the ShelterType's table. I honestly don't know how to track the relationship of the Shelter's table in the corresponding events table. I've been unsuccessful in trying to create a new ShelterType class that is similar to the NoteNextSteps class defined here [apps/betterangels-backend/notes/models.py#L251](https://github.com/BetterAngelsLA/monorepo/blob/65b34264e35d5607d4af792101aba24afa7bd0a9/apps/betterangels-backend/notes/models.py#L251). Additionally that did not create a relationship between the notes_noteevent and the notes_notnextstepsevent table either see attached ERD ERD(top left of shared image).

At this time this PR will add tracking for all the related shelter fields(see top right) but at this time there is no relationship defined for any of the tables we are tracking

quipcode: I don't know how to go about creating a custom pghistory event model such that it maintains the same relationships that the actual Shelter's table has and I'm not sure if that is in the scope of the ticket [DEV-630](https://betterangels.atlassian.net/browse/DEV-630)

PR #507, opened 2024-08-07
Client profile demographic fields (DEV-175)

mikefeldberg:
DEV-175

In the process, I found an error that was introduced in #512 and masked by a typo introduced to a test in #482. Fixed the test and error


apps/betterangels-backend/accounts/models.py
```diff
@@ -123,18 +131,46 @@ class Meta:
 class ClientProfile(models.Model):
     user = models.OneToOneField(User, on_delete=models.CASCADE, related_name="client_profile")
     address = models.TextField(blank=True, null=True)
+    city_of_birth = models.CharField(max_length=100, blank=True, null=True)
```
lena, line 134: city of birth sounds like "Los Angeles", not "Los Angeles, CA", i think place is better because then people can write it more generic things like "Ukraine"
```suggestion
    place_of_birth = models.CharField(max_length=100, blank=True, null=True)
```

apps/betterangels-backend/accounts/models.py
```diff
@@ -123,18 +131,46 @@ class Meta:
...
     date_of_birth = models.DateField(blank=True, null=True)
+    eye_color = TextChoicesField(choices_enum=EyeColorEnum, blank=True, null=True)
     gender = TextChoicesField(choices_enum=GenderEnum, blank=True, null=True)
+    hair_color = TextChoicesField(choices_enum=HairColorEnum, blank=True, null=True)
+    height_in_inches = models.FloatField(blank=True, null=True)
```
lena, line 139:
```suggestion
    height_in_inches = models.DecimalField(decimal_places=2, max_digits=5, null=True)
```

apps/betterangels-backend/accounts/schema.py
```diff
@@ -42,9 +44,19 @@
...
-    )
+    @strawberry_django.field(extensions=[HasRetvalPerm(perms=[ClientProfilePermissions.VIEW])])
+    def client_profile(self, info: Info, pk: strawberry.ID) -> ClientProfileType:
+        client_profile = ClientProfile.objects.prefetch_related(
+            Prefetch(
+                "contacts",
+                queryset=ClientContact.objects.filter(
+                    relationship_to_client=RelationshipTypeEnum.CURRENT_CASE_MANAGER
+                ).order_by("created_at"),
+                to_attr="case_managers",
```
lena, line 55: nice work!

apps/betterangels-backend/accounts/schema.py
```diff
@@ -241,17 +253,9 @@ def update_client_profile(self, info: Info, data: UpdateClientProfileInput) -> C
...
-                client_profile,
-                {
-                    **client_profile_data,
-                },
-            )
-
             if hmis_profiles:
                 hmis_profile_updates_by_id = {hp["id"]: hp for hp in hmis_profiles if hp.get("id")}
-                hmis_profiles_to_create = [hp for hp in hmis_profiles if hp.get("id") is None]
+                hmis_profiles_to_create = [hp for hp in hmis_profiles if not hp.get("id")]
```
lena, line 258: i like both ways but the old way is a lil easier to read for me, just a thought

apps/betterangels-backend/accounts/tests/test_models.py
```diff
@@ -54,6 +54,19 @@ def test_is_outreach_authorized(
         remove_organization_permission_group(unauthorized_org)
         self.assertEqual(user.is_outreach_authorized, should_succeed)
 
+    def test_display_pronouns(self) -> None:
+        client_profile = baker.make(ClientProfile)
+        self.assertIsNone(client_profile.display_pronouns)
+
+        client_profile.pronouns = PronounEnum.HE_HIM_HIS
+        client_profile.save()
+        self.assertEqual(client_profile.display_pronouns, PronounEnum.HE_HIM_HIS.label)
```
lena, line 63:
```suggestion
        self.assertEqual(client_profile.display_pronouns, "he/him/his")
```
just because otherwise if `.label` were coming back as `None` for some reason, this test wouldn't fail since you're using that method in the code AND in the test

apps/betterangels-backend/accounts/types.py
```diff
@@ -211,17 +219,16 @@ class ClientProfileType(ClientProfileBaseType):
...
     hmis_profiles: Optional[List[Optional[HmisProfileType]]] = strawberry_django.field()
     household_members: Optional[List[ClientHouseholdMemberType]]
 
     @strawberry.field
-    def age(self) -> Optional[int]:
-        if not self.date_of_birth:
-            return None
+    def display_case_manager(self, info: Info) -> str:
+        if case_managers := getattr(self, "case_managers", None):
+            return str(case_managers[-1].name)
```
lena, line 229: lovely

lena (approved):

apps/betterangels-backend/accounts/models.py
mikefeldberg: good call

apps/betterangels-backend/accounts/schema.py
mikefeldberg, line 258: for posterity: this was a substantive change, not just style/pattern. When `.get()` doesn't get anything, it returns falsey, but not `None`

apps/betterangels-backend/accounts/tests/test_models.py
mikefeldberg: good catch

apps/betterangels-backend/accounts/schema.py
mikefeldberg: shame that this was necessary. I was hoping the query optimizer would take care of stuff like this

PR #516, opened 2024-08-14
Username and Passwork login workflow for app review

vecchp:
DEV-672


mikefeldberg (approved):
inputs hidden by keyboard, as you noted

PR #522, opened 2024-08-19
Client profile uploads backend (DEV-679)

mikefeldberg:
DEV-679

This PR includes a lot of unrelated frontend type updates and temporary backend type updates, related to DEV-694


apps/betterangels-backend/accounts/tests/test_permissions.py
```diff
@@ -150,3 +152,139 @@ def test_view_client_profiles_permission(self, user_label: str, should_succeed:
 
         expected_client_count = client_count if should_succeed else 0
         self.assertTrue(len(response["data"]["clientProfiles"]) == expected_client_count)
+
+
+@override_settings(DEFAULT_FILE_STORAGE="django.core.files.storage.InMemoryStorage")
+class ClientDocumentPermessionTestCase(ClientProfileGraphQLBaseTestCase):
```
mikefeldberg, line 158: Mirroring attachment permission structure because we don't have document-specific permission requirements. I'm assuming that we'll want CMs from other orgs to be able to create and view client docs

apps/betterangels-backend/accounts/types.py
```diff
@@ -130,12 +145,14 @@ class UserBaseType:
 
 @strawberry_django.type(User)
 class UserType(UserBaseType):
+    # TODO: has_accepted_tos, has_accepted_privacy_policy, is_outreach_authorized shouldn't be optional.
+    # Temporary fix while we figure out type generation
     id: ID
-    username: auto
+    has_accepted_tos: Optional[bool]
```
mikefeldberg, line 151: `has_accepted_tos`, `has_accepted_privacy_policy`, `is_outreach_authorized` shouldn't be optional. Temporary fix while we figure out type generation. Without `Optional`, typescript requires user queries to include these fields on the request (including queries wherein `user` is a field, e.g., client profile, note)

mikefeldberg, line 151: `has_accepted_tos`, `has_accepted_privacy_policy`, `is_outreach_authorized` shouldn't be optional. This is a temporary fix while we figure out type generation. Without `Optional`, these fields must be included on every user query (including queries wherein `user` is a field, e.g., client profile, note) to avoid type errors. I assume this means we're doing something wrong because, in a graphql query, all available fields are optional; no field _must_ be included on a request

lena (approved):
looks great overall 👏 
a few small non-blocking comments

mikefeldberg:
closing because this branch was inadvertently merged into and committed as part of #531. Will review the above review comments and consider implementing elsewhere

PR #537, opened 2024-08-29
Delete contacts and household (DEV-678)

mikefeldberg:
DEV-678

Didn't quite follow the AC. Deleting by omission instead. ClientProfile.contacts will be compared against the list of contacts on the update mutation. Any contacts not present in the incoming list will be deleted. Same for household members


mikefeldberg:
incorporated into #539

PR #549, opened 2024-09-16
Limit digits of phone number while adding/editing client with error message (DEV-435)

davelindqvist:
[DEV-435](https://betterangels.atlassian.net/browse/DEV-435?atlOrigin=eyJpIjoiNWM5NjVhMGI2M2VjNDJhYjg1Y2Y5ODBjYjNhOTdjMmUiLCJwIjoiaiJ9)

[DEV-435]: https://betterangels.atlassian.net/browse/DEV-435?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

---

As spoken with Mike before, there is hesitation to add another dependency if we don't _truly_ need it. It would be nice to have it since the error message library works very well with react-hook-form, as it's made by the same creator.

Otherwise, I can update the code to match what is shown [here](https://github.com/BetterAngelsLA/monorepo/blob/e3ef7852c6ceead6c1c6bb4255012b2f2c308241/libs/expo/betterangels/src/lib/screens/AddEditClient/ContactInfo.tsx#L40)

---

The phone number has two validations:
1) `Enter a 10-digit phone number without space or special characters`
2) `The phone number is not a valid US number.`

Validation 1 activates when the user inputs a phone number in. Validation 2 activates upon upon submission after the backend finds the phone number to be not valid. For example, a phone number like 123-123-1234 is not valid, but 212-555-2121 is.

Validation 2 is not a requirement created by the product team, but it complements with extra security and error handling.


libs/expo/betterangels/src/lib/screens/AddEditClient/ContactInfo.tsx
```diff
@@ -34,6 +36,12 @@ export default function ContactInfo(props: IContactInfoProps) {
 
   const isContactInfo = expanded === 'Contact Info';
 
+  useEffect(() => {
+    if (phoneNumber === '') {
+      setValue('phoneNumber', null);
+    }
+  }, [phoneNumber]);
+
```
davelindqvist, line 44: I created this because of this error here:

`[ApolloError: Variable '$data' got invalid value '' at 'data.phoneNumber'; Invalid phone number format: (1) The string supplied did not seem to be a phone number.]`

libs/expo/betterangels/src/lib/screens/AddEditClient/RelevantContacts/Contact/index.tsx
```diff
@@ -21,9 +29,25 @@ interface IContactProps {
...
+  );
+
+  useEffect(() => {
+    if (contactPhoneNumber === '') {
+      setValue(
+        `contacts[${index}].phoneNumber` as `contacts.${number}.phoneNumber`,
+        null
+      );
+    }
+  }, [contactPhoneNumber, index]);
```
davelindqvist, line 50: Same as here: https://github.com/BetterAngelsLA/monorepo/pull/549/files#r1783653802

libs/expo/betterangels/src/lib/screens/AddEditClient/index.tsx
```diff
@@ -62,24 +64,90 @@ export default function AddEditClient({ id }: { id?: string }) {
...
+
+  const extractErrorFields = (errors: ApolloError) => {
+    return errors.graphQLErrors
+      .map((error: GraphQLFormattedError) => {
+        const regex = /data\.(.*?);/;
+        const match = error.message.match(regex);
+        return match ? match[1] : null;
+      })
+      .filter(Boolean);
+  };
```
davelindqvist, line 97: This extracts out the node that's after `data` from the error message, and sets the error.

E.g., `[ApolloError: Variable '$data' got invalid value '1231231313' at 'data.phoneNumber'; ]` - extracts `phoneNumber`

E.g., [ApolloError: Variable '$data' got invalid value '1231231313' at 'data.contact[0].phoneNumber'; ] - extracts `contact[0].phoneNumber`

libs/expo/betterangels/src/lib/screens/AddEditClient/index.tsx
```diff
@@ -62,24 +64,90 @@ export default function AddEditClient({ id }: { id?: string }) {
...
+      if (extensions?.message) {
+        extractedFields.map((dataField) => {
+          return methods.setError(
+            dataField as
+              | keyof UpdateClientProfileInput
+              | keyof CreateClientProfileInput,
+            {
+              type: 'custom',
+              message: extensions.message,
+            }
```
davelindqvist, line 128: This forcefully sets the error within react-hook-form to the nodes referred from lines [89 - 97](https://github.com/BetterAngelsLA/monorepo/pull/549/files#r1783655683)

libs/expo/betterangels/src/lib/screens/AddEditClient/index.tsx
```diff
@@ -172,6 +240,9 @@ export default function AddEditClient({ id }: { id?: string }) {
       }
     } catch (err) {
       console.log(err);
+      if (methods.formState.errors) {
+        return;
+      }
```
davelindqvist, line 245: Helpful so that clicking `Update` or `Create` doesn't throw an error.

johnr54321 (approved):
Tested validation of error message 1 on android and ios.  No way to test validation 2 manually.  Passing.

libs/expo/betterangels/src/lib/screens/AddEditClient/RelevantContacts/Contact/index.tsx
Davit-BetterAngels: don't like this effect. lets walk through together

libs/expo/betterangels/src/lib/screens/AddEditClient/index.tsx
Davit-BetterAngels: I'm  thinking why not require a user to write correct number on front end but send it to back, check it, throw error, show error. 

just validate with react-hook form and show front error message. 

lets walk through this together

apps/betterangels-backend/common/graphql/types.py
```diff
@@ -1,20 +1,35 @@
 from datetime import datetime
-from typing import Optional, Union
+from typing import Any, Optional, Union
```
shiozakiyuko, line 2: Hi @davelindqvist 

I confirmed the following:
- I confirmed that the "Enter a 10-digit phone number without space or special characters" when the number was 9 digits, or it had a special character.
- I confirmed that it doesn't let me add more than 10 digits.

One thing I noticed
- I didn't see the message "The phone number is not a valid US number." when I entered 10 digits but gibberish number such as 1234567890 and tapped on Update button.

Could you check on this one? Thanks!

davelindqvist:
Closing pull request as @Davit-BetterAngels implemented this with [client profile form validation (DEV-655) #580](https://github.com/BetterAngelsLA/monorepo/pull/580)

PR #556, opened 2024-09-18
Add client ada accommodation backend (DEV-714)

mikefeldberg:
DEV-714


Davit-BetterAngels:
@mikefeldberg maybe missing but can't see adaAccomodationOther field

Davit-BetterAngels (commented):
where in the PRD I need to see like of accommodation OTHER needs to have accommodationOther or not ?

apps/betterangels-backend/clients/enums.py
```diff
@@ -3,6 +3,14 @@
 from django.utils.translation import gettext_lazy as _
```
Davit-BetterAngels, line 3: where in the PRD I need to see like of accommodation OTHER needs to have accommodationOther or not ?

mikefeldberg: it's in the ticket https://betterangels.atlassian.net/browse/DEV-714

> “other” does not include a custom text input

Davit-BetterAngels (approved):

PR #560, opened 2024-09-21
Client multiple phone numbers & primary number (DEV-177)

mikefeldberg:
DEV-177

This PR:
- creates a generic `PhoneNumber` model in `common`
- moves associated phone number types to `common`
- adds a `phone_numbers` field to `ClientProfile`

Follow-up tickets: DEV-751 (fe update & cleanup), DEV-752 (be update & cleanup)


apps/betterangels-backend/clients/schema.py
```diff
@@ -32,31 +32,74 @@
...
-    model_cls = apps.get_model("clients", model_cls_name)
+    model_cls = apps.get_model(module, model_cls_name)
 
     item_updates_by_id = {item["id"]: item for item in data if item.get("id")}
     items_to_create = [item for item in data if not item.get("id")]
-    items_to_update = model_cls.objects.filter(id__in=item_updates_by_id.keys(), client_profile=client_profile)
+    items_to_update = model_cls.objects.filter(id__in=item_updates_by_id.keys())
     model_cls.objects.exclude(id__in=item_updates_by_id).delete()
 
+    if generic_relation:
```
mikefeldberg, line 83: not sure if this is overloading the method and we should split this up differently

apps/betterangels-backend/clients/schema.py
```diff
@@ -32,31 +32,74 @@
     UpdateClientProfileInput,
 )
 
-CLIENT_RELATED_CLS_NAME_BY_RELATED_NAME = {
-    "contacts": "ClientContact",
-    "hmis_profiles": "HmisProfile",
-    "household_members": "ClientHouseholdMember",
-    "social_media_profiles": "SocialMediaProfile",
+CLIENT_RELATED_CLS_NAME_BY_RELATED_NAME: dict[str, dict[str, Any]] = {
```
mikefeldberg, line 35: this dict can be simplified by setting some default values in the mutation. e.g.

```python
    related_cls_map["module"],
    related_cls_map["cls_name"],
    related_cls_map["generic_relation"],
```
-> 
```python
    related_cls_map.get("module", "clients"),
    related_cls_map["cls_name"],
    related_cls_map.get("generic_relation", "False"),
```

apps/betterangels-backend/common/models.py
```diff
@@ -220,6 +221,25 @@ def get_or_create_location(cls, location_data: Dict[str, Any]) -> "Location":
         return location
 
 
+class PhoneNumber(models.Model):
```
vecchp, line 224: Probably need to follow the index pattern similar to the attachment model.  We likely want to have an index on `context_type` and `object_id`

apps/betterangels-backend/common/models.py
```diff
@@ -220,6 +221,25 @@ def get_or_create_location(cls, location_data: Dict[str, Any]) -> "Location":
...
+    number = PhoneNumberField(region="US", blank=True, null=True)
+    is_primary = models.BooleanField(default=False)
+
+    content_type = models.ForeignKey(ContentType, on_delete=models.CASCADE)
+    object_id = models.PositiveIntegerField()
+    content_object: GenericForeignKey = GenericForeignKey("content_type", "object_id")
```
vecchp, line 230: Is this used automatically?

mikefeldberg: not sure I understand the question. Automatically as opposed to needing to be defined/invoked somewhere?

vecchp: Related to previous question, if generic foreign keys create indices then the previous recommendation isn't necessary.

vecchp: See: https://docs.djangoproject.com/en/5.1/ref/contrib/contenttypes/#generic-relations

apps/betterangels-backend/common/graphql/types.py
```diff
@@ -1,11 +1,33 @@
...
+    DjangoPhoneNumber,
+    serialize=lambda v: str(v.national_number),
+    parse_value=lambda v: parse(v, "US"),
+)
+
+
+@strawberry_django.type(PhoneNumber)
+class PhoneNumberType:
+    id: ID
+    number: Optional[PhoneNumberScalar]  # type: ignore
```
vecchp, line 21: What is the type error here, why ignored?

apps/betterangels-backend/clients/schema.py
vecchp: This function is getting pretty big.

apps/betterangels-backend/clients/schema.py
vecchp: Isn't there a way to directly get the cls from the relation?

apps/betterangels-backend/common/graphql/types.py
mikefeldberg: this is the same error we're ignoring elsewhere in [ClientProfileBaseType](https://github.com/BetterAngelsLA/monorepo/blob/ccab1e629b0f44831e4d15823e95bb1241878f1a/apps/betterangels-backend/clients/types.py#L162) and [ClientContactBaseType](https://github.com/BetterAngelsLA/monorepo/blob/ccab1e629b0f44831e4d15823e95bb1241878f1a/apps/betterangels-backend/clients/types.py#L186). It's a known strawberry/mypy issue we ran into when we implemented django-phonenumber-field

apps/betterangels-backend/common/models.py
mikefeldberg: ya, there's no index created. I updated that

vecchp (approved):

PR #576, opened 2024-10-02
remove public note from note card (DEV-779)

Davit-BetterAngels:
DEV-779


johnr54321 (approved):
GIRP text is removed from interaction card on user list as well as client list.

tglaz (approved):

tglaz (approved):

PR #577, opened 2024-10-02
Hide moods (DEV-774)

Davit-BetterAngels:
DEV-774

### DIFF
--- a/apps/betterangels/src/app/(private-screens)/add-note/[noteId].tsx
+++ b/apps/betterangels/src/app/(private-screens)/add-note/[noteId].tsx
@@ -21,7 +21,6 @@ import { useLocalSearchParams, useNavigation, useRouter } from 'expo-router';
 import React, { useLayoutEffect, useMemo, useRef, useState } from 'react';
 import { ScrollView, View } from 'react-native';
 import Location from './Location';
-import Mood from './Mood';
 import NextStep from './NextStep';
 import ProvidedServices from './ProvidedServices';
 import PublicNote from './PublicNote';
@@ -222,10 +221,12 @@ export default function AddNote() {
     );
   };
 
-  const MoodAttachments = useMemo(
-    () => filterAttachments(NoteNamespaceEnum.MoodAssessment),
-    [data]
-  );
+  // TODO: Will be back with moods
+  // const MoodAttachments = useMemo(
+  //   () => filterAttachments(NoteNamespaceEnum.MoodAssessment),
+  //   [data]
+  // );
+
   const RequestedAttachments = useMemo(
     () => filterAttachments(NoteNamespaceEnum.RequestedServices),
     [data]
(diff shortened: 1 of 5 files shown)


johnr54321 (approved):
Moods are hidden on interaction card, interaction form, and interaction summary on ios and android.

apps/betterangels/src/app/(private-screens)/add-note/[noteId].tsx
```diff
@@ -21,7 +21,6 @@ import { useLocalSearchParams, useNavigation, useRouter } from 'expo-router';
 import React, { useLayoutEffect, useMemo, useRef, useState } from 'react';
 import { ScrollView, View } from 'react-native';
 import Location from './Location';
-import Mood from './Mood';
```
Comment (shiozakiyuko) [inline apps/betterangels/src/app/(private-screens)/add-note/[noteId].tsx:24]: Hi @Davit-BetterAngels 

I found one thing.

I confirmed that the Mood section is removed in Edit view and View-only sections. However, I found the mood related content in the Public Note. I believe we need to remove that too when we don't have the Mood section overall.

libs/expo/betterangels/src/lib/screens/Note/NoteServices.tsx
```diff
@@ -100,17 +49,14 @@ const ICONS = [
...
     customService?: string | null | undefined;
   }[]
 ) => {
   return array.slice(0, 4).map((icon) => {
     let iconEnumValue = '';
 
-    if ('descriptor' in icon) {
-      iconEnumValue = icon.descriptor as string;
-    } else if ('service' in icon) {
+    if ('service' in icon) {
```
tglaz, line 59: if want, can get rid of the `if` statement with `let iconEnumValue = icon.service || '';`

apps/betterangels/src/app/(private-screens)/add-note/[noteId].tsx
```diff
@@ -222,10 +221,12 @@ export default function AddNote() {
...
-  const MoodAttachments = useMemo(
-    () => filterAttachments(NoteNamespaceEnum.MoodAssessment),
-    [data]
-  );
+  // TODO: Will be back with moods
+  // const MoodAttachments = useMemo(
+  //   () => filterAttachments(NoteNamespaceEnum.MoodAssessment),
+  //   [data]
+  // );
+
```
Comment (tglaz) [inline apps/betterangels/src/app/(private-screens)/add-note/[noteId].tsx:229]: Not blocking, but typically I'd rather see these just deleted. 
Unless there's a ticket to undo this for this sprint, there's no guarantee these things actually get put back in. Even so, this may come back in some other form etc...

tglaz (approved):
Approved with some optional feedback

tglaz:
@Davit-BetterAngels - think there may be a typo in the JIRA link comment up to - should be `DEV-774` vs 744

apps/betterangels/src/app/(private-screens)/add-note/[noteId].tsx
Comment (Davit-BetterAngels) [reply apps/betterangels/src/app/(private-screens)/add-note/[noteId].tsx:229]: why remove the logic if we are going to bring it back a bit later and then think of implementing same logic ?

I think its better to leave as a comment then just uncomment it. We can discuss it as a parking lot item

apps/betterangels/src/app/(private-screens)/add-note/[noteId].tsx
Comment (shiozakiyuko) [reply apps/betterangels/src/app/(private-screens)/add-note/[noteId].tsx:24]: Hey @Davit-BetterAngels 
I still see the mood in the public note... I removed my dev app once and reinstalled it and tested again but still see the mood.

shiozakiyuko (approved):

PR #596, opened 2024-10-07
Fetching Active Clients using DRF (DEV-716)

CruiseDevice:
(DEV-716)

- Pulls a list of active clients from the backend using Django Rest Framework (DRF)
- Implemented pagination and filtering. Using django-filters package to achieve this

### DIFF
--- a/apps/betterangels-backend/accounts/serializers.py
+++ b/apps/betterangels-backend/accounts/serializers.py
@@ -25,6 +25,8 @@ try:
     from allauth.socialaccount.helpers import complete_social_login
 except ImportError:
     raise ImportError("allauth needs to be added to INSTALLED_APPS.")
+from accounts.models import User
+from clients.models import ClientProfile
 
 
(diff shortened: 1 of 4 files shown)


tglaz:
@CruiseDevice - i looked it over with @vecchp and here's some feedback:

**1. Skip permissions and make it dev only?**
To prove out the concept perhaps we can just keep this restricted to the `development mode`, in which case the REST endpoints would not be too concerned with permissions. 

To do this, could you take a look how we could use Django Waffle to manage this via feature flags perhaps? The goal would be to keep this REST api accessible only in dev env

**2. Permissions**
FYI: These requests do not seem to be verifying permissions in the same way that `ClientProfilePermissions` are checked in apps/betterangels-backend/clients/schema.py (for example). May not need to do anything if we can keep this `dev only`

Here's a chatgpt thread with a simple example of how to do that with DRF: https://chatgpt.com/share/6706d183-5d8c-8008-ac34-61a9a9ab2ff2

tglaz (changes requested):

PR #599, opened 2024-10-10
add .env.local to gitignore

tglaz:
https://betterangels.atlassian.net/browse/DEV-860

* Allows using a `.env.local` file in local repos.
  * this file takes precendence over `.env` file values
* To bypass CORS issues, on web at least, you can:
  *  add a `.env.local` file
  * set `EXPO_PUBLIC_API_URL=http://localhost:8000`

### DIFF
--- a/docs/frontend_readme.md
+++ b/docs/frontend_readme.md
@@ -91,6 +91,9 @@ Run the following on the host machine (not in the container):
 
 1. Optional: add a `.env.local` file:
    - See `.env.local.sample` file for refernce
+   - .env.local file should override the .env values on your local machine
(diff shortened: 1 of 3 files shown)


docs/frontend_readme.md
```diff
@@ -85,12 +89,15 @@ Run the following on the host machine (not in the container):
    nvm install node && nvm alias default node
    ```
 
+1. Optional: add a `.env.local` file:
```
vecchp, line 92: Maybe add an explanation as to why.  Like if you'd like to override environment variables and then point them to the NX docs for further information? https://nx.dev/recipes/tips-n-tricks/define-environment-variables

vecchp (approved):

PR #602, opened 2024-10-12
DEV-205 add snackbar component

tglaz:
**Fix frontend error handling**

Creates `Snackbar` component with Provider, context and hook.
* component itself wraps the Snackbar component from `react-native-paper`, so that can be swapped in future
* since we don't have any specs for this feature, and initially will be used to surface backend errors, the styling is a bit raw (some screenshots below)

**Note**: the implementation will follow in a separate PR

Screenshot 2024-10-12 at 12 39 49 PM

Screenshot 2024-10-12 at 12 41 22 PM

Screenshot 2024-10-12 at 12 40 14 PM


davelindqvist (approved):
LGTM!

libs/expo/shared/ui-components/src/lib/Snackbar/Snackbar.tsx
```diff
@@ -0,0 +1,63 @@
...
+  if (type === 'success') {
+    return {
+      backgroundColor: Colors.SUCCESS,
+    };
+  }
+
+  return {
+    backgroundColor: Colors.PRIMARY,
+  };
+}
```
Davit-BetterAngels, line 63: Suggested Change: 

`interface ISnackbar extends SnackbarProps {
  type?: TSnackbarType;
  onActionPress?: () => void;
  actionLabel?: string;
}

const typeStyles = {
  error: {
    backgroundColor: Colors.ERROR,
  },
  success: {
    backgroundColor: Colors.SUCCESS,
  },
  default: {
    backgroundColor: Colors.PRIMARY,
  },
};

export function Snackbar(props: ISnackbar): ReactElement {
  const { style, onActionPress, type, actionLabel = 'Close', ...rest } = props;

  return (
    <RnpSnackbar
      style={[style, typeStyles[type || 'default']]}
      action={{
        label: actionLabel,
        onPress: onActionPress,
      }}
      {...rest}
    />
  );
}`

Davit-BetterAngels (changes requested):

libs/expo/shared/ui-components/src/lib/Snackbar/Snackbar.tsx
tglaz: Updated. FYI - the reason for the original interface design was to fully separate it from the react-paper one, and only expose what we allow to control. If we swapped out for another one then `Snackbar` interface wouldn't change. But in this case probably doesn't matter

Davit-BetterAngels (approved):

PR #603, opened 2024-10-14
Make purpose optional for now (DEV-868)

mikefeldberg:
DEV-868

I should have done this when I added `purpose` 🤦  in https://github.com/BetterAngelsLA/monorepo/pull/583


apps/betterangels-backend/notes/models.py
```diff
@@ -196,7 +196,10 @@ class Note(BaseModel):
     _private_details: Optional[str]
 
     def __str__(self) -> str:
-        return self.purpose
+        if self.purpose:
```
vecchp, line 199: In code golf this can be

```python
def __str__(self) -> str:
    return self.purpose or str(self.id)
```

mikefeldberg: yea, I got a mypy error for that but it looks like it was a false positive cus it's gone after restart

vecchp (approved):

PR #605, opened 2024-10-14
show errors in snackbar on note and client forms

tglaz:
Fix frontend error handling - implementation of `snackbar` component

This is an implementation follow up to PR  #602

**Note:** this PR uses `snackbar` to display mostly generic errors to the user in the UI. There are probably other places we want to to show the user that `oops, something work`, but may need to handle that on case by case basis? 
If you know of specific places to add it then let me know.


tglaz:
closing - will reopen after i fix conflicts

PR #609, opened 2024-10-16
Dev 821/keyboard done button

tglaz:
### [Clean Up] The keyboard (alphabet and numeric) should have "Done" button on the top of the keyboard
https://betterangels.atlassian.net/browse/DEV-821

* shows `KeyboardToolbar` via `KeyboardToolbarProvider`
  * accepts options to show/hide arrows in keyboard
* updated `MainScrollContainer` to a use `KeyboardAwareScrollView` optionally
  * we can make this change global, using KeyboardAwareScrollView always, though may have some performance impact (not sure how much)

**Note**: I misread the requirements initially and added extra functionality to the provider. And so had to add onFocus prop to the `Input` components. I could undo it, but figured it may stay there.


libs/expo/betterangels/src/lib/providers/keyboardToolbar/keyboardToolbarProvider.tsx
```diff
@@ -0,0 +1,33 @@
+import { ReactNode, useState } from 'react';
+import { KeyboardToolbar } from 'react-native-keyboard-controller';
+
+import { KeyboardToolbarContext } from './keyboardToolbarContext';
+
+type TKeyboardToolbarProvider = {
+  children: ReactNode;
+};
+
+export default function KeyboardToolbarProvider(
```
Davit-BetterAngels, line 10: Not sure why do we need context. Either app should have arrow or not at all.

Davit-BetterAngels (changes requested):
Not sure why do we need context. Either app should have the arrows or not at all.

libs/expo/betterangels/src/lib/providers/keyboardToolbar/keyboardToolbarProvider.tsx
tglaz: So I kind of made a mistake and misread the ACs at first.
I coded it as if we'd need to use it conditionally. I can easily remove the context, but if we end up wanting to show the arrows then I'm not sure we'd want to do it universally, cause when you mix inputs and checkboxes on same form, for example, then acts weird. And so I figured if we ever want to customize the toolbar in any way then we'd need context, so I kept it.
Our of scope for this, I realize, so I can remove it.

Davit-BetterAngels (approved):

shiozakiyuko (approved):

PR #611, opened 2024-10-17
Pill Component (DEV-844)

Davit-BetterAngels:
DEV-844


libs/expo/shared/ui-components/src/lib/Pill/Pill.tsx
```diff
@@ -0,0 +1,38 @@
+import { Colors, Radiuses, Spacings } from '@monorepo/expo/shared/static';
+import { StyleSheet, View } from 'react-native';
+import TextRegular from '../TextRegular';
+
+interface IPillProps {
+  color?: 'green' | 'blue';
```
tglaz, line 6: Think it'd be safer to use `type` instead of color here?
The types could remain `blue` or `green` or perhaps be more generic, like `primary`, secondary` etc. 
It's a simple component so not a huge deal, but may be simpler to maintain should colors change etc

Davit-BetterAngels: any suggestion what to call the types? 

couldn't understand which one to put 'primary' or secondary etc.

tglaz: In the mocks it states that that `green` is default, so maybe that one. 
But that was an arbitrary decision I think. 

I'd be ok with either for now, but I typically prefer components with more flexible APIs that we can then display in `storybook` - which would be the `toolbox` that product would use to create features. Can start very simple, but should be extendible.

example Pill component api:
https://instructure.design/#Pill

Davit-BetterAngels: I don't like adding more props because most of them become redundant. 
Just adding per requirement. 

I would name the type as color names as they match. Primary, success etc.

tglaz (approved):

PR #613, opened 2024-10-17
Add team to note (interaction) model (DEV-835)

mikefeldberg:
DEV-835


apps/betterangels-backend/notes/models.py
```diff
@@ -181,6 +182,7 @@ class Note(BaseModel):
     purpose = models.CharField(max_length=100, null=True, blank=True)
     purposes = models.ManyToManyField(Task, blank=True, related_name="purpose_notes")
     requested_services = models.ManyToManyField(ServiceRequest, blank=True, related_name="requested_notes")
+    team = TextChoicesField(SelahTeamEnum, null=True, blank=True)
```
tglaz, line 185: I believe we agreed to just throw `team` in there for now, so ok.
I'd love to use a different name and model this differently but think we just have to eat this one 😢

tglaz (approved):

apps/betterangels-backend/notes/models.py
mikefeldberg: let's chat about this when you have time? https://betterangels.atlassian.net/browse/DEV-875

PR #615, opened 2024-10-17
show snackbar messages for node and client form errors

tglaz:


Davit-BetterAngels (approved):

apps/betterangels/src/app/(private-screens)/add-note/[noteId].tsx
```diff
@@ -5,6 +5,7 @@ import {
   Ordering,
   useDeleteNoteMutation,
   useRevertNoteMutation,
+  useSnackbar,
```
Comment (shiozakiyuko) [inline apps/betterangels/src/app/(private-screens)/add-note/[noteId].tsx:8]: Hi @tglaz 

Thanks for the repro steps! It worked and I confirmed I saw the error snackbar. Thanks so much! 
I can approve it but I'm wondering the specific words that I see currently can be modified later on, correct?

shiozakiyuko (approved):

PR #617, opened 2024-10-21
Dev 822/hmis id validate

tglaz:
https://betterangels.atlassian.net/browse/DEV-822

* Error message if either one of the type or HMIS ID number field are not filled but the other is, `Enter HMIS ID or remove this entry`

### DIFF
--- a/libs/expo/betterangels/src/lib/screens/AddEditClient/PersonalInfo/HmisProfiles.tsx
+++ b/libs/expo/betterangels/src/lib/screens/AddEditClient/PersonalInfo/HmisProfiles.tsx
@@ -5,6 +5,7 @@ import {
   Radio,
   TextBold,
   TextButton,
+  TextRegular,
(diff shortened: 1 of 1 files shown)


libs/expo/betterangels/src/lib/screens/AddEditClient/PersonalInfo/HmisProfiles.tsx
```diff
@@ -5,9 +5,10 @@ import {
   Radio,
```
shiozakiyuko, line 5: Hi @tglaz 

I confirmed I see the error message when I tried to submit without the HMIS ID#. 
I read the comment in the JIRA ticket regarding the AC#1. When are you planning to fix the part? 
Let me know! Thanks!

shiozakiyuko (approved):

Davit-BetterAngels (approved):

PR #627, opened 2024-10-24
updating interaction redirects to client profile (DEV-831)

mikefeldberg:
DEV-831

## Summary by Sourcery

Enhancements:
- Update GraphQL mutations to include client profile information.

## Summary by Sourcery

Enhancements:
- Update GraphQL queries to include client profile information for better data retrieval.


apps/betterangels/src/app/(private-screens)/add-note/[noteId].tsx
```diff
@@ -210,10 +210,7 @@ export default function AddNote() {
         throw new Error(`Failed to update interaction: ${updateError}`);
       }
 
-      if (revertBeforeTimestamp) {
-        return router.replace('/');
-      }
-      setSubmitted(true);
+      return router.replace(`/client/${data?.note.client?.clientProfile?.pk}`);
```
Comment (tglaz) [inline apps/betterangels/src/app/(private-screens)/add-note/[noteId].tsx:213]: seems like `clientProfile.pk` should not be optional here?
and if it is then we should handle the case when it's missing

apps/betterangels/src/app/(private-screens)/add-note/[noteId].tsx
```diff
@@ -329,7 +330,7 @@ export default function AddNote() {
         firstName={data.note.client?.firstName}
         closeModal={() => {
           setSubmitted(false);
-          router.navigate('/');
+          router.navigate(`/client/${data.note.client?.clientProfile?.pk}`);
```
Comment (tglaz) [inline apps/betterangels/src/app/(private-screens)/add-note/[noteId].tsx:333]: same comment as above, that either the `pk` is there or isn't

libs/expo/betterangels/src/lib/apollo/graphql/queries.ts
```diff
@@ -101,6 +101,9 @@ export const GET_NOTE = gql`
         email
         firstName
         lastName
+        clientProfile {
+          pk
```
tglaz, line 105: [not blocking] just a question (for another day maybe), but should we rename  pk to id?

tglaz (changes requested):

apps/betterangels/src/app/(private-screens)/add-note/[noteId].tsx
Comment (mikefeldberg) [reply apps/betterangels/src/app/(private-screens)/add-note/[noteId].tsx:217]: iiuc, the reason it's optional is because, while all clientprofiles have a user, only some users have a clientprofile

Comment (mikefeldberg) [reply apps/betterangels/src/app/(private-screens)/add-note/[noteId].tsx:217]: re: `getClientProfileUrl()`, sure thing. where should that live?

libs/expo/betterangels/src/lib/apollo/graphql/queries.ts
mikefeldberg: aliased `pk` to `id` for consistency

apps/betterangels/src/app/(private-screens)/add-note/[noteId].tsx
Comment (tglaz) [reply apps/betterangels/src/app/(private-screens)/add-note/[noteId].tsx]: sure. But my question is what happens when we navigate to `/client/undefined`?

shiozakiyuko (approved):
It looks good to me!

apps/betterangels/src/app/(private-screens)/add-note/[noteId].tsx
Comment (mikefeldberg) [reply apps/betterangels/src/app/(private-screens)/add-note/[noteId].tsx]: sounds important
```
  const getClientProfileUrl = (clientProfileId: string | undefined) =>
    clientProfileId ? `/client/${clientProfileId}` : '/';
```

tglaz (approved):

PR #647, opened 2024-10-28
Make client results and no-results mutually exclusive (DEV-916)

mikefeldberg:
DEV-916

## Summary by Sourcery

Bug Fixes:
- Ensure that the no-results message and the results list are mutually exclusive in the Clients screen.

### DIFF
--- a/libs/expo/betterangels/src/lib/screens/Clients/index.tsx
+++ b/libs/expo/betterangels/src/lib/screens/Clients/index.tsx
@@ -187,7 +187,7 @@ export default function Clients({ Logo }: { Logo: ElementType }) {
             setClients({});
           }}
         />
-        {search && !loading && sections.length < 1 && (
+        {search && !loading && sections.length < 1 ? (
           <View
             style={{
               flexGrow: 1,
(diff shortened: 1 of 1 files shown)


shiozakiyuko (approved):

libs/expo/betterangels/src/lib/screens/Clients/index.tsx
```diff
@@ -187,7 +187,7 @@ export default function Clients({ Logo }: { Logo: ElementType }) {
             setClients({});
           }}
         />
-        {search && !loading && sections.length < 1 && (
+        {search && !loading && sections.length < 1 ? (
```
tglaz, line 190: I kind of have issues with long and nested `ternary` expressions. I won't block on this, but it seems pretty difficult to read especially as the ternary has its own nested ternarys.

Maybe try something like this.

1. define `const` for `hasResults` (or `visibleSearchResults` or ....)


```javascript
const hasResults = search && !loading && !!sections.length; // or however we define it
```

2.  Could create separate component for no-results: `< NoSearchResultsComponent />`? Could keep it in same folder and specific to this screen for now, though eventually can have a generic "no results found" display.

3.  Render based on `hasResults`

```javascript

{hasResults && (
    <SearchResultsComponent />
)}

{!hasResults && (
    <NoSearchResultsComponent />
)}
```

tglaz (changes requested):

libs/expo/betterangels/src/lib/screens/Clients/index.tsx
mikefeldberg: replaced the dirty ternary. added tickets to [create](https://betterangels.atlassian.net/browse/DEV-942) and [implement](https://betterangels.atlassian.net/browse/DEV-943) reusable "no results" component

libs/expo/betterangels/src/lib/screens/Clients/index.tsx
```diff
@@ -155,6 +155,7 @@ export default function Clients({ Logo }: { Logo: ElementType }) {
   }, [data, offset]);
 
   const sections = useMemo(() => Object.values(clients || {}), [clients]);
+  const hasClients = !loading && !!sections.length;
```
tglaz, line 158: ❤️

tglaz (approved):

PR #648, opened 2024-10-28
[Demo Only] Picker demo 

tglaz:
### Demo of Picker on Preferred Language form.

Notes: 
* can test on Client -> Personal Info -> Preferred Language
* based on [react-native-picker/picker](https://github.com/react-native-picker/picker?tab=readme-ov-file#react-native-pickerpicker)
* Would still need:
  * test on Android
  * styling
  * UX design
    * probably want to show this Picker component only if a user clicks something
    * probably want to keep it always at bottom of screen

### Demo screenshot

image


vecchp:
How did we install this package?

```
 expo install @react-native-picker/picker
```
Did we run the above command?

tglaz:
@vecchp - no, did a regular yarn install with sync-deps and all that fun stuff.

PR #654, opened 2024-10-30
Fix related field form sections (DEV-926)

Davit-BetterAngels:
DEV-926

initial bug this ticket says is normal behavior for the state. you are removing and adding same thing. We are using react hook form. And its state is handled to keep data for the performance. 

What Mike is saying fixed in the PR for this ticket. we can discuss this more

## Summary by Sourcery

Fix the bug related to contact removal in the RelevantContacts component by ensuring that removing a contact does not bring in the next contact's data. Enhance the useFieldArray hook by introducing a unique keyName 'fieldId' for better field management.

Bug Fixes:
- Fix the issue where removing a contact would inadvertently bring in the next contact's data.

Enhancements:
- Use a unique keyName 'fieldId' for contacts in the useFieldArray hook to ensure proper identification and handling of contact fields.


mikefeldberg (changes requested):
the behavior is slightly different but still not correct
contacts_dont_clear_5

libs/expo/betterangels/src/lib/screens/AddEditClient/RelevantContacts/Contact/index.tsx
```diff
@@ -59,7 +59,7 @@ export default function Contact(props: IContactProps) {
       null
```
shiozakiyuko, line 59: Hi @Davit-BetterAngels 

It worked great on iOS. 
As for android, the loading takes forever and I can't check with this device because of it... I will let you know when it's done.

shiozakiyuko: I was finally able to test the Android, and it worked good to me. Btw, I didn't experience what Mike experienced above (video)

mikefeldberg (approved):

PR #655, opened 2024-10-30
Refactor upload button components; add plus icon (DEV-945)

mikefeldberg:
DEV-945 ... sorta. I was refactoring the upload buttons and threw in the plus. The icon is the wrong size, because we have an 8 and a 16 but not a 12 (which is what this needs to be). DEV-945 isn't in the sprint, so I figure we can use this for now and update the size later.

## Summary by Sourcery

Refactor the upload button components by creating a reusable FileUploadTab component, which simplifies the code structure and enhances maintainability. Add a new plus icon to improve the user interface of the upload buttons.

New Features:
- Introduce a new plus icon for use in the upload button components.

Enhancements:
- Refactor the upload button components by introducing a reusable FileUploadTab component to simplify and streamline the code.

### DIFF
--- a/libs/expo/betterangels/src/lib/screens/Client/Docs/UploadModal/FileUploadTab.tsx
+++ b/libs/expo/betterangels/src/lib/screens/Client/Docs/UploadModal/FileUploadTab.tsx
@@ -1,0 +1,68 @@
+import { CheckIcon, PlusIcon } from '@monorepo/expo/shared/icons';
+import { Colors, Radiuses, Spacings } from '@monorepo/expo/shared/static';
+import { Button } from '@monorepo/expo/shared/ui-components';
+import { StyleSheet, View } from 'react-native';
+import { Docs, ITab } from './types';
+
+export interface IFileUploadTabProps {
+  docs: Docs;
+  isSingleUpload: boolean;
+  setTab: (tabKey: ITab) => void;
+  tabKey: string;
+  title: string;
+}
+
+const FileUploadTab = (props: IFileUploadTabProps) => {
+  const { docs, isSingleUpload, setTab, tabKey, title } = props;
+  const disabled = isSingleUpload && !!docs[tabKey as keyof Docs];
+
+  return (
+    <View
+      style={{
+        flexDirection: 'row',
+        alignItems: 'center',
+      }}
+    >
+      <View
+        style={[
+          styles.tabIcon,
+          {
+            backgroundColor: disabled ? Colors.SUCCESS : Colors.WHITE,
+            borderColor: Colors.NEUTRAL_LIGHT,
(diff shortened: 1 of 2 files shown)


shiozakiyuko (commented):
It looks good to me! FYI I read and understood about the icon. I believe the right icon size is coming later.

libs/expo/betterangels/src/lib/screens/Client/Docs/UploadModal/index.tsx
```diff
@@ -239,345 +236,77 @@ export default function UploadModal(props: IUploadModalProps) {
...
-                  height="md"
-                  weight="regular"
-                  align="flex-start"
-                  size="full"
-                  variant="secondary"
-                  title="SSN"
-                  accessibilityHint="opens Social Security Card upload screen"
-                />
-              </View>
+              <FileUploadTab
```
tglaz, line 239: ❤️

tglaz (approved):

libs/expo/betterangels/src/lib/screens/Client/Docs/UploadModal/FileUploadTab.tsx
```diff
@@ -0,0 +1,70 @@
...
+  docs: Docs;
+  setTab: (tabKey: ITab) => void;
+  tabKey: keyof Docs;
+  title: string;
+}
+
+const FileUploadTab = (props: IFileUploadTabProps) => {
+  const { docs, setTab, tabKey, title } = props;
+  const isMultiUpload = Array.isArray(docs[tabKey as keyof Docs]);
+  const disabled = !isMultiUpload && !!docs[tabKey as keyof Docs];
```
tglaz, line 17: Perhaps also define what `docs[tabKey as keyof Docs` is?
Thinking something like

```javascript
  const uploadValue = docs[tabKey as keyof Docs];

  const isMultiUpload = Array.isArray(uploadValue);
  const disabled = !isMultiUpload && !!uploadValue];
```

mikefeldberg: my threshold for storing something in a var is: (1) used more than twice; OR (2) used more than once and far apart from eachother. I'm not committed to this approach, so I can be convinced otherwise

tglaz: sounds good. one other consideration I have is that if it's not immediately clear what something like `docs[tabKey as keyof Docs]` is, then it's helpful to define it with a const.

mikefeldberg: 🤔 yea that's a good one

PR #656, opened 2024-10-30
Add california_id to ClientProfile (DEV-907)

mikefeldberg:
DEV-907

## Summary by Sourcery

Add a 'california_id' field to the ClientProfile model, update the GraphQL schema and input types to include this field, and implement corresponding tests to validate its functionality.

New Features:
- Introduce a new 'california_id' field to the ClientProfile model, which is a unique, optional, 8-character string validated to match a pattern of one letter followed by seven numbers.

Tests:
- Add tests for creating and updating the ClientProfile with the new 'california_id' field to ensure proper functionality and validation.

PR #657, opened 2024-10-30
Add california_id to client search (DEV-896)

mikefeldberg:
DEV-896

note: don't merge until #656

## Summary by Sourcery

Enhance the client search functionality by adding the ability to search for client profiles using a California ID. Update the client search input to support new optional fields for California ID and excluded client profile ID. Add corresponding test cases to ensure the new search capability works as expected.

New Features:
- Add support for searching client profiles by California ID in the client search functionality.

Enhancements:
- Refactor the client search input to include optional fields for California ID and excluded client profile ID.

Tests:
- Introduce new test cases to verify the client search functionality by California ID, including scenarios with and without matches.


apps/betterangels-backend/clients/tests/test_queries.py
```diff
@@ -257,6 +257,40 @@ def test_client_profiles_query_search_client(
         client_profiles = response["data"]["clientProfiles"]
         self.assertEqual(len(client_profiles), expected_client_profile_count)
 
+    @parametrize(
+        ("california_id, excluded_client_profile, expected_client_profile_count"),
+        [
+            ("X0000000", None, 0),  # no match
+            ("L1234567", None, 1),  # match exists
+            ("L1234567", "client_profile_1", 0),  # match exists only on current client
```
tglaz, line 265: perhaps could we add a few other test cases, such as:
* with non-existent `excludedClientProfileId`
* with partial match californiaId `L123`
* with invalid californiaId `Lxxx`

or maybe can create a ticket for a generic `test_client_profiles_query_search_client` to test out common things like `no filters` or exhaustive tests for exact vs partial matches and stuff like that. 
Not sure we necessarily need exhaustive tests for each search param, other than our lists of user_fields vs client fields are correct.

apps/betterangels-backend/clients/types.py
```diff
@@ -124,11 +127,19 @@ def search_client(
         """
         Returns client profiles with exact match on all provided search fields (case insensitive).
         All search fields are optional.
+
+        Accepts an excluded_client_profile_id param, to prevent the client profile being edited
+        from being flagged as a duplicate entry.
```
tglaz, line 132: I'd suggest removing references to duplicates as this is just a generic search. 
Maybe something like this is enough: `Accepts an excluded_client_profile_id param to exclude from result set.`

tglaz (approved):

PR #660, opened 2024-10-30
Added description to HMIS ID section (DEV-862)

telishreyas10:
[DEV-862](https://betterangels.atlassian.net/browse/DEV-862)

## Summary by Sourcery

Enhancements:
- Add a subtitle to the HMIS IDs section to guide users to fill in both HMIS ID Type and ID # when profiles are present.

[DEV-862]: https://betterangels.atlassian.net/browse/DEV-862?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #661, opened 2024-10-30
Added description to HMIS ID section (DEV-862)

mikefeldberg:
DEV-862

depends on #660

## Summary by Sourcery

Enhancements:
- Add a subtitle to the HMIS IDs section to guide users on filling in both HMIS ID Type and ID #.

### DIFF
--- a/libs/expo/betterangels/src/lib/screens/AddEditClient/PersonalInfo/HmisProfiles.tsx
+++ b/libs/expo/betterangels/src/lib/screens/AddEditClient/PersonalInfo/HmisProfiles.tsx
@@ -39,6 +39,9 @@ export default function HmisProfiles() {
     <CardWrapper
       {...(hmisProfiles.length !== 0 && { onReset })}
       title="HMIS IDs"
+      subtitle={
+        hmisProfiles.length !== 0 ? 'Fill in both HMIS ID Type and ID #' : ''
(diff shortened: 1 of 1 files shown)


libs/expo/betterangels/src/lib/screens/AddEditClient/PersonalInfo/HmisProfiles.tsx
```diff
@@ -39,6 +39,9 @@ export default function HmisProfiles() {
     <CardWrapper
       {...(hmisProfiles.length !== 0 && { onReset })}
       title="HMIS IDs"
+      subtitle={
+        hmisProfiles.length !== 0 ? 'Fill in both HMIS ID Type and ID #' : ''
```
tglaz, line 43: `!==` compares are not very fun for me, even if necessary sometimes. would prefer:

```javascript
hmisProfiles.length ?  '': 'Fill in both HMIS ID Type and ID #' 
```
or 
```javascript
!!hmisProfiles.length ? 'Fill in both HMIS ID Type and ID #' : ''
```
but FWIW my preference is to define logic in function body, even if in this case it's pretty straight forward

tglaz:
Closing - replaced with #683 
- replaced branch to avoid conflicts

PR #672, opened 2024-11-01
Styling on Important notes Textarea now aligned vertically flex-start

eddiearnoldz:
- On Android the text area text and placeholder appeared centred in the Textarea component (not on iOS)
- Added the textAlignVertical='top' attribute to textArea, resolved issue on Android. No change on iOS.

## Summary by Sourcery

Bug Fixes:
- Fix vertical alignment of text and placeholder in the Textarea component on Android by setting textAlignVertical to 'top'.

### DIFF
--- a/libs/expo/betterangels/src/lib/screens/AddEditClient/ImportantNotes/index.tsx
+++ b/libs/expo/betterangels/src/lib/screens/AddEditClient/ImportantNotes/index.tsx
@@ -40,6 +40,7 @@ export default function ImportantNotes(props: IImportantNotesProps) {
           }}
         >
           <Textarea
+            textAlignVertical='top'
             placeholder="Enter important notes"
             name="importantNotes"
             control={control}
--- a/libs/expo/betterangels/src/lib/screens/AddEditClient/ImportantNotes/index.tsx
+++ b/libs/expo/betterangels/src/lib/screens/AddEditClient/ImportantNotes/index.tsx
@@ -40,7 +40,6 @@ export default function ImportantNotes(props: IImportantNotesProps) {
           }}
         >
           <Textarea
-            textAlignVertical='top'
             placeholder="Enter important notes"
             name="importantNotes"
             control={control}
--- a/libs/expo/shared/ui-components/src/lib/Textarea/Textarea.tsx
+++ b/libs/expo/shared/ui-components/src/lib/Textarea/Textarea.tsx
@@ -52,6 +52,7 @@ interface ITextareaProps extends TextInputProps {
   ml?: TSpacing;
   mr?: TSpacing;
   height?: number | 'auto';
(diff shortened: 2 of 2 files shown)


libs/expo/betterangels/src/lib/screens/AddEditClient/ImportantNotes/index.tsx
```diff
@@ -40,6 +40,7 @@ export default function ImportantNotes(props: IImportantNotesProps) {
           }}
         >
           <Textarea
+            textAlignVertical='top'
```
tglaz, line 43: Maybe we could we set this on all `Textarea` components as a default for consistency?
As long as we could still override it when needed.

tglaz (changes requested):

eddiearnoldz:
I have made this the default behaviour now for all TextArea components, thanks Tom

libs/expo/shared/ui-components/src/lib/Textarea/Textarea.tsx
```diff
@@ -52,6 +52,7 @@ interface ITextareaProps extends TextInputProps {
   ml?: TSpacing;
   mr?: TSpacing;
   height?: number | 'auto';
+  textAlignVertical?: 'top' | 'center' | 'bottom';
```
tglaz, line 55: i don't believe you'd need this line since  ITextareaProps already extends `TextInputProps`

libs/expo/shared/ui-components/src/lib/Textarea/Textarea.tsx
```diff
@@ -121,6 +123,7 @@ export function Textarea(props: ITextareaProps) {
                 fontFamily: 'Poppins-Regular',
                 fontSize: FontSizes.md.fontSize,
                 lineHeight: FontSizes.md.lineHeight,
+                textAlignVertical: textAlignVertical,
```
tglaz, line 126: It looks like `textAlignVertical` is also accepted as a style? I thought it'd be used as a `TextInput` prop (not a style).
Is it the same in both cases?

tglaz (changes requested):

libs/expo/shared/ui-components/src/lib/Textarea/Textarea.tsx
eddiearnoldz: Hey Tom, thanks for questioning this. It works on both after testing but I moved to be its own property as is expected based on how its used from the component itself.

libs/expo/shared/ui-components/src/lib/Textarea/Textarea.tsx
eddiearnoldz: Tested this and yes its redundant, thanks for the feedback. Have removed

tglaz (approved):
👍

PR #681, opened 2024-11-05
Set cleared input fields to null (DEV-973)

mikefeldberg:
DEV-973

Before: 

After:
[insert profile successfully updated with empty email here]

## Summary by Sourcery

Bug Fixes:
- Set cleared input fields to null instead of an empty string in Input and Textarea components.


tglaz:
@mikefeldberg  - left a comment on JIRA ticket cause perhaps i'm not fully understanding the issue.

But if we need to send `null`, we also have code like this below in many places. 

```javascript
  const onReset = () => {
    setValue('user.firstName', '');
    setValue('user.middleName', '');
    setValue('user.lastName', '');
    setValue('nickname', '');
  };
```

mikefeldberg:
> @mikefeldberg - left a comment on JIRA ticket cause perhaps i'm not fully understanding the issue.
> 
> But if we need to send `null`, we also have code like this below in many places.
> 
> ```js
>   const onReset = () => {
>     setValue('user.firstName', '');
>     setValue('user.middleName', '');
>     setValue('user.lastName', '');
>     setValue('nickname', '');
>   };
> ```

good point. We should fix all of those, too, at some point. I updated the ticket to only address this email field

libs/expo/shared/ui-components/src/lib/Input/Input.tsx
```diff
@@ -149,7 +149,7 @@ export function Input(props: IInputProps) {
                 accessibilityRole="button"
                 accessibilityLabel="delete icon"
                 accessibilityHint="deletes input's value"
-                onPress={() => onChange('')}
+                onPress={() => onChange(null)}
```
tglaz, line 152: I'd rather avoid doing this globally as may have unexpected consequences (types aren't checked here and I'm not sure what setting values to `null` on any single field would do).

If the server really can't accept an empty string for `email`, then perhaps may be safer to cast empty string to `null` somewhere in the [AddEditClient/index.tsx](https://github.com/BetterAngelsLA/monorepo/blob/dd7bd55be46db37a6735dba763c5b678a4587d9e/libs/expo/betterangels/src/lib/screens/AddEditClient/index.tsx#L156)

would something like this work?
don't remember if we have other email fields (for other contacts maybe) that  would need a similar thing.

```javascript
    if (typeof values.user?.email === 'string') {
      values.user.email = values.user.email || null;
    }
```

tglaz (changes requested):

libs/expo/shared/ui-components/src/lib/Input/Input.tsx
mikefeldberg: that's a good point. I think your idea will work

tglaz (approved):

PR #688, opened 2024-11-06
Remove deprecated `hmis_id` (DEV-524)

aaronipollock:
Left a few that seemed like they might cause unwanted (at the moment) UI changes.

## Summary by Sourcery

Remove the 'hmis_id' field from the HmisProfile and ClientProfile models, including associated constraints, database migrations, and update related tests and GraphQL queries.

Enhancements:
- Remove the 'hmis_id' field from the HmisProfile and ClientProfile models, along with associated constraints and database migrations.

Tests:
- Update tests to reflect the removal of the 'hmis_id' field from the HmisProfile and ClientProfile models.

Ticket: DEV-524


apps/betterangels-backend/accounts/migrations/0012_clientprofile_hmis_id.py
```diff
@@ -11,7 +11,6 @@ class Migration(migrations.Migration):
     operations = [
         migrations.AddField(
             model_name="clientprofile",
-            name="hmis_id",
```
vecchp, line 14: Typically, when removing fields in Django, you don’t modify existing migrations. Instead, keep the old migrations intact and create a new migration to handle the field removal.

apps/betterangels-backend/clients/tests/utils.py
```diff
@@ -71,7 +70,6 @@ def setUp(self) -> None:
             }
             hmisProfiles {
                 id
-                hmisId
```
vecchp, line 74: I don’t think we want to remove this field—this is removing hmisId from the HmisProfile, which I believe is part of the new method. @mikefeldberg, can you confirm?

apps/betterangels-backend/clients/types.py
```diff
@@ -113,7 +112,6 @@ def search(
 @strawberry_django.type(HmisProfile)
 class HmisProfileType:
     id: auto
-    hmis_id: auto
```
vecchp, line 116: Same comment as above

vecchp (changes requested):

PR #698, opened 2024-11-11
DEV-733 require any one of the four client name fields

tglaz:
### Client Profile - add validation to form name section
https://betterangels.atlassian.net/browse/DEV-733

## Summary by Sourcery

Bug Fixes:
- Add validation to ensure at least one of the client name fields is filled out in the client profile form.


libs/expo/betterangels/src/lib/screens/AddEditClient/PersonalInfo/FullName.tsx
```diff
@@ -1,26 +1,58 @@
...
+} from '@monorepo/expo/shared/ui-components';
+import { useEffect } from 'react';
+import { useFormContext, useWatch } from 'react-hook-form';
 import { Text } from 'react-native';
 import {
   CreateClientProfileInput,
   UpdateClientProfileInput,
 } from '../../../apollo';
 
+type TFullNameForm = (UpdateClientProfileInput | CreateClientProfileInput) & {
```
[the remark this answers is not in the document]
tglaz: we have `zod` installed but not using it, so will ignore for now.
when we migrate this to it's own screen and own form then yup!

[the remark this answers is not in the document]
tglaz: as for the display, the text differs, with the parenthesis, and will just move on

libs/expo/betterangels/src/lib/screens/AddEditClient/PersonalInfo/FullName.tsx
```diff
@@ -1,26 +1,58 @@
...
+      clearErrors('emptyFormError');
+
+      return;
+    }
+
+    setError('emptyFormError', {
+      type: 'manual',
+      message: 'Filling out one of the fields is required',
+    });
+  }, [firstName, middleName, lastName, nickname, setError, clearErrors]);
```
Davit-BetterAngels, line 54: why not input level error check rather than in use effect? I would say just create a rule required if other values are not there

Davit-BetterAngels (changes requested):

libs/expo/betterangels/src/lib/screens/AddEditClient/PersonalInfo/FullName.tsx
tglaz: I looked into it and still still having some issues, so let's reconnect tomorrow on this. 
But here are my thoughts:

1. I couldn't make it work properly, so that's the big one lol. I agree with you that this approach should work, so maybe we can pair tomorrow?
2. `Filling out one of the fields is required` scenario  is not super common, so seems ok to treat it as a one-off, but it still feels a bit strange to put a `required` rule on an input when that specific input isn't really "required" (only any 1 out of 4 is required). Not huge of an issue since it's a one-off, but 🤷

Davit-BetterAngels (approved):

libs/expo/betterangels/src/lib/screens/AddEditClient/PersonalInfo/FullName.tsx
```diff
@@ -1,26 +1,64 @@
 import { Colors } from '@monorepo/expo/shared/static';
-import { CardWrapper, Input } from '@monorepo/expo/shared/ui-components';
+import {
```
shiozakiyuko, line 2: Hi @tglaz 

I have a couple of comments: 

1. Can you reduce the spacing between the title "Full Name" and "Filling out one of the..." to probably 8px? It seems to be maybe around 16px so reduce it to half if the number is true. (Please see the references from the attachments)
2. I don't see AC No.3 is happening. Upon submit, the action doesn't take me to the error location. 

Let me know if you have any questions. Thanks!

**Screenshot**


My design
image

tglaz: @shiozakiyuko regarding AC 3: `I don't see AC No.3 is happening. Upon submit, the action doesn't take me to the error location.`

I'm not really sure if that's the best way to handle it. I understand the issue is that currently a user can click `Add Client` and then directly hit `Submit` and it'll error out (because no name is provided). But here are a couple of thoughts handling it, and preferably in separate tickets. 

1. If required fields are not filled then the `submit` button should be disabled
2. If `name` is required to create a Client, then let's open the form to that section right away. Currently a user can enter something else in another section, click Submit and get an error. And sure, we could then take them to the Personal Info section and tell them that they need to add the name, but perhaps we shouldn't allow the user to get there in the first place?

tglaz:
asdf

PR #699, opened 2024-11-12
Interaction document upload backend (DEV-855)

mikefeldberg:
DEV-855

turns out we didn't need to do anything. We can use the existing attachments backend for the new flow. The only backend change was updating the clientprofile docs properties (`doc_ready_documents`, `consent_form_documents`, `other_documents`) to return Attachments filtered on `associated_with`, rather than `self.documents`


apps/betterangels-backend/clients/schema.py
```diff
@@ -153,14 +153,14 @@ class Mutation:
     def create_client_document(self, info: Info, data: CreateClientDocumentInput) -> ClientDocumentType:
         with transaction.atomic():
             user = cast(User, get_current_user(info))
+            permission_group = get_user_permission_group(user)
```
mikefeldberg, line 156: this is just a lil refactor. I moved permission checks to the top of any op they were in

apps/betterangels-backend/clients/types.py
```diff
@@ -46,6 +46,7 @@ class ClientDocumentType(AttachmentInterface):
 class CreateClientDocumentInput:
     client_profile: ID
     file: Upload
+    note_id: Optional[ID]
```
vecchp, line 49: Not a fan of leaking note information into the client app

vecchp (changes requested):
The current approach is problematic as it retrieves an unrestricted number of documents each time the client profile loads, which could lead to performance issues as we scale. We should instead use a dedicated document endpoint with pagination capabilities.

We need to create a ticket to replace these model properties with paginated endpoints. Once the ticket is created, we can reference it here and proceed with the merge. This change should be prioritized since it’s a quick fix that addresses a potential scalability issue.

mikefeldberg:
this was a quick and dirty implementation of something that's no longer urgent and should be redone with pagination when we have the time

PR #700, opened 2024-11-12
Add searchClient to ClientProfiles query to support finding clients with identical names (DEV-912)

lena:
https://betterangels.atlassian.net/browse/DEV-912

Adds a new search filter: `searchClient` to `ClientProfiles` query. 

Example usage: 
```
        query = """
            query ClientProfiles($searchClient: ClientSearchInput) {
                clientProfiles(filters: {searchClient: $searchClient}) {
                    id
                }
            }
        """

        response = self.execute_graphql(
            query,
            variables={"searchClient": {"firstName":"Marc Anthony", "lastName": "Ramirez"},
        )
```

## Summary by Sourcery

Add a full name search feature to the ClientProfiles query to enable deduplication of clients with identical names, and implement corresponding tests to validate the functionality.

New Features:
- Introduce a full name search feature in the ClientProfiles query to allow deduplication of clients with identical names.

Tests:
- Add tests for the new full name search feature in the ClientProfiles query to ensure it returns only exact matches on first and last names.

### DIFF
--- a/apps/betterangels-backend/clients/tests/test_queries.py
+++ b/apps/betterangels-backend/clients/tests/test_queries.py
@@ -209,8 +209,9 @@ class ClientProfileQueryTestCase(ClientProfileGraphQLBaseTestCase):
     @parametrize(
         ("first_name, last_name, middle_name, expected_client_profile_count"),
         [
-            ("", "", None, 0),  # no filters
-            ("Todd", "", None, 0),  # missing last name
+            ("", "", "", 3),  # no filters
+            ("Todd", None, None, 1),  # exact match on first name only
+            (None, "Chavez", None, 2),  # exact match on last name only
             ("Tod", "Chavez", None, 0),  # inexact match on first name
             ("Todd", "Chave", None, 0),  # inexact match on last name
             ("Todd", "Chavez", "Eleanor", 0),  # inexact match on middle name
@@ -223,9 +224,6 @@ class ClientProfileQueryTestCase(ClientProfileGraphQLBaseTestCase):
     def test_client_profiles_query_search_client(
         self, first_name: str, last_name: str, middle_name: str | None, expected_client_profile_count: int
     ) -> None:
-        """
-        Assert that when using the full name client search, only exact first & last name matches are returned.
-        """
         self.graphql_client.force_login(self.org_1_case_manager_1)
 
         # create a new client with similar name to client 1, with space in first name
@@ -248,10 +246,13 @@ class ClientProfileQueryTestCase(ClientProfileGraphQLBaseTestCase):
             }
         """
 
-        response = self.execute_graphql(
-            query,
-            variables={"searchClient": {"firstName": first_name, "lastName": last_name, "middleName": middle_name}},
-        )
+        search_fields = {
+            **({"firstName": first_name} if first_name else {}),
+            **({"lastName": last_name} if last_name else {}),
+            **({"middleName": middle_name} if middle_name else {}),
+        }
+
+        response = self.execute_graphql(query, variables={"searchClient": search_fields})
 
         client_profiles = response["data"]["clientProfiles"]
         self.assertEqual(len(client_profiles), expected_client_profile_count)
--- a/apps/betterangels-backend/clients/tests/test_queries.py
+++ b/apps/betterangels-backend/clients/tests/test_queries.py
@@ -206,6 +206,50 @@ class ClientProfileQueryTestCase(ClientProfileGraphQLBaseTestCase):
         client_profiles = response["data"]["clientProfiles"]
         self.assertEqual(len(client_profiles), expected_client_profile_count)
 
+    @parametrize(
+        ("full_name, expected_client_profile_count"),
+        [
+            ("", 0),  # no filters
+            ("Todd", 0),  # too few search fields
+            ("Todd,Gustav,Chavez", 0),  # too many search fields
+            ("Tod,Chavez", 0),  # inexact match on first name
+            ("Todd,Chave", 0),  # inexact match on last name
+            ("Todd,Chavez", 1),  # exact match on first & last name
+            ("Todd , Chavez", 1),  # exact match on first & last name
+            ("Todd Gustav,Chavez", 1),  # exact match on first & last name
+        ],
+    )
+    def test_client_profiles_query_full_name_search(self, full_name: str, expected_client_profile_count: int) -> None:
+        """
+        Assert that when using the full name client search, only exact first & last name matches are returned.
+        """
+        self.graphql_client.force_login(self.org_1_case_manager_1)
+
+        # create a new client with similar name to client 1, with space in first name
+        self._create_client_profile_fixture(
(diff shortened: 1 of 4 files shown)


apps/betterangels-backend/clients/tests/test_queries.py
```diff
@@ -205,6 +205,49 @@ def test_client_profiles_query_search(
         client_profiles = response["data"]["clientProfiles"]
         self.assertEqual(len(client_profiles), expected_client_profile_count)
```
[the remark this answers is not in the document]
lena: happy to add this if people want it, but seems unnecessary for such a simple query

apps/betterangels-backend/clients/types.py
```diff
@@ -104,10 +104,34 @@ def search(
...
         queryset = queryset.filter(reduce(and_, q_objects))
 
-        return (
-            queryset,
-            Q(),
-        )
+        return (queryset, Q())
+
+    @strawberry_django.filter_field
+    def full_name_search(
```
tglaz, line 110: Is there a reason we couldn't use or update `search` above for functionality? 
Unless there's a strawberry related (some other ) reason, would we want to have a single general purpose search function for this model?

apps/betterangels-backend/clients/types.py
```diff
@@ -104,10 +104,34 @@ def search(
...
+        info: Info,
+        value: Optional[str],
+        prefix: str,
+    ) -> Tuple[QuerySet[ClientProfile], Q]:
+        """
+        Returns only exact match on first and last name (case insensitive).
+        Requires a "," delimiter to accommodate spaces in first/last names:
+        "{FIRST NAME},{LAST NAME}"
+
+        Ignores middle names and partial first name matches, e.g., "Marc,Ramirez" != "Marc Anthony,Ramirez"
```
tglaz, line 122: Could we pass in specific params to get the exact search needed? We may be relying a bit much on the comments for implementation detail, down to the order of search params and which fields are ignored. I'd prefer this to be specific.

apps/betterangels-backend/clients/tests/test_queries.py
```diff
@@ -205,6 +205,50 @@ def test_client_profiles_query_search(
...
+        ("full_name, expected_client_profile_count"),
+        [
+            ("", 0),  # no filters
+            ("Todd", 0),  # too few search fields
+            ("Todd,Gustav,Chavez", 0),  # too many search fields
+            ("Tod,Chavez", 0),  # inexact match on first name
+            ("Todd,Chave", 0),  # inexact match on last name
+            ("Todd,Chavez", 1),  # exact match on first & last name
+            ("Todd , Chavez", 1),  # exact match on first & last name
+            ("Todd Gustav,Chavez", 1),  # exact match on first & last name
```
tglaz, line 218: It's a bit confusing why 217 and 218 would both pass. 
I may be misunderstanding this, but can we guarantee that `Gustav` here is a middle name and not actually stored as `firstName`?

tglaz (changes requested):

apps/betterangels-backend/clients/tests/test_queries.py
lena: @tglaz oh sorry i wasn't clear, "Todd Gustav" _is_ the first name (for new client created on line 228 below). this test is showing that this search doesn't use the middle name at all, but does support spaces in the first name. 
whereas the 2 lines above return the `self.client_profile_1 ` client

apps/betterangels-backend/clients/types.py
tglaz: yeah, just noticed that. seems like a very specific implementation of `search`, but it happens

apps/betterangels-backend/clients/tests/test_queries.py
tglaz: so is the first name is `partial` match?

apps/betterangels-backend/clients/types.py
tglaz: Typically I’d imagine the interface to a generic search function to accept named params, so if I wanted to search for a client , and passed in `first_name + last_name` then it would search by those 2 fields. 
If I passed in `first_name + nickname + last_name` it would search by those 3 fields, etc…. 

Otherwise we’ll need to write a separate search function for each type of search.
A generic function, in theory, could also accept options for things such as:
* case sensitivity
* exact or partial match
* etc…

That’s not requested here, but I'd hope we could come back and extend this function to add that functionality.
If there are other limitations here then ok

Having said that, I can also defer to @mikefeldberg and Paul

apps/betterangels-backend/clients/tests/test_queries.py
lena: no it's an exact match, just case insensitive @tglaz

apps/betterangels-backend/clients/types.py
lena: kk, i can make a query input like: 
```{"firstName": "Marc Anthony", "lastName": "Ramirez"}``` 
and middle name can be optional

File: apps/betterangels-backend/clients/types.py [hunk 5]
```diff
@@ -104,10 +111,34 @@ def search(
...
+        info: Info,
+        value: UserSearchInput,
+        prefix: str,
+    ) -> Tuple[QuerySet[ClientProfile], Q]:
+        """
+        Returns only exact match on first and last name (case insensitive).
+        Accepts optional middle name param. If middle name is passed, middle name must be an exact match as well.
+        """
+        if not value.first_name or not value.last_name:
+            return (queryset.none(), Q())
```
lena, line 129: @tglaz i was thinking about this endpoint specifically for deduping purposes, so that's why it requires first and last name... that might not be expected behavior so that's why i'm calling it out
like if you want to find all the people with "Anthony" in their name, you would use the traditional `search` filter above. however, that would return people with "Anthony" in either their first OR last name.

apps/betterangels-backend/clients/types.py
```diff
@@ -104,10 +111,31 @@ def search(
...
+        """
+        filters = {}
+
+        user_fields = ["first_name", "middle_name", "last_name"]
+
+        for field in user_fields:
+            if field_value := getattr(value, field):
+                filters[f"user__{field}__iexact"] = field_value.strip()
+
+        queryset = queryset.filter(**filters)
```
tglaz, line 136: can I assume `queryset.filter` returns null set if filters are empty?

tglaz (changes requested):

apps/betterangels-backend/clients/types.py
lena: opposite, it returns everything. i can make it return nothing

tglaz: I'd typically assume that a search with no valid inputs would either return an empty set or an error. But I guess i've seen it done both ways. Probably safer to return nothing though

tglaz (approved):

PR #702, opened 2024-11-12
zoom images (827)

Davit-BetterAngels:
DEV-827

## Summary by Sourcery

Implement image zoom functionality by introducing the ImagesWithZoom component and integrating it into the FileScreenComponent. Update dependencies to include react-native-image-zoom-viewer to support this feature.

New Features:
- Introduce the ImagesWithZoom component to enable zoom functionality on images within the application.

Enhancements:
- Refactor the FileScreenComponent to utilize the new ImagesWithZoom component for displaying images with zoom capability.

Build:
- Add react-native-image-zoom-viewer as a new dependency in the package.json files to support image zoom functionality.


shiozakiyuko (approved):
Overall, it worked (android and iOS). However, it loads the files pretty slow on iOS side. It may be because of my wifi situation, or the files themselves are really big and I can't pin point the issue. I just wanted to throw what I experienced here.

libs/expo/shared/ui-components/src/lib/ImagesWithZoom/ImagesWithZoom.tsx
```diff
@@ -0,0 +1,75 @@
...
+        >
+          <PlusIcon size="md" color={Colors.BLACK} rotate="45deg" />
+        </Pressable>
+        {title && (
+          <View
+            style={{
+              position: 'absolute',
+              top: insets.top + Spacings.sm,
+              left: 0,
+              right: 0,
```
tglaz, line 57: think it may need an offset for the close icon. 
could also separate the modal into `header` and `body` sections, with title and close Icon inside the header. That way could just use flex (or relative position anyway)

apps/betterangels/package.json
```diff
@@ -81,7 +81,8 @@
     "@teovilla/react-native-web-maps": "*",
     "react-native-paper": "*",
     "dotenv": "*",
-    "@react-native-picker/picker": "*"
+    "@react-native-picker/picker": "*",
+    "react-native-image-zoom-viewer": "*"
```
tglaz, line 85: Any concerns with this lib being archived and not updated in over 4 years?
Looks like it's still getting more weekly downloads then the alternatives, so perhaps it's fine.

tglaz (changes requested):
main issue I see is that the filename can overlap the `close icon`, making hard to see. otherwise looks good

apps/betterangels/package.json
Davit-BetterAngels: other one was 3 years ago. but downloads were 2 times less.
react-native-image-viewing

tglaz (approved):

PR #705, opened 2024-11-13
Add California ID to client profile frontend (DEV-894)

mikefeldberg:
DEV-894

on hold until DEV-896 goes out

## Summary by Sourcery

Add support for California ID in client profiles, including input validation and duplicate checking.

New Features:
- Add support for California ID in client profiles, allowing users to input and validate California ID numbers.

Enhancements:
- Improve duplicate checking by integrating a query to check for existing client profiles with the same California ID.

## Summary by Sourcery

Add support for California ID in client profiles, including input validation and duplicate checking. Enhance the system to prevent duplicate entries by checking existing profiles for matching California IDs. Update tests to cover new functionality.

New Features:
- Add support for California ID in client profiles, allowing users to input and validate California ID numbers.

Enhancements:
- Improve duplicate checking by integrating a query to check for existing client profiles with the same California ID.

Tests:
- Add tests for searching client profiles by California ID to ensure correct duplicate detection.


libs/expo/betterangels/src/lib/screens/AddEditClient/PersonalInfo/index.tsx
```diff
@@ -17,6 +21,49 @@ interface IPersonalInfoProps {
...
+    const shouldCheck =
+      californiaId &&
+      californiaId.length === californiaIdLength &&
+      !errors['californiaId'];
+
+    setCheckForDuplicates(shouldCheck);
+  }, [californiaId, errors]);
+
+  useEffect(() => {
+    if (data && data?.clientProfiles?.length > 0) {
```
mikefeldberg, line 58: alternatively, we could check if the current client == the returned client. Then we could remove the excluded id param. I think it's faster to do this work on the backend, so I went with that approach

apps/betterangels-backend/clients/types.py
```diff
@@ -58,9 +58,11 @@ class ClientProfileOrder:
 
 @strawberry.input
 class ClientSearchInput:
-    first_name: str | None = None
-    last_name: str | None = None
-    middle_name: str | None = None
+    excluded_client_profile_id: Optional[str] = None
```
mikefeldberg, line 61: see: https://github.com/BetterAngelsLA/monorepo/pull/705/files#r1839631579

libs/expo/betterangels/src/lib/screens/AddEditClient/PersonalInfo/index.tsx
```diff
@@ -17,6 +21,49 @@ interface IPersonalInfoProps {
...
+          excludedClientProfileId: (clientProfileId as string) || null,
+          californiaId: californiaId,
+        },
+      },
+    },
+    fetchPolicy: 'network-only',
+    nextFetchPolicy: 'network-only',
+  });
+
+  useEffect(() => {
```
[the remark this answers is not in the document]
mikefeldberg: seems unnecessary since we're only hitting the API when the contents are an exact length

libs/expo/betterangels/src/lib/hooks/californiaIdValidation/useCaliforniaIdValidation.ts
```diff
@@ -0,0 +1,24 @@
+import { useClientProfilesQuery } from '../../screens/Clients/__generated__/Clients.generated';
```
mikefeldberg, line 1: maybe this needs its own query. something that _only_ returns the id

libs/expo/betterangels/src/lib/screens/AddEditClient/index.tsx
```diff
@@ -206,6 +206,15 @@ export default function AddEditClient({ id }: { id?: string }) {
             message: 'User with this Email already exists.',
           });
           return;
+        } else if (
+          operationResult.messages[0].message ===
```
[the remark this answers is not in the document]
tglaz: and given the `operationResult.messages` returned payload (see below), we can maybe compare on `field` and `kind` only? I say this assuming that `VALIDATION` means a validation error.

```javascript
[
    {
        "__typename": "OperationMessage",
        "field": "californiaId",
        "kind": "VALIDATION",
        "message": "California ID must be 1 letter followed by 7 numbers"
    }
]
```

libs/expo/betterangels/src/lib/hooks/californiaIdValidation/useCaliforniaIdValidation.ts
```diff
@@ -0,0 +1,24 @@
...
+
+export default function useCaliforniaIdValidation(
```
tglaz, line 3: I'd consider renaming as this just checks uniqueness part of validation and not full validation.
I'd also expect validation to return a boolean or a message etc.

libs/expo/betterangels/src/lib/screens/AddEditClient/PersonalInfo/index.tsx
```diff
@@ -17,6 +21,25 @@ interface IPersonalInfoProps {
 
 export default function PersonalInfo(props: IPersonalInfoProps) {
   const { scrollRef, expanded, setExpanded } = props;
+  const { id: clientProfileId } = useLocalSearchParams();
+  const { watch, setError, clearErrors } = useFormContext();
+  const californiaId = watch('californiaId');
```
tglaz, line 26: I believe `watch` may trigger re-renders in parent and sibling components with each change, so kind of expensive. 
`useWatch` should be safer in this particular case

```javascript
  const { control, setError, clearErrors } = useFormContext();

  const [californiaId] = useWatch({
    control,
    name: ['californiaId'],
  });
```

tglaz (changes requested):

libs/expo/betterangels/src/lib/screens/AddEditClient/index.tsx
```diff
@@ -206,6 +206,15 @@ export default function AddEditClient({ id }: { id?: string }) {
             message: 'User with this Email already exists.',
           });
           return;
+        } else if (
```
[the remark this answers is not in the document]
mikefeldberg: I think the error here is the `return`. there should be one `return` at the bottom of the block. updated

libs/expo/betterangels/src/lib/screens/AddEditClient/index.tsx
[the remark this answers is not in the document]
mikefeldberg: I added this to the list of things we're looking for. Tho form validation should prevent this from being submitted. side note: maybe we should have an `errorMessagesMap` where all of these live

libs/expo/shared/static/src/lib/regex.ts
```diff
@@ -4,4 +4,5 @@ export const Regex = {
   time: /^(?:[01]\d|2[0-3]):[0-5]\d$/,
   email: /^[a-zA-Z0-9._%+-]+@[a-zA-Z0-9.-]+\.[a-zA-Z]{2,}$/,
   phoneNumber: /^[2-9]\d{2}[2-9]\d{6}$/,
+  californiaId: /^[A-Z]\d{7}$/,
```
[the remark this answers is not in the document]
mikefeldberg: the backend validation requires capitalized first letter. I added `autoCapitalize` to the input for good measure

libs/expo/betterangels/src/lib/screens/AddEditClient/PersonalInfo/index.tsx
mikefeldberg: neat, thanks!

libs/expo/betterangels/src/lib/hooks/californiaIdValidation/useCaliforniaIdValidation.ts
mikefeldberg: @tglaz I added a slim query just for this. not sure if it should be colocated with the hook 🤷‍♂️

libs/expo/betterangels/src/lib/screens/AddEditClient/index.tsx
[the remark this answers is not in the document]
tglaz: So perhaps i'm reading it wrong, but below is what I see the function doing.
Only 1 out of these blocks will ever execute, with thrown error being the fallback. 
The `if/else` chain obscures that in my view

```javascript
      if (
        operationResult?.__typename === 'OperationInfo' &&
        operationResult.messages.length > 0
      ) {
        const resultMessage = operationResult.messages[0].message;

        if (resultMessage === 'X') {
          methods.setError('user.email' ...

          return;
        } 
        
        if (resultMessage === 'Y') {
          methods.setError('californiaId' ...

          return;
        } 
        
        if (resultMessage === 'Z') {
          methods.setError('californiaId' ...

          return;
        } 
        
        throw new Error(
          `Failed to update a client profile: ${resultMessage}`
        );
      }
```

libs/expo/betterangels/src/lib/hooks/californiaIdUniqueCheck/useCaliforniaIdUniqueCheck.ts
```diff
@@ -0,0 +1,25 @@
+import { Regex } from '@monorepo/expo/shared/static';
+import { useClientCaliforniaUniqueCheckQuery } from './__generated__/ClientCaliforniaUniqueCheck.generated';
+
+export default function useCaliforniaIdUniqueCheck(
+  californiaId: string,
+  clientProfileId?: string
+) {
+  const { data } = useClientCaliforniaUniqueCheckQuery({
+    skip: !californiaId || !Regex.californiaId.test(californiaId),
```
tglaz, line 9: just the regex would be enough: `skip: !Regex.californiaId.test(californiaId)`

libs/expo/betterangels/src/lib/screens/AddEditClient/index.tsx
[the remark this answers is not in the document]
tglaz: form validation is completely separate, and better to turn it off (ignore) when testing this functionality

As for `errorMessagesMap` yeah, think you're right and we should communicate those in a better way.
I wonder if we could specify the type of validation failed and return that with the payload.

libs/expo/shared/static/src/lib/regex.ts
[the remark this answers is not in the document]
tglaz: ok, though this may cause some confusion down the road

libs/expo/betterangels/src/lib/hooks/californiaIdValidation/useCaliforniaIdValidation.ts
tglaz: I'd go with whatever we're doing today I suppose
I'm not used to doing that, but I realize there are some graphQl conventions etc that would say yes.

libs/expo/betterangels/src/lib/screens/AddEditClient/index.tsx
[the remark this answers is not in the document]
mikefeldberg: wfm

libs/expo/betterangels/src/lib/hooks/californiaIdUniqueCheck/useCaliforniaIdUniqueCheck.ts
mikefeldberg: so it would

libs/expo/betterangels/src/lib/screens/AddEditClient/index.tsx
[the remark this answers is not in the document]
tglaz: 🙏

tglaz (approved):

shiozakiyuko (commented):
Hi @mikefeldberg 

I didn't get to see the visual feedback after I submitted a CA ID. I tried again by rescanning this QR code to refresh and start over again but I get the same result. 
The visual feedback: what I meant is after tapping the Submit button, I expected the page guides me to view-only mode of Client Profile page but it didn't happen. 

Although as I tapped the Back button (thinking not saving the ID info), the info I entered was there. So I know it's saved but somehow, it just doesn't behave the way it should be. 

It's happening on both, Android and iOS.

Please check. thanks!

libs/expo/betterangels/src/lib/hooks/californiaIdUniqueCheck/ClientCaliforniaUniqueCheck.graphql
```diff
@@ -0,0 +1,7 @@
+query ClientCaliforniaUniqueCheck($filters: ClientProfileFilter) {
```
shiozakiyuko, line 1: Hi @mikefeldberg

I didn't get to see the visual feedback after I submitted a CA ID. I tried again by rescanning this QR code to refresh and start over again but I get the same result.
The visual feedback: what I meant is after tapping the Submit button, I expected the page guides me to view-only mode of Client Profile page but it didn't happen.

Although as I tapped the Back button (thinking not saving the ID info), the info I entered was there. So I know it's saved but somehow, it just doesn't behave the way it should be.

It's happening on both, Android and iOS.

Please check. thanks!

shiozakiyuko: Thanks, @mikefeldberg 

I confirmed the submit action takes me to the view-only mode page now. 

One more thing I noticed though.

This is only happening on iOS (Android was fine). 

When I submitted the invalid CA ID, I see the validation message at the beginning. But if I ignored it and tapped on the Submit button anyway, the message disappears. But the error snicker bar (alert) shows up (understandable). My questions is: shouldn't the validation message be persistent till the user enters the valid format or remove the entry?

Thanks!

mikefeldberg: @shiozakiyuko I see what happened here. fixed!

libs/expo/betterangels/src/lib/screens/AddEditClient/index.tsx
```diff
@@ -192,27 +192,46 @@ export default function AddEditClient({ id }: { id?: string }) {
...
-        } else {
-          throw new Error(
-            `Failed to update a client profile: ${operationResult.messages[0].message}`
-          );
         }
-      }
 
+        if (
+          resultMessage ===
+          'California ID must be 1 letter followed by 7 numbers'
```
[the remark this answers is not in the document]
mikefeldberg: I moved the uniqueness check into the child component. Tom noted that the client notification will pop up in a modal, which can be triggered from anywhere, so there's no point in keeping this logic higher

I also added the uniqueness check to the field rules. This fixes the behavior that yuko noted [above](https://github.com/BetterAngelsLA/monorepo/pull/705/files#r1841373779)

Davit-BetterAngels (approved):

shiozakiyuko (approved):
Looks good to me!

PR #750, opened 2024-12-03
Fix form submission multiple times

CruiseDevice:
- implemented a lock mechanism using React's state management.

## Summary by Sourcery

Bug Fixes:
- Implement a lock mechanism to prevent multiple form submissions by using a state variable 'isSubmitting'.


libs/expo/betterangels/src/lib/screens/AddEditClient/index.tsx
```diff
@@ -391,7 +393,7 @@ export default function AddEditClient({ id }: { id?: string }) {
             )}
           </MainScrollContainer>
           <BottomActions
-            disabled={!!methods.formState.errors.californiaId}
+            disabled={!!methods.formState.errors.californiaId || isCreating || isUpdating}
```
mikefeldberg, line 396: it looks like you only have to add `isCreating` and `isUpdating` to `isLoading`, so you can remove it here. Updating either prop will have the same effect, but `isLoading` is more appropriate, since we're specifically passing a loading state

`libs/expo/shared/ui-components/src/lib/BottomActions/index.tsx` ([link](https://github.com/BetterAngelsLA/monorepo/blob/main/libs/expo/shared/ui-components/src/lib/BottomActions/index.tsx#L11-L12)) is a bit confusing, because it takes both a `disabled` prop and an `isLoading` prop. But the `<Button>` component's `disabled` prop combines them:
```js
<Button
  loading={isLoading}
  disabled={isLoading || disabled}
  fontSize="sm"
  height="md"
  variant="primary"
  style={{ minWidth: 90 }}
  accessibilityHint="submit the form"
  title={submitTitle}
  onPress={onSubmit}
/>
```
which is why passing the loading state to either one will have the same effect

libs/expo/betterangels/src/lib/screens/AddEditClient/index.tsx
```diff
@@ -128,39 +129,40 @@ export default function AddEditClient({ id }: { id?: string }) {
...
-      values.contacts = values.contacts.map((contact) => ({
-        ...contact,
-        phoneNumber: contact.phoneNumber === '' ? null : contact.phoneNumber,
-      }));
-    }
-
-    const filteredSocialMediaProfiles =
-      values.socialMediaProfiles?.filter((item) => item.platformUserId) || [];
+    if (isCreating || isUpdating) return;
+    try {
```
mikefeldberg, line 133: I don't think we should be moving all these operations into the `try` block. 

The `try`/`catch` are here because it's possible for something to go wrong when the frontend tries to connect to the backend. So the frontend needs to be protected in case the mutation fails or returns an error. But everything before `let operationResult;` is just the frontend cleaning the data before sending it to the backend, so it should be safe. 

Technically, it is possible for something to go wrong here, but that should be caught/prevented with type checking, field validation, etc. By the time we're cleaning the data here, we know exactly what shape it's going to be, so we don't have to worry about something breaking

libs/expo/betterangels/src/lib/screens/AddEditClient/index.tsx
```diff
@@ -128,39 +129,40 @@ export default function AddEditClient({ id }: { id?: string }) {
...
-    if (values.contacts && values.contacts?.length > 0) {
-      values.contacts = values.contacts.map((contact) => ({
-        ...contact,
-        phoneNumber: contact.phoneNumber === '' ? null : contact.phoneNumber,
-      }));
-    }
-
-    const filteredSocialMediaProfiles =
-      values.socialMediaProfiles?.filter((item) => item.platformUserId) || [];
+    if (isCreating || isUpdating) return;
```
mikefeldberg, line 132: I think you can remove this. Adding `isCreating || isUpdating` to `isLoading`, on line 406, should achieve the desired result

libs/expo/betterangels/src/lib/screens/AddEditClient/index.tsx
```diff
@@ -85,24 +85,25 @@ export default function AddEditClient({ id }: { id?: string }) {
...
-        variables: {
-          pagination: { limit: 20 + 1, offset: 0 },
-          filters: {
-            search: '',
-          },
-          order: {
-            user_FirstName: Ordering.AscNullsFirst,
-            id: Ordering.Desc,
+  const [updateClient, { loading: isUpdating }] = useUpdateClientProfileMutation();
+  const [createClient, { loading: isCreating }] =
```
mikefeldberg, line 89: nice 👍

PR #776, opened 2024-12-11
Dev 1098 #2/update service list fe

jluna2000:
Regenerated types with the ynx-generate command and made corresponding frontend changes after the merging of DEV-1106 to reflect the backend changes in the frontend. Modified the files that made use of the regenerated types to display the new choices properly in the app.
Frontend_Screenshots

PR #777, opened 2024-12-12
Update services list FE (DEV-1098)

mikefeldberg:
closes DEV-1098
replaces #776


libs/expo/betterangels/src/lib/static/constants.ts
```diff
@@ -6,13 +6,20 @@ export const ServicesByCategory = [
     items: [
       ServiceEnum.Food,
       ServiceEnum.Water,
+      ServiceEnum.Bag,
```
shiozakiyuko, line 9: Hi @mikefeldberg 

I noticed a few things.

1. **Socks - Clothes or Clothes - Socks**
In the spreadsheet, it shows "Clothes - Socks" - it should be consistent. 

2. **The new items from Harm Reduction are missing.**
- Harm Reduction - Safe Smoking
- Harm Reduction - Narcan
- Harm Reduction - Test Kit

3. **Search is not fully functioning**
When the items are with multiple words such as Sleeping Bag, or HMIS Consent, when I typed the first word, I see it in the search result. However, when I fully typed the name in, the search result shows "No Results".

FYI: I only tested this on iOS. I am still having the same issue with Android (I need to reach out Paul for this).

mikefeldberg: 1. updated
2. we decided to remove these for now, since we already have "harm reduction"
3. this is the same behavior as in main

shiozakiyuko: Ok it looks good. Approved.

shiozakiyuko (approved):

libs/expo/betterangels/src/lib/static/constants.ts
```diff
@@ -6,13 +6,20 @@ export const ServicesByCategory = [
...
+      ServiceEnum.Batteries,
       ServiceEnum.Blanket,
       ServiceEnum.Book,
       ServiceEnum.Clothes,
+      ServiceEnum.ClothesSocks,
```
vecchp, line 14: How did this item make it in again? We mentioned that we were going to collapse ALL sections into general categories for now.

Davit-BetterAngels (approved):

PR #786, opened 2024-12-17
Dev 884/each block in hmis id section has its own clear button

aaronipollock:
— Each HMIS ID block has its own "Clear" (instead of "Reset") and "Remove" button.

— The onItemReset (now onItemClear) functionality has been changed to set the input field in the block to null (though **in the code it is set to an empty string, not null** because my linter didn't like null) and the CoC is set to LAHSA.

— The "Add another HMIS ID" button still appears at the bottom of HMIS ID section.

## Summary by Sourcery

Update the HMIS ID section to provide individual 'Clear' and 'Remove' buttons for each block, replacing the 'Reset' button. Modify the onItemClear function to set the agency to LAHSA and clear the HMIS ID field. Add a tsconfig.json file to the project.

Enhancements:
- Each HMIS ID block now has its own 'Clear' and 'Remove' button, replacing the previous 'Reset' button.

Chores:
- Added a tsconfig.json file extending the base Expo TypeScript configuration.


libs/expo/betterangels/src/lib/screens/AddEditClient/PersonalInfo/HmisProfiles.tsx
```diff
@@ -37,9 +36,13 @@ export default function HmisProfiles() {
     setValue('hmisProfiles', []);
   };
 
-  const onItemReset = (index: number) => {
-    resetField(`hmisProfiles.${index}.agency`);
-    resetField(`hmisProfiles.${index}.hmisId`);
+  const onItemClear = (index: number) => {
+    const updatedProfiles = [...hmisProfiles];
+    updatedProfiles[index] = {
+      agency: HmisAgencyEnum.Lahsa,
```
tglaz, line 42: small request: could you set `HmisAgencyEnum.Lahsa` as a default agency and then use it here and below where we call `append` so those two are in sync?

At top of a file you could just declare `const DEFAULT_AGENCY = HmisAgencyEnum.Lahsa` and then just use that const.

tsconfig.json
```diff
@@ -0,0 +1,4 @@
+{
```
tglaz, line 1: Is there a specific reason to commit this? 
I may be wrong, but I'm not sure we actually need it for this project as we use `tsconfig.base.json`.

libs/expo/betterangels/src/lib/screens/AddEditClient/PersonalInfo/HmisProfiles.tsx
aaronipollock: Done. Sorry for not getting to these sooner.

tglaz (approved):

PR #730, opened 2024-11-22
Expo 52 + Nx 20 Upgrade

vecchp:
DEV-1105
DEV-1104
DEV-1064
DEV-1063

## Summary by Sourcery

Upgrade the project to Expo 52 and Nx 20, including updates to dependencies and configurations. Refactor the tab layout to use a reusable TabIcon component, improving code maintainability. Enhance the Storybook setup with additional addons and updated story paths. Adjust build scripts and dependencies to align with the new versions.

Enhancements:
- Refactor tab icon rendering by introducing a reusable TabIcon component to simplify the code and improve maintainability.
- Update screen options configuration for Tabs to enhance styling and consistency across the application.

Build:
- Update package.json scripts and dependencies to align with Expo 52 and Nx 20 upgrades.

CI:
- Modify Storybook configuration to include new addons and update the story paths for better organization.

PR #807, opened 2025-01-06
Added fixes for tab sizing issue (DEV-1063)

telishreyas10:
[DEV-1063](https://betterangels.atlassian.net/browse/DEV-1063)

## Summary by Sourcery

Bug Fixes:
- Fixed tab sizing issues on native platforms by applying specific styles to tab bar icons.

[DEV-1063]: https://betterangels.atlassian.net/browse/DEV-1063?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/betterangels/src/app/(tabs)/_layout.tsx
+++ b/apps/betterangels/src/app/(tabs)/_layout.tsx
@@ -89,7 +89,12 @@ export default function TabLayout() {
             },
             headerShadowVisible: false,
             tabBarIcon: ({ color, focused }) => (
-              <View style={{ alignItems: 'center', width: 80, height: 15 }}>
+              <View
+                style={{
+                  alignItems: 'center',
+                  ...(isNative && styles.tabBarIconStyle),
+                }}
+              >
                 {focused ? (
                   <HouseSolidIcon color={color} />
                 ) : (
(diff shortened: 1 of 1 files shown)


apps/betterangels/src/app/(tabs)/_layout.tsx
```diff
@@ -290,4 +315,8 @@ const styles = StyleSheet.create({
     justifyContent: 'center',
     backgroundColor: Colors.WHITE,
   },
+  tabBarIconStyle: {
+    width: 80,
+    height: 15,
```
tglaz, line 320: I'm assuming this looks correctly when tested, but setting the height to 15 seems wrong. Any reason it's set so small?

Also, I'm not 100% familiar with the issue, but does setting a height break things on web? Or could we just have the same code for all of platforms given that it's just styling/css.

tglaz (changes requested):

apps/betterangels/src/app/(tabs)/_layout.tsx
telishreyas10: The above changes are not required as the fixes were already made in https://github.com/BetterAngelsLA/monorepo/pull/730. 

@vecchp has closed this PR.

PR #810, opened 2025-01-07
Showing Area Code Error (DEV-1121)

Davit-BetterAngels:
DEV-1121

## Summary by Sourcery

Bug Fixes:
- Fixed an issue where invalid area codes were accepted when adding or editing phone numbers.

### DIFF
--- a/libs/expo/betterangels/src/lib/screens/AddEditClient/ContactInfo/PhoneNumbers.tsx
+++ b/libs/expo/betterangels/src/lib/screens/AddEditClient/ContactInfo/PhoneNumbers.tsx
@@ -43,8 +43,15 @@ export default function PhoneNumbers() {
               }
               rules={{
                 validate: (value: string) => {
-                  if (value && !Regex.phoneNumber.test(value)) {
-                    return 'Enter a 10-digit phone number without space or special characters';
+                  if (value) {
+                    const areaCode = value.substring(0, 3);
+
+                    if (
+                      !Regex.phoneNumber.test(value) ||
+                      !Regex.usAreaCode.test(areaCode)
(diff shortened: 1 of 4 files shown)


libs/expo/shared/static/src/lib/regex.ts
```diff
@@ -4,5 +4,6 @@ export const Regex = {
   time: /^(?:[01]\d|2[0-3]):[0-5]\d$/,
   email: /^[a-zA-Z0-9._%+-]+@[a-zA-Z0-9.-]+\.[a-zA-Z]{2,}$/,
   phoneNumber: /^[2-9]\d{2}[2-9]\d{6}$/,
+  usAreaCode: /^[2-9]/,
```
tglaz, line 7: doesn't `phoneNumber` regex cover this?
also, a code like 555555 would pass, so while checking that the number starts with /2-9/ is good, it may be inaccurate to label it `usAreaCode`

tglaz (changes requested):

libs/expo/betterangels/src/lib/screens/AddEditClient/ContactInfo/PhoneNumbers.tsx
```diff
@@ -43,8 +43,15 @@ export default function PhoneNumbers() {
               }
               rules={{
                 validate: (value: string) => {
-                  if (value && !Regex.phoneNumber.test(value)) {
```
shiozakiyuko, line 46: Hi @Davit-BetterAngels 

I don't see the error message under the textbox with red border.
I think this error should be specific that needs to say "Enter valid area code"

PR #812, opened 2025-01-07
Dev 1155/replace client profiles with client profiles paginated

aaronipollock:
DEV-1155

This PR replaces the existing clientProfiles queries with paginated versions and updates the schema accordingly.

- Updated GraphQL schema
- Updated GraphQL queries
- Automated types updates (ynx-generate-types)
- Modified components to work with the new paginated data structure
- Updated and passed relevant tests

## Summary by Sourcery

Replace the clientProfiles queries with paginated versions.

New Features:
- Added pagination to the client profiles query.

Tests:
- Updated tests to reflect the changes in the client profiles query.


apps/betterangels-backend/clients/tests/test_permissions.py
```diff
@@ -132,13 +132,15 @@ def test_view_client_profiles_permission(self, user_label: str, should_succeed:
...
         mutation = """
             query ViewClientProfiles {
-                clientProfiles {
-                    id
-                    user {
-                        firstName
-                        lastName
-                        middleName
-                        email
+                clientProfilesPaginated(pagination: {offset: 0, limit: 10}) {
```
mikefeldberg, line 135: `pagination` is an optional argument that only has to be passed when needed. We don't need to paginate these results, so it's ok to omit `paginated`
```suggestion
                clientProfilesPaginated {
```

apps/betterangels-backend/clients/tests/test_queries.py
```diff
@@ -112,16 +112,23 @@ def test_client_profiles_query(self) -> None:
 
         query = f"""
             query ViewClientProfiles {{
-                clientProfiles{{
-                    {self.client_profile_fields}
+                clientProfilesPaginated(pagination: {{offset: $offset, limit: $limit}}) {{
```
mikefeldberg, line 115: I think this test can be left as is, since `test_client_profiles_paginated_query`, below, is testing the paginated query

libs/expo/betterangels/src/lib/screens/Clients/Clients.graphql
```diff
@@ -29,37 +29,44 @@ mutation CreateNote($data: CreateNoteInput!) {
...
+        user {
+            id
+            email
+            firstName
+            lastName
+            username
+        }
+        displayCaseManager
       }
-      displayCaseManager # TODO: This shouldn't be here. Temporary fix while we figure out type generation.
```
mikefeldberg, line 62: TODO comments should be left in unless you're certain that the thing they're describing has been resolved

libs/expo/betterangels/src/lib/screens/Clients/index.tsx
```diff
@@ -35,7 +35,7 @@ export default function Clients({ Logo }: { Logo: ElementType }) {
   const [createNote] = useCreateNoteMutation();
   const [offset, setOffset] = useState<number>(0);
```
mikefeldberg, line 36: > We can remove the local offset state and rely on pageInfo.offset from the response. This way, we always use the server’s knowledge of the current offset rather than maintaining our own copy.
https://github.com/BetterAngelsLA/monorepo/pull/822#discussion_r1915416761

libs/expo/betterangels/src/lib/screens/Clients/index.tsx
```diff
@@ -35,7 +35,7 @@ export default function Clients({ Logo }: { Logo: ElementType }) {
...
   const [hasMore, setHasMore] = useState<boolean>(true);
-  const { data, loading } = useClientProfilesQuery({
+  const { data, loading } = useClientProfilesPaginatedQuery({
     variables: {
       pagination: { limit: paginationLimit + 1, offset },
```
mikefeldberg, line 40: > We don't need to do +1 anymore given we now know how many results there are. We originally used the over-fetching approach since we didn't know the total count.
https://github.com/BetterAngelsLA/monorepo/pull/822#discussion_r1915419752

mikefeldberg (commented):
I left a few comments, but I think that Paul's feedback on #822 applies here, as well.

https://github.com/BetterAngelsLA/monorepo/pull/822#pullrequestreview-2550722095

PR #821, opened 2025-01-13
Select Button component

Davit-BetterAngels:
Select Button for filters


libs/expo/shared/ui-components/src/lib/SelectButton/SelectButton.tsx
```diff
@@ -0,0 +1,51 @@
+import { ChevronLeftIcon } from '@monorepo/expo/shared/icons';
+import { Colors, Radiuses, Spacings } from '@monorepo/expo/shared/static';
+import { Pressable } from 'react-native';
+import { BaseContainer } from '../Container';
+import TextRegular from '../TextRegular';
+
+interface ISelectButtonProps {
+  defaultLabel?: string;
+  selected?: { id: string; label: string }[];
```
tglaz, line 9: this works, but may possibly simplify by accepting simply `string[]` (since not really using Id)

libs/expo/shared/ui-components/src/lib/SelectButton/SelectButton.tsx
```diff
@@ -0,0 +1,51 @@
...
+        borderRadius: Radiuses.xxl,
+        flexDirection: 'row',
+        alignItems: 'center',
+        gap: Spacings.xxs,
+        paddingVertical: Spacings.xs,
+        paddingLeft: Spacings.xs,
+        backgroundColor: backgroundColor,
+      }}
+    >
+      <TextRegular color={textColor}>{computedLabel}</TextRegular>
```
tglaz, line 45: we may need to add something like `text-overflow: ellipsis` for when the button has a fixed width

libs/expo/shared/ui-components/src/lib/SelectButton/SelectButton.spec.tsx
```diff
@@ -0,0 +1,64 @@
+import { fireEvent, render } from '@testing-library/react-native';
+import { SelectButton } from './SelectButton';
+
+describe('SelectButton Component', () => {
+  it('renders with default label when no selection is made', () => {
+    const { getByText } = render(
+      <SelectButton defaultLabel="All" selected={[]} onPress={jest.fn()} />
```
tglaz, line 7: defaultLabel already defaults to `All`, so I'd use a different one in at least one test, or not pass one at all

tglaz (changes requested):

libs/expo/shared/ui-components/src/lib/SelectButton/SelectButton.tsx
Davit-BetterAngels: what should be the max width? nowhere was specified

libs/expo/shared/ui-components/src/lib/SelectButton/SelectButton.spec.tsx
Davit-BetterAngels: done

libs/expo/shared/ui-components/src/lib/SelectButton/SelectButton.tsx
Davit-BetterAngels: done

tglaz (approved):

PR #824, opened 2025-01-14
Dev 1156/replace client documents with client documents paginated

jluna2000:
Replace all remaining use of the clientDocuments query with clientDocumentsPaginated and rename them to clientDocuments.

## Summary by Sourcery

Enhancements:
- Rename clientDocumentsPaginated to clientDocuments.


apps/betterangels-backend/clients/tests/test_permissions.py
```diff
@@ -266,14 +266,17 @@ def test_view_client_documents_permission(self, user_label: str, should_succeed:
         query = """
             query ViewClientDocuments {
                 clientDocuments {
-                    id
+                    totalCount
```
mikefeldberg, line 269: we don't need to include `totalCount` here. It's not hurting anything, but it's not being used either 🤷

apps/betterangels-backend/clients/tests/test_permissions.py
```diff
@@ -282,6 +285,6 @@ def test_view_client_documents_permission(self, user_label: str, should_succeed:
             )
         else:
             self.assertTrue(
-                len(response["data"]["clientDocuments"]) == 0,
+                len(response["data"]["clientDocuments"]["results"]) == 0,
```
mikefeldberg, line 288: actually, with the above comment in mind, if you wanted to, you could replace `len()` with `totalCount`
```suggestion
               response["data"]["clientDocuments"]["totalCount"] == 0,
```

mikefeldberg (commented):
one nit, looks good otherwise 👍

apps/betterangels-backend/clients/tests/test_permissions.py
mikefeldberg, line 288: come to think of it, I'm not sure why this was originally written with a `assertTrue`. Should have been `self.assertEqual(<len_results>, 0)`

mikefeldberg: come to think of it, I'm not sure why this was originally written with a `assertTrue`. Should have been `self.assertEqual(<len_results>, 0)`

PR #856, opened 2025-01-22
Merge disable button

vecchp:
## Summary by Sourcery

Style disabled buttons.

New Features:
- Introduce CSS variables for disabled button text and background colors.

Enhancements:
- Standardize the style of disabled buttons across the application.


apps/wildfires/src/assets/styles/global.css
```diff
@@ -50,3 +52,10 @@ textarea:focus {
   border-radius: 50px !important;
   border-color: var(--color-brand-yellow);
 }
+
+button:disabled,
+.button:disabled {
```
mikefeldberg, line 57: This may not work as expected, depending on where it's used. `button:disabled` will work on `button` elements with the `disabled` tag. `.button:disabled` will work on any element that has `class="button"` AND responds to `disabled` tags. e.g., if we have a pressable `<div class="button" disabled>`, this won't have the desired effect, because `<div>`s don't respond to `disabled`

PR #860, opened 2025-01-23
best practices (DEV-1273)

Davit-BetterAngels:
DEV-1273

## Summary by Sourcery

New Features:
- Add a result page that displays personalized action plans with best practices information.


apps/wildfires/src/app/shared/components/bestPractices/BestPracticesCard.tsx
```diff
@@ -0,0 +1,37 @@
...
+        onClick={() => setExpand(!expand)}
+        className={`h-6 w-6 absolute md:hidden top-6 right-5 ${
+          expand ? 'rotate-90' : '-rotate-90'
+        }`}
+      />
+      <div className="min-w-[50px] h-[50px] md:min-w-[100px] md:h-[100px] rounded-full bg-white flex items-center justify-center">
+        <Icon className="h-6 w-6 md:h-10 md:w-10 text-brand-dark-blue" />
+      </div>
+      <div>
+        <h3 className="text-xl md:text-[32px] md:leading-[41.6px] font-bold mb-4 pr-6 md:pr-0">
```
tglaz, line 28: md:leading-[`41.6px`] 🤣 

I'd suggest being conservative with css. We'll need this to work on older devices, so I think it'd be better to use standard values. Not sure we need that type of precision lol

apps/wildfires/src/app/shared/components/bestPractices/BestPracticesCard.tsx
```diff
@@ -0,0 +1,37 @@
...
+}
+
+export default function BestPracticesCard(props: IBestPracticesCardProps) {
+  const { Icon, title, description, bgColor } = props;
+  const [expand, setExpand] = useState(false);
+  return (
+    <div
+      className={`${bgColor} rounded-[20px] w-full p-6 md:p-10 flex items-center gap-4 mb-4 md:mb-10 relative`}
+    >
+      <ChevronLeftIcon
```
tglaz, line 18: not sure if it applies, but we have this: `ExpandableContainer`.
I didn't see the mocks so I've no idea if that helps, but just FYI

https://github.com/BetterAngelsLA/monorepo/blob/main/libs/react/components/src/lib/ExpandableContainer/ExpandableContainer.tsx

apps/wildfires/src/app/shared/components/bestPractices/BestPractices.tsx
```diff
@@ -0,0 +1,37 @@
+import { CameraIcon, WFAnnouncement } from '@monorepo/react/icons';
+
+import BestPracticesCard from './BestPracticesCard';
+
+export default function BestPractices() {
+  return (
+    <div className="px-4 md:px-10 w-full my-8 md:my-24">
```
mikefeldberg, line 7: 1. these cards shouldn't be full width
2. descriptions not displayed
3. too much bottom padding
4. this isn't from this ticket, and seems to be an app-wide issue, but the little `|` line at the start of each heading should only be as tall as the first line. I don't think this is an easy fix, so feel free to ignore. I'm just pointing it out in case it's easy

design
image


implementation
image

apps/wildfires/src/app/shared/components/bestPractices/BestPracticesCard.tsx
```diff
@@ -0,0 +1,37 @@
...
+        onClick={() => setExpand(!expand)}
+        className={`h-6 w-6 absolute md:hidden top-6 right-5 ${
```
mikefeldberg, line 20: It seems that the reason the descriptions are missing here (above comment) is because they're behind an expand. But this component shouldn't be expandable or have a chevron; the description should always be displayed. Also, I'm not sure why the chevron is hard-coded `hidden`, making it impossible to expand

mikefeldberg (changes requested):

apps/wildfires/src/app/shared/components/bestPractices/BestPracticesCard.tsx
Davit-BetterAngels: what design requires. it is what on the paper. 🤣

apps/wildfires/src/app/shared/components/bestPractices/BestPracticesCard.tsx
Davit-BetterAngels: by design the chevron is only visible on mobile. will fix on desktop

apps/wildfires/src/app/shared/components/bestPractices/BestPractices.tsx
Davit-BetterAngels: 1.fixed
2.fixed
3.fixed

apps/wildfires/src/app/shared/components/bestPractices/BestPracticesCard.tsx
Davit-BetterAngels: ready

mikefeldberg: gah. sorry! I didn't realize mobile had chevron 🤦

mikefeldberg (approved):

PR #904, opened 2025-01-25
firefox pdf bug (DEV-1360)

Davit-BetterAngels:
DEV-1360

## Summary by Sourcery

Bug Fixes:
- Fix PDF margin bug in Firefox.


mikefeldberg (commented):
this works for me, but mine also works on `main`, so it would be good if someone who experienced the original error verified that this fixes it

PR #960, opened 2025-01-28
Update wildfire hero image (DEV-1429)

mikefeldberg:
DEV-1429

## Summary by Sourcery

Update the hero image for the wildfires app.

Enhancements:
- Replace the existing hero image with a new one.

Documentation:
- Add attribution for the new hero image.


vecchp (approved):

tglaz (changes requested):
@mikefeldberg the mobile should be a .jpeg - should drop size to < 100K

tglaz (commented):
desktop can also be smaller - making it 2400px should drop it to ~250K

if we want to optimize further we could later serve it from sanity which would allow us to a responsive image `srcset`. sanity cms images work basically like imgix

tglaz (approved):

PR #976, opened 2025-01-29
make interaction authors endpoint (DEV-1215)

jluna2000:
* Added OffsetPaginated field to Note Query to return outreach authorized users

* Making InteractionAuthorType to return outreach authorized users, with test cases

* Adding NotePermissions to the InteractionAuthor endpoints in Notes queries

## Summary by Sourcery

Add interaction authors endpoint to retrieve outreach authorized users.

New Features:
- Added a new endpoint to retrieve interaction authors, along with pagination support.

Tests:
- Added tests for the new interaction author endpoint.


jluna2000:
Added 1 to the expected query count in the `test_delete_client_profile_mutation` test case since it was querying the newly added `InteractionAuthorType`.

apps/betterangels-backend/notes/tests/test_queries.py
```diff
@@ -191,6 +191,56 @@ def test_notes_paginated_query(self) -> None:
...
+                        firstName
+                        lastName
+                        middleName
+                    }
+                }
+            }
+        """
+        response = self.execute_graphql(query)
+
+        self.assertFalse(response["data"] is None, f"Data is None: {response}")
```
mikefeldberg, line 235: with this as the only assertion, this test may provide false security. That is, it's confirming that `data` isn't null, but it's not checking that the payload contains the expected data or is free of errors. 

I think this test would benefit from a test user object being created for it. That is, atm, the returned data is 
```
[{'id': '7', 'firstName': None, 'lastName': None, 'middleName': None},
{'id': '1', 'firstName': None, 'lastName': None, 'middleName': None},
{'id': '5', 'firstName': None, 'lastName': None, 'middleName': None},
{'id': '6', 'firstName': None, 'lastName': None, 'middleName': None},
{'id': '3', 'firstName': 'Carolyn', 'lastName': None, 'middleName': None}]
```
so you couldn't assert that at least one of the returned users has the expected first, last, middle names, because none of the existing users even have that info filled out. I think the thing to do here is:
1. create a test user with a first, last, middle name
2. add that user to an org, so it passes the "outreach authorized" check
3. add assertions to verify that that user is present in the payload

You can take a look [here](https://github.com/BetterAngelsLA/monorepo/blob/main/apps/betterangels-backend/common/tests/utils.py#L16-L60), to see how we create users and add them to orgs. But no worries if this doesn't make sense. We can catch up about it after stand up tomorrow or Wed

mikefeldberg (changes requested):
looking good! noted some improvements and clean up below

apps/betterangels-backend/notes/tests/test_queries.py
```diff
@@ -191,6 +195,45 @@ def test_notes_paginated_query(self) -> None:
...
+
+        test_new_interaction_author = test_user_map["test_new_interaction_author"]
+        test_new_interaction_author.first_name = "Wanda"
+        test_new_interaction_author.last_name = "Maximoff"
+        test_new_interaction_author.middle_name = "J."
+        test_new_interaction_author.save()
+
+        caseworker_permission_group_template = PermissionGroupTemplate.objects.get(name="Caseworker")
+        perm_group = permission_group_recipe.make(template=caseworker_permission_group_template)
+        perm_group.organization.add_user(test_new_interaction_author)
```
mikefeldberg, line 225: you don't need to create a template here. `get_outreach_authorized_users` filters on `PermissionGroup` name, so you just need a permission group with the right name
```suggestion
        perm_group = permission_group_recipe.make(name=GroupTemplateNames.CASEWORKER)
        perm_group.organization.add_user(case_manager)
```

apps/betterangels-backend/notes/tests/test_queries.py
```diff
@@ -191,6 +195,45 @@ def test_notes_paginated_query(self) -> None:
...
+
+        response = self.execute_graphql(query)
+
+        self.assertFalse(response["data"] is None, f"Data is None: {response}")
+
+        firstNames = [result["firstName"] for result in response["data"]["interactionAuthors"]["results"]]
+
+        self.assertTrue("Wanda" in firstNames)
+
+        self.assertFalse(self.client_user_1.first_name in firstNames)
```
mikefeldberg, line 235: `next` will return the first user that matches the condition (`u["id"] == str(case_manager.pk)`), and avoid iterating over the entire list. This approach also gets back the entire user dict for the target user, rather than just a list of names. This way, you can assert that all the fields returned as expected

```suggestion
        results = response["data"]["interactionAuthors"]["results"]
        returned_author: dict = next((u for u in results if u["id"] == str(case_manager.pk)))

        self.assertEqual(returned_author["firstName"], "Wanda")
        self.assertEqual(returned_author["lastName"], "Maximoff")
        self.assertEqual(returned_author["middleName"], "J.")
```

apps/betterangels-backend/accounts/utils.py
```diff
@@ -45,3 +46,18 @@ def get_user_permission_group(user: Union[AbstractBaseUser, AnonymousUser]) -> P
         raise PermissionError("User lacks proper organization or permissions")
 
     return permission_group
+
+
+def get_outreach_authorized_users() -> Union[User, QuerySet[User]]:
```
mikefeldberg, line 51: since we removed `interaction_author`, this function will only ever return a queryset
```suggestion
def get_outreach_authorized_users() -> QuerySet[User]:
```

apps/betterangels-backend/accounts/utils.py
```diff
@@ -45,3 +46,18 @@ def get_user_permission_group(user: Union[AbstractBaseUser, AnonymousUser]) -> P
...
+    # Subquery to check if the user has any related permission group in an authorized group
+    permission_group_exists = PermissionGroup.objects.filter(
+        organization__users=OuterRef("pk"),  # Matches `User` to `Organization`
+        template__name__in=authorized_permission_groups,
+    )
+
+    # Use Exists to avoid duplicate users without `distinct()`
+    outreach_authorized_users = User.objects.filter(Exists(permission_group_exists))
+
+    return outreach_authorized_users
```
mikefeldberg, line 63:
```suggestion
    return User.objects.filter(Exists(permission_group_exists))
```

apps/betterangels-backend/notes/types.py
```diff
@@ -286,3 +287,11 @@ class UpdateTaskLocationInput:
 class RevertNoteInput:
     id: ID
     revert_before_timestamp: datetime
+
+
+@strawberry_django.type(User)
+class InteractionAuthorType:
+    id: ID  # type: ignore
```
mikefeldberg, line 294: I don't think you need the ignore here

apps/betterangels-backend/notes/schema.py
```diff
@@ -90,6 +92,12 @@ class Query:
 
     tasks: List[TaskType] = strawberry_django.field(extensions=[HasRetvalPerm(TaskPermissions.VIEW)])
 
+    @strawberry_django.offset_paginated(
+        OffsetPaginated[InteractionAuthorType], extensions=[HasPerm(NotePermissions.ADD)]
+    )
+    def interaction_authors(self) -> QuerySet[User]:
+        return cast(QuerySet, get_outreach_authorized_users())
```
mikefeldberg, line 99: `get_outreach_authorized_users` returns a queryset, so no need to cast 
```suggestion
        return get_outreach_authorized_users()
```

apps/betterangels-backend/accounts/utils.py
mikefeldberg, line 51: we should add a unit test for this function. or a `# TODO` to add one in the near future

apps/betterangels-backend/notes/tests/test_queries.py
```diff
@@ -191,6 +195,45 @@ def test_notes_paginated_query(self) -> None:
...
+
+        test_user_map = {
+            "test_new_interaction_author": baker.make(User, username=f"test_new_interaction_author_{uuid.uuid4()}")
+        }
+
+        test_new_interaction_author = test_user_map["test_new_interaction_author"]
+        test_new_interaction_author.first_name = "Wanda"
+        test_new_interaction_author.last_name = "Maximoff"
+        test_new_interaction_author.middle_name = "J."
+        test_new_interaction_author.save()
```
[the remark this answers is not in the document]
mikefeldberg: we can meet and I can explain why we use that user label map in the base test case

apps/betterangels-backend/notes/schema.py
```diff
@@ -90,6 +92,12 @@ class Query:
 
     tasks: List[TaskType] = strawberry_django.field(extensions=[HasRetvalPerm(TaskPermissions.VIEW)])
 
+    @strawberry_django.offset_paginated(
+        OffsetPaginated[InteractionAuthorType], extensions=[HasPerm(NotePermissions.ADD)]
```
mikefeldberg, line 96: can you please add a permission test or a `TODO` that we need to add one?

PR #979, opened 2025-01-29
add RV to living situation enum

bckohan:
It is important to differentiate between RVs and Vehicles. For example, LA City only has one tow truck that can tow RVs.

It might be tempting to use the county PIT dwelling types here, but I think its fine to not differentiate between cars/vans and makeshifts/tents.

## Summary by Sourcery

New Features:
- Add recreational vehicle (RV) as a living situation option.


apps/betterangels-backend/clients/enums.py
```diff
@@ -89,6 +89,7 @@ class LivingSituationEnum(models.TextChoices):
     SHELTER = "shelter", _("Shelter")
     TENT = "tent", _("Tent")
     VEHICLE = "vehicle", _("Vehicle")
+    RV = "rv", _("RV")
```
mikefeldberg, line 92: please alphabetize list

mikefeldberg, line 92: after updating a schema or type (including the options available for a given input), you need to run `yarn nx affected -t generate-graphql-schema` to regen the schema and `yarn nx affected -t generate-graphql-types` to regen the types. alternatively `ynx-generate` will do both of the above

mikefeldberg (changes requested):

PR #1016, opened 2025-02-05
Update CSRF Link Handling

vecchp:
## Summary by Sourcery

Refactor CSRF token handling to use a centralized manager and improve storage logic.

Enhancements:
- Centralize CSRF token management to avoid redundant fetches and storage.
- Store CSRF tokens based on the API origin to prevent conflicts.
- Improve token extraction from GraphQL and REST responses.

Tests:
- Update tests to reflect the new CSRF handling logic.

### DIFF
--- a/libs/expo/betterangels/src/lib/ui-components/LoginForm/index.tsx
+++ b/libs/expo/betterangels/src/lib/ui-components/LoginForm/index.tsx
@@ -83,6 +83,7 @@ export default function LoginForm({
       }
     } catch (error) {
       setErrorMessage('Something went wrong. Please try again.');
+      switchEnvironment('production');
     }
     setIsLoading(false);
   };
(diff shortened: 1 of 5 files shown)


mikefeldberg (commented):
works great in local, but there's one issue in dev client: if I try to log in with an @example.com account, and that fails, and then I try to do oauth, oauth hangs. I think it's because of this code, in `libs/expo/betterangels/src/lib/ui-components/LoginForm/index.tsx`

```
  const handleUsernameChange = (newUsername: string) => {
    setUsername(newUsername);
    if (checkIfDemoMode(newUsername)) {
      console.log('Switching to Demo API');
      switchEnvironment('demo');
    } else {
      console.log('Switching to Production API');
      switchEnvironment('production');
    }
  };

  const handleLogin = async () => {
    ...

      if (error) {
        setErrorMessage('Something went wrong. Please try again.');
      } else if (!loading && data && 'login' in data) {
        refetchUser();
      } else {
        setErrorMessage('Either email or password is incorrect.');
      }
    } catch (error) {
      setErrorMessage('Something went wrong. Please try again.');
    }
    setIsLoading(false);
  };
```

if demo login fails, `handleLogin` doesn't switch the env back to prod. so the subsequent oauth login attempt tries to hit the local server. I think this would fix it

i.e.,
```
    } catch (error) {
      setErrorMessage('Something went wrong. Please try again.');
      switchEnvironment('production')         <----------------
    }
```

if, after a failed @example.com login attempt, I type anything in the email input, and then do oauth, oauth works. because

mikefeldberg (approved):
seamless. thanks for fixing this 🙏

PR #1044, opened 2025-02-12
update veteran options (DEV-1496)

mikefeldberg:
DEV-1496

this is an unfortunately elaborate process because we have to swap out the enum without breaking the be/fe contract. so, this PR is step 1 of 5 😓 

1. backend side by side
    1. add `VeteranStatusEnum`
    2. add `temp_veteran_status` field to `ClientProfileType/Input`
    3. update `create` and `update` mutations to use `temp_veteran_status` to write `veteran_status` if the former is available
    4. update tests
2. frontend cutover 1
    1. update query/mutation to read/write `temp_veteran_status`
    3. update components to use `VeteranStatusEnum` choices
3. backend cutover
    1. update `veteran_status` to match `temp_veteran_status`
4. frontend cutover 2
    1. revert `tempVeteranStatus` to `veteranStatus` everywhere
5. backend clean up
    1. remove all uses of `temp_veteran_status`
        1. field def & input fields
        2. create & update mutations
        3. tests

## Summary by Sourcery

Introduce a new `VeteranStatusEnum` and a temporary `temp_veteran_status` field to `ClientProfile`. Update create and update mutations to use `temp_veteran_status` to write `veteran_status`. Update tests accordingly.

New Features:
- Add a new `VeteranStatusEnum` to replace the existing `YesNoPreferNotToSayEnum` for veteran status.

Tests:
- Update tests to reflect the changes in the client profile mutations and queries.


apps/betterangels-backend/clients/tests/test_queries.py
```diff
@@ -102,6 +102,8 @@ def test_client_profile_query(self) -> None:
             "spokenLanguages": [LanguageEnum.ENGLISH.name, LanguageEnum.SPANISH.name],
             "user": self.client_profile_1["user"],
             "veteranStatus": YesNoPreferNotToSayEnum.NO.name,
```
tglaz, line 104: could this be changed to use `VeteranStatusEnum`?

apps/betterangels-backend/clients/types.py
```diff
@@ -222,7 +224,7 @@ class ClientProfileBaseType:
     race: auto
     residence_address: auto
     spoken_languages: Optional[List[LanguageEnum]]
-    veteran_status: auto
+    veteran_status: Optional[YesNoPreferNotToSayEnum]
```
tglaz, line 227: not sure if we can use `VeteranStatusEnum` here?

tglaz (approved):

apps/betterangels-backend/clients/tests/test_queries.py
mikefeldberg: I wish. Then we wouldn't have to play this multi-phase game of switcheroo. The fe expects this field to be a `YesNoPreferNotToSayEnum`. Even if the underlying options are identical, swapping this for `VeteranStatusEnum` will cause a type error

apps/betterangels-backend/clients/types.py
mikefeldberg: same as above. 

> 1. backend side by side
>     1. add `VeteranStatusEnum`       `<------------------ we are here`
>     2. add `temp_veteran_status` field to `ClientProfileType/Input`
>     3. update `create` and `update` mutations to use `temp_veteran_status` to write `veteran_status` if the former is available
>     4. update tests
> 2. frontend cutover 1
>     1. update query/mutation to read/write `temp_veteran_status`
>     3. update components to use `VeteranStatusEnum` choices       `<------------------ we need to get here`

apps/betterangels-backend/clients/tests/test_queries.py
tglaz: sounds good. since they're technically strings my first thought was "maybe", but guess not.
thx

PR #1050, opened 2025-02-13
use clients paginated query and fix list rerender (DEV-1155, DEV-1486)

mikefeldberg:
DEV-1155
DEV-1486

~also partly fixed the duplicate key error. It still happens when performing certain actions, but not when loading more clients. One pattern I've seen while working on this: press client card -> edit client -> delete client & return to list. I'm sure there are other ways to trigger it, but that's one I observed while working on this. Seems like a cache management issue~

slack thread re design changes & removal of `SectionList`: https://betterangels-group.slack.com/archives/C05ECN9FLGL/p1739399728703899

## Summary by Sourcery

New Features:
- Added pagination to the client lists on the Home and Clients screens.


libs/expo/betterangels/src/lib/screens/Clients/index.tsx
```diff
@@ -116,55 +121,22 @@ export default function Clients({ Logo }: { Logo: ElementType }) {
...
-      });
+    const { results, totalCount } = data.clientProfilesPaginated;
+    setTotalCount(totalCount);
 
-      return mergedClients;
-    });
+    if (offset === 0) {
+      setClients(results);
+    } else {
+      setClients((prevClients) => [...prevClients, ...results]);
```
tglaz, line 134: could we just do 
```javascript
setClients((prev) => {
      const prevClients = prev || [];

      return [...prevClients, ...results]
});
```

libs/expo/betterangels/src/lib/screens/Clients/index.tsx
```diff
@@ -224,12 +208,15 @@ export default function Clients({ Logo }: { Logo: ElementType }) {
             </TextRegular>
           </View>
         )}
-        {hasClients && (
-          <SectionList
+        {clients && (
```
tglaz, line 211: let's change to `!!clients` or `!!clients.length` so it's always a `bool`

libs/expo/betterangels/src/lib/screens/Clients/index.tsx
```diff
@@ -4,40 +4,40 @@ import {
...
-}
-
 export default function Clients({ Logo }: { Logo: ElementType }) {
-  const { title, select } = useLocalSearchParams();
-  const [search, setSearch] = useState<string>('');
-  const [filterSearch, setFilterSearch] = useState<string>('');
-  const [createNote] = useCreateNoteMutation();
+  const [currentClient, setCurrentClient] =
+    useState<
+      ClientProfilesPaginatedQuery['clientProfilesPaginated']['results'][number]
```
tglaz, line 28: not blocking, but perhaps clean clean up the typings a bit?

// at top of file
```
type TClientProfile = ClientProfilesPaginatedQuery['clientProfilesPaginated']['results'][number];
```
// in code
```
const [currentClient, setCurrentClient] = useState<TClientProfile>()
```
optionally could do same for [clients, setClients]

libs/expo/betterangels/src/lib/screens/Clients/index.tsx
```diff
@@ -4,40 +4,40 @@ import {
...
-import { SectionList, View } from 'react-native';
+import { FlatList, View } from 'react-native';
 import { Ordering } from '../../apollo';
 import { useSnackbar } from '../../hooks';
 import { ClientCard, ClientCardModal, Header } from '../../ui-components';
 import {
-  ClientProfilesQuery,
-  useClientProfilesQuery,
+  ClientProfilesPaginatedQuery,
+  useClientProfilesPaginatedQuery,
```
tglaz, line 19: just curious if you were to determine why using a shared query did not work?
still seems a bit weird we can't reuse it, though frankly not exactly sure either.

libs/expo/betterangels/src/lib/screens/Clients/index.tsx
```diff
@@ -49,17 +49,22 @@ export default function Clients({ Logo }: { Logo: ElementType }) {
...
-  function loadMoreClients() {
-    if (hasMore && !loading) {
+  async function loadMoreClients() {
+    if (hasMore && !loading && !isLoadingMore) {
+      setIsLoadingMore(true);
       setOffset((prevOffset) => prevOffset + paginationLimit);
+
+      setTimeout(() => {
+        setIsLoadingMore(false);
+      }, 600);
```
tglaz, line 67: do we really want to use a `setTimeout` and `isLoadingMore`? 

perhaps should just use `useClientProfilesPaginatedLazyQuery` and then rely on `loading` only>

tglaz (changes requested):

libs/expo/betterangels/src/lib/screens/Clients/index.tsx
mikefeldberg: tl;dr: `cache-and-network` caused clients already fetched by one component to be refetched by the other component, despite already being in cache. see: https://betterangels-group.slack.com/archives/C05RETRGDA7/p1739923998208249

mikefeldberg:
replaced by #1064

PR #1065, opened 2025-02-19
Filter by author (DEV-1127)

Davit-BetterAngels:
DEV-1127

## Summary by Sourcery

This pull request introduces the ability to filter interactions by author, enhancing the filtering capabilities of the Interactions screen. It adds a new `AuthorsFilter` component and integrates it into the `InteractionsFilters` component, allowing users to filter interactions based on the author.

New Features:
- Introduces the ability to filter interactions by author.

Enhancements:
- Updates the Interactions screen to include author filtering capabilities.
- Improves the Interactions filter component to include both team and author filters.


apps/betterangels/src/app/(private-screens)/add-note/[noteId].tsx
```diff
@@ -98,7 +98,7 @@ export default function AddNote() {
         variables: {
           pagination: { limit: 10 + 1, offset: 0 },
```
Comment (shiozakiyuko) [inline apps/betterangels/src/app/(private-screens)/add-note/[noteId].tsx:99]: Hi @Davit-BetterAngels 

I see some discrepancy between iOS and Android and on each device from ACs. 

##**iOS**

1. In the Interactions page, "Me" shows only 1 interactions. -> This default should show all the interactions I/the user wrote (Android shows correctly.)
2. When I open the filter, I see some blanks in the selections (Attachment iOS No1)
3. I filtered my name "Yuko Shiozaki" and I found 2 Yukos (Adam has 2, Andre has 2, but not Farima for example)
4. Between 2 Yukos, Yuko1 has 75 interactions, and Yuko2 has 63 interactions. (Attachment iOS No2 and 3)
5. Between the filters and the line of "Displaying 10 of xxx interactions" is too tall (this should be 40px in between)







##**Android**

As for Android, they have very similar but not exactly same issues. 

1. On Android, it shows the total # of my interactions that I created (I believe). 
2. I see blank checkbox items in Filter like iOS.
3. When I filtered, I see 2 Yukos like on iOS but the number of interactions are different. Yuko1 shows 75, Yuko2 shows 0, and Me shows 63.
4. The UI issue between the filters and "Displaying 10 out of xxx" is same. Please adjust the height like I mentioned in iOS section. 

Sorry I wanted to attach some screen shots from Android but the app clashed on me and I can't reload atm. 
Let me know if you want me to clarify anything from my comments. Thanks!

Comment (kimchoi-jjiggae) [reply apps/betterangels/src/app/(private-screens)/add-note/[noteId].tsx:99]: Also, select all only selects a subset of all authors

kimchoi-jjiggae:
Select all still only selects the loaded authors, not all

apps/betterangels/src/app/(private-screens)/add-note/[noteId].tsx
Comment (Davit-BetterAngels) [reply apps/betterangels/src/app/(private-screens)/add-note/[noteId].tsx:99]: 1. I guess we talked that "Me" should always stay at top.

2. its not the authors name but Client

Comment (shiozakiyuko) [reply apps/betterangels/src/app/(private-screens)/add-note/[noteId].tsx:99]: @Davit-BetterAngels , 
As for 1, I will ask Michelle what she wants... (if I miss the convo about it somewhere in the past, my apology in the advance)
As for 2, yeah I noticed that. But what I'm asking is why the outcomes are different based on the name I use for the filter...

libs/expo/betterangels/src/lib/screens/Interactions/InteractionsFilters/AuthorsFilter.tsx
```diff
@@ -0,0 +1,176 @@
...
+            showsVerticalScrollIndicator={false}
+          >
+            <MultiSelect
+              filterPlaceholder="Search"
+              withFilter
+              title="Filter - Authors"
+              search={search}
+              onSearch={onSearch}
+              onChange={(e: { id: string; label: string }[]) => setSelected(e)}
+              options={[{ id: user.id, label: 'Me' }, ...authorsState]}
```
mikefeldberg, line 153: @Davit-BetterAngels what about something like this? 
```suggestion
              options={search ? [{ id: user.id, label: 'Me' }] : [...authorsState]}
```

mikefeldberg: actually... if we're always showing the list of authors, it would be 
```suggestion
              options={
                search === ''
                  ? [{ id: user.id, label: 'Me' }, ...authorsState]
                  : [...authorsState]
              }
```

apps/betterangels/src/app/(private-screens)/add-note/[noteId].tsx
Comment (mikefeldberg) [reply apps/betterangels/src/app/(private-screens)/add-note/[noteId].tsx:99]: @shiozakiyuko re: item 2 - the card is displaying your name because you're the client. There are five "Yuko Shiozaki" users in the dev db. You're logged in as one and looking at interactions you created for another

mikefeldberg (commented):
just a few small things

libs/expo/betterangels/src/lib/screens/Interactions/InteractionsFilters/AuthorsFilter.tsx
```diff
@@ -0,0 +1,154 @@
...
+};
+
+interface IAuthorsFilterProps {
+  setFilters: (filters: TFilters) => void;
+  filters: TFilters;
+}
+
+const paginationLimit = 25;
+
+export default function AuthorsFilter(props: IAuthorsFilterProps) {
```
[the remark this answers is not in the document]
mikefeldberg: @Davit-BetterAngels thoughts? This seems like a good idea. Not for this PR, but it would probably be good to make a ticket to extract this into a shared component

apps/betterangels/src/app/(private-screens)/add-note/[noteId].tsx
Comment (mikefeldberg) [reply apps/betterangels/src/app/(private-screens)/add-note/[noteId].tsx:99]: I updated the name of two interactions so they would be easier to find in admin. You are the "client" on both of these.
image
image

You're right that your users don't have associated ClientProfiles, but a user doesn't need to have a ClientProfile in order to have an interaction. I'm not exactly sure how those interactions were created for your user, but I have a guess: we added the ClientProfile model in April 17, 2024. These interactions are all from early May. It's likely that these interactions were created _before_ we we started using the ClientProfile model on the front end, when the "clients" list was still just showing users.

Comment (mikefeldberg) [reply apps/betterangels/src/app/(private-screens)/add-note/[noteId].tsx:99]: (there are 75 of these)
image

libs/expo/betterangels/src/lib/hooks/scroll/useInfiniteScroll.ts
```diff
@@ -0,0 +1,35 @@
+// useInfiniteScroll.ts
```
mikefeldberg, line 1: copypasta

mikefeldberg (approved):

PR #1070, opened 2025-02-20
view pdf via expo-web-browser (DEV-1116)

tglaz:
https://betterangels.atlassian.net/browse/DEV-1116

- switches to rendering PDF in `WebBrowser` (from 'expo-web-browser')
  - `react-native-pdf` does not work fully with expo (not technically supported) and breaks in android.
  - `react-native-webview` has similar issues for Android, where a loaded pdf would probably need to be shown in an iframe. It's good for rendering html in webview, but not files via url
  - other pdf solutions have questionable support
  - Expo Go seems to be a limitation
  - the browser solution seems the most stable
    - the android experience is still sub-optimal due to android ux

## Summary by Sourcery

Enhancements:
- Replaced the PDF viewer component with a link that opens the PDF in the Expo web browser, improving stability across platforms.

### DIFF
--- a/libs/expo/betterangels/src/lib/screens/FileScreenComponent/index.tsx
+++ b/libs/expo/betterangels/src/lib/screens/FileScreenComponent/index.tsx
@@ -3,7 +3,6 @@ import {
   BaseModal,
   ImageViewer,
   Loading,
-  PdfViewer,
   TextBold,
   TextRegular,
 } from '@monorepo/expo/shared/ui-components';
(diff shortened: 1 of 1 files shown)


shiozakiyuko (commented):
Hi @tglaz 

### **iOS and Android**
Please find a way to replace the title of the domain to the file name.
Atm, we see "cdn.dev.betterangels.la" on the header of the View File. 
Please see the attachment.





### **Android**
Only for Android, I don't see the file name but I see something else on the file when it's downloading and visible on the browser. Please see the attachment.

tglaz:
closing in favor of #1061

PR #1071, opened 2025-02-20
surface client profile validation errors (DEV-1161)

mikefeldberg:
DEV-1161

## Summary by Sourcery

Surfaces client profile validation errors to the user when creating or updating a client profile. This includes displaying errors for email, phone number, and HMIS ID fields.

Bug Fixes:
- Fixes an issue where client profile validation errors were not being displayed to the user.
- Fixes an issue where HMIS ID errors were not being displayed correctly.

Enhancements:
- Improves the email input fields by disabling auto-capitalization and auto-correction.


libs/expo/betterangels/src/lib/screens/AddEditClient/index.tsx
```diff
@@ -57,6 +63,38 @@ const defaultSocialMedias = [
...
+    if (error.location) {
+      if (error.location.includes('__')) {
+        const [index, subField] = error.location.split('__');
+        fieldKey = `${error.field}[${index}].${subField}`;
+      } else {
+        fieldKey = `${error.field}.${error.location}`;
+      }
+    } else {
+      fieldKey = error.field;
+    }
```
tglaz, line 90: Not sure if this version below is preferable, but guess i've a thing against `else` statements 🤷 
But your call

```javascript
let fieldKey = error.field;

if (error.location) {
  fieldKey = `${error.field}.${error.location}`;

  if (error.location.includes('__')) {
     const [index, subField] = error.location.split('__');

     fieldKey = `${error.field}[${index}].${subField}`;
  }
}

```

libs/expo/betterangels/src/lib/screens/AddEditClient/index.tsx
```diff
@@ -183,17 +222,40 @@ export default function AddEditClient({ id }: { id?: string }) {
...
+        Object.entries(formErrors).forEach(([key, message]) => {
+          methods.setError(
+            key as keyof (UpdateClientProfileInput | CreateClientProfileInput),
+            {
+              type: 'manual',
+              message,
+            }
+          );
+        });
+        return;
```
tglaz, line 256: Seems like this `return` statement means the code won't continue to the block below. 
Would it be possible to get both `operationErrors` and errors inside `operationResult` returned together?

libs/expo/betterangels/src/lib/screens/AddEditClient/index.tsx
```diff
@@ -183,17 +222,40 @@ export default function AddEditClient({ id }: { id?: string }) {
...
+          errorPolicy: 'all',
         });
         operationResult = createResponse.data?.createClientProfile;
+        operationErrors = createResponse.errors?.[0].extensions?.['errors'] as
+          | TValidationError[]
+          | undefined;
       }
+
+      if (operationErrors) {
+        const formErrors = parseValidationErrors(operationErrors);
```
tglaz, line 246: curious if it'd be worth wrapping this in a try/catch, but then not really sure how best to respond after.
guess main comment would be that we'd want thorough manual testing on this, and then will be rewriting these forms soon anyhow

libs/expo/betterangels/src/lib/helpers/parseClientProfileErrors.ts
```diff
@@ -0,0 +1,47 @@
...
+        fieldKey = `${error.field}[${index}].${subField}`;
+      }
+    }
+
+    formErrors[fieldKey] = message;
+  });
+  return formErrors;
+}
+
+export function applyValidationErrors<T>(
```
tglaz, line 39: is this meant to be called somewhere in this PR or would be applied separately?

libs/expo/betterangels/src/lib/helpers/parseClientProfileErrors.ts
```diff
@@ -0,0 +1,47 @@
+export type TValidationError = {
```
shiozakiyuko, line 1: Hi @mikefeldberg 

On both, iOS and Android, I don't see the error message is visible when I attempted to add a duplicate HMIS ID.

The example I used is LAHSA 1234gyat from Alissa Wonder and I entered the exactly same one to Egor Ivanov.
It just ignores what I entered and goes back to default setting. No error message is happening (I'm assuming it shows underneath of the textbox?
Let me know if I'm misunderstanding the AC.

kimchoi-jjiggae: Android:
I also don't see the error message specified by the AC's - I just see the generic red bar that says "Sorry there was an error updating this profile."

iOS: I don't see any error at all

libs/expo/betterangels/src/lib/screens/AddEditClient/index.tsx
mikefeldberg: love that

libs/expo/betterangels/src/lib/screens/AddEditClient/index.tsx
mikefeldberg: I removed `operationResult`. We were using it to parse errors, and all errors are now returned by `operationErrors`

libs/expo/betterangels/src/lib/helpers/parseClientProfileErrors.ts
mikefeldberg: imported and used the wrong function. fixed

libs/expo/betterangels/src/lib/helpers/parseClientProfileErrors.ts
shiozakiyuko: Hi @mikefeldberg 

iOS looks good. As for Android, it's taking forever to load the app, and I finally got... the red error screen...
Please see the attachment.

Thanks!

shiozakiyuko (approved):

libs/expo/betterangels/src/lib/helpers/parseClientProfileErrors.ts
shiozakiyuko: I installed the latest build for my Android and retested it. After readjusting my wifi speed, everything works correctly. Approved.

libs/expo/betterangels/src/lib/screens/AddEditClient/PersonalInfo/CaliforniaId.tsx
```diff
@@ -43,14 +43,24 @@ export default function CaliforniaId() {
   }
 
   useEffect(() => {
-    if (featureAvailable(californiaId) && uniqueCheckError) {
-      // If keyboard not dismissed, modal does not render fullscreen on some Android devices.
-      Keyboard.dismiss();
+    if (!californiaId || californiaId.length < 8) {
```
tglaz, line 46: can you give context for this line? can't really tell why we'd clear error here.

we already have a regex validation for `californiaId` inside the input and this has different criteria

libs/expo/betterangels/src/lib/screens/AddEditClient/index.tsx
```diff
@@ -183,45 +190,32 @@ export default function AddEditClient({ id }: { id?: string }) {
...
+          | undefined;
+
+        if (!operationErrors) {
+          refetch();
+        }
       } else {
         const input = values as CreateClientProfileInput;
         const createResponse = await createClient({
           variables: { data: input as CreateClientProfileInput },
+          errorPolicy: 'all',
```
tglaz, line 207: any specific reason for this change?
if there are errors, do we do something with the data?

tglaz (changes requested):

libs/expo/betterangels/src/lib/screens/AddEditClient/index.tsx
mikefeldberg: without `errorPolicy: 'all'`, the errors returned by the backend are returned flat, and can't be accessed or parsed

libs/expo/betterangels/src/lib/screens/AddEditClient/PersonalInfo/CaliforniaId.tsx
mikefeldberg: **before**
- CA ID unique check only triggered on submit
- clearing the field doesn't clear the error; "Submit" remains disabled until the user touches the CA ID field again

https://github.com/user-attachments/assets/c2d35c56-937d-4bc3-b70a-d16d01a5e17e

---

**after**
- CA IS unique check triggered as soon as field is fully filled (8 chars long)
- clearing field clears the error

https://github.com/user-attachments/assets/e007c03a-68b2-492f-aba1-1561d7ce6b6b

---

afaict, this should have been enabled as of DEV-47
> Do NOT allow submit until the user changes the CA ID to a number that is unique or blank.

tglaz: If it's a matter of revalidating on pressing `clear`, would this solve the issue?

```javascript
        rules={{
          validate: (value: string) => {
            if (value === '') {
              return true;
            }
            ....
```

I'd prefer to keep field validation logic in one place if possible

mikefeldberg:
```js
            if (value === '') {
              return true;
            }
```
doesn't solve the issue. what would solve the issue is removing 
```js
            if (uniqueCheckError) {
              return uniqueCheckError;
            }
```
from `rules`. 

Regarding keeping validation logic in one place, afaiu, my `useEffect` change is implementing `uniqueCheckError`, as specified in DEV-47. Am I missing something there?

tglaz (approved):

PR #1072, opened 2025-02-20
add duplicate and invalid cali id error codes (DEV-1161)

mikefeldberg:
DEV-1161 (ok, not really, but we're already here)

## Summary by Sourcery

Tests:
- Added a test case to verify that an error is raised when a duplicate California ID is used.

### DIFF
--- a/apps/betterangels-backend/clients/schema.py
+++ b/apps/betterangels-backend/clients/schema.py
@@ -1,3 +1,4 @@
+import re
 from typing import Any, Dict, List, Optional, cast
 
 import phonenumber_field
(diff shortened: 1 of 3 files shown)


apps/betterangels-backend/clients/schema.py
```diff
@@ -104,6 +105,22 @@ def _validate_user_name(user_data: dict, nickname: str, user: Optional[User] = N
     return errors
 
 
+def _validate_california_id(california_id: str) -> list[dict[str, Any]]:
+    errors = []
+
+    california_id_pattern = r"^[A-Z]\d{7}$"
```
tglaz, line 111: since these are user entered, wouldn't we want it to be case insensitive?

tglaz (changes requested):

tglaz (approved):

PR #1074, opened 2025-02-21
fix client profile error handling backend (DEV-1534)

mikefeldberg:
DEV-1534

- fixed client profile validation utils that were breaking the create and update mutations
- added unit tests for the validation utils

## Summary by Sourcery

This pull request fixes and improves the client profile validation process. It corrects validation logic, properly handles unset values, and includes new unit tests to ensure the validation utilities function correctly.

Bug Fixes:
- Fixed client profile validation utils to prevent errors during create and update mutations by correcting validation logic and handling of unset values.

Tests:
- Added unit tests for the client profile validation utilities to ensure correct validation behavior.

PR #1075, opened 2025-02-23
remove ca id backend check (DEV-1534)

mikefeldberg:
DEV-1534

reverts #1072. removes the ca-id check that's breaking client profile mutations

more updates to backend error handling to come in #1074


tglaz (approved):
seems like california_id_pattern regex check is happening on model level already

PR #1078, opened 2025-02-25
update and test client profile backend validation (DEV-1538)

mikefeldberg:
DEV-1538

- update `User` save:
  - `email` always converted to lowercase; empty string saved as `None`
- update `ClientProfile` save:
  - `california_id` always converted to lowercase; empty string saved as `None`
- add regex constants for email, phone number, california ID
- add validation functions for all client profile data
  - payload contains at least one name field
  - `email` regex and unique constraint
  - `california_id` regex and unique constraint
  - `hmisId` completeness and unique constraint
  - client phone numbers
  - relevant contact phone numbers


apps/betterangels-backend/clients/enums.py
```diff
@@ -3,6 +3,17 @@
 from django.utils.translation import gettext_lazy as _
 
 
+class ErrorCodeEnum(models.TextChoices):
+    CA_ID_INVALID = "ca_id_invalid", _("California ID invalid")
+    CA_ID_IN_USE = "ca_id_in_us", _("California ID is already in use")
```
lena, line 8:
```suggestion
    CA_ID_IN_USE = "ca_id_in_use", _("California ID is already in use")
```

apps/betterangels-backend/clients/schema.py
```diff
@@ -67,60 +65,124 @@ def _format_graphql_error(error: Exception) -> str:
     return str(error)
 
 
-def _validate_user_email(user_data: dict, user: Optional[User] = None) -> list[dict[str, Any]]:
+def value_is_set(value: Optional[str]) -> bool:
+    return not (value is strawberry.UNSET or value is None or value.strip() == "")
```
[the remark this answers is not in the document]
lena: i would probably replace this with: 
```
all((value is not strawberry.UNSET, value is not None, value != ""))
```
and i don't know if i would strip here since i think the strip should happen elsewhere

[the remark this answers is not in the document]
tglaz: so you decided to leave `value.strip()`?

apps/betterangels-backend/clients/schema.py
```diff
@@ -67,60 +65,124 @@ def _format_graphql_error(error: Exception) -> str:
...
+                    "location": f"{idx}__number",
+                    "errorCode": ErrorCodeEnum.PHONE_NUMBER_INVALID.name,
+                }
+            )
 
     return errors
 
 
-def _validate_hmis_profiles(hmis_profiles: list[dict[str, Any]]) -> list[dict[str, Any]]:
+def validate_hmis_profiles(hmis_profiles: list[dict[str, Any]]) -> list[dict[str, Any]]:
```
lena, line 168: i think this function should take a user since you're looking to exclude existing user but don't pull the HMIS id off of them

mikefeldberg: the hmis_profile has its own id
```python
        ...
        hmis_profile_id = {"id": hmis_profile["id"]} if hmis_profile.get("id") else {}

        if (
            HmisProfile.objects.exclude(**hmis_profile_id)
            .filter(
                agency=hmis_profile["agency"],
                hmis_id__iexact=hmis_profile["hmis_id"],
            )
            .exists()
        ):
            ...
```

apps/betterangels-backend/clients/schema.py
```diff
@@ -67,60 +65,124 @@ def _format_graphql_error(error: Exception) -> str:
...
     for idx, hmis_profile in enumerate(hmis_profiles):
-        hmis_profile_id = {"id": hmis_profile["id"]} if hmis_profile.get("id") is not strawberry.UNSET else {}
+        hmis_id = hmis_profile.get("hmis_id")
+
+        if not value_is_set(hmis_id):
+            errors.append(
+                {
+                    "field": "hmisProfiles",
+                    "location": f"{idx}__hmisId",
+                    "errorCode": ErrorCodeEnum.HMIS_ID_NOT_PROVIDED.name,
```
lena, line 179: what if the agency instead of the id is not set?

apps/betterangels-backend/clients/schema.py
```diff
@@ -67,60 +65,124 @@ def _format_graphql_error(error: Exception) -> str:
...
+    # If the payload contains at least one name field, the data is valid
+    if payload_has_name:
         return errors
 
-    email = user_data["email"].lower()
+    # If the payload isn't clearing all existing name fields, the data is valid
+    if user:
+        client_has_name = any(
+            (
+                user.first_name and user_data.get("first_name") is strawberry.UNSET,
```
tglaz, line 92: I think this function can become a bit unclear, because what if `user_data.get("first_name") === " "`?
Wouldn't the first line evaluate to `True`?
So while this case may be covered by the `payload_has_name` check above, it gets me more confused than I'd like. 

Is the `UNSET` check necessary here? Seems like if `value_is_set(user_data.get("first_name"))` then this code would be unreachable

apps/betterangels-backend/clients/tests/test_mutations.py
```diff
@@ -421,6 +415,7 @@ def test_update_client_profile_null_email(self) -> None:
         self.assertEqual(client_profile_user["email"], None)
 
     def test_update_client_profile_duplicate_email_lower_mutation(self) -> None:
+        self.maxDiff = None
```
lena, line 418: remove and line 440

apps/betterangels-backend/clients/schema.py
```diff
@@ -67,60 +65,124 @@ def _format_graphql_error(error: Exception) -> str:
...
+                {
+                    "field": "hmisProfiles",
+                    "location": f"{idx}__hmisId",
+                    "errorCode": ErrorCodeEnum.HMIS_ID_NOT_PROVIDED.name,
+                }
+            )
+
+            continue
+
+        hmis_profile_id = {"id": hmis_profile["id"]} if hmis_profile.get("id") else {}
```
lena, line 185: for consistency:
```suggestion
        # exclude id of already persisted hmis profile to avoid incorrectly flagging duplicate 
        exclude_arg = {"id": hmis_profile["id"]} if hmis_profile.get("id") else {}
```
or something like that

apps/betterangels-backend/clients/schema.py
```diff
@@ -130,46 +192,68 @@ def _validate_hmis_profiles(hmis_profiles: list[dict[str, Any]]) -> list[dict[st
...
+                    "location": f"{idx}__phoneNumber",
+                    "errorCode": ErrorCodeEnum.PHONE_NUMBER_INVALID.name,
+                }
+            )
 
     return errors
 
 
-def _validate_client_profile_data(data: dict) -> None:
+def validate_client_profile_data(data: dict) -> dict[Any, Any]:
```
lena, line 227: can go back to returning None since returned data is never used

apps/betterangels-backend/clients/schema.py
mikefeldberg: agency has a default value in the fe and can't be cleared by the user

apps/betterangels-backend/clients/schema.py
```diff
@@ -130,46 +192,68 @@ def _validate_hmis_profiles(hmis_profiles: list[dict[str, Any]]) -> list[dict[st
...
     """Validates the data for creating or updating a client profile."""
-    errors = []
+    errors: list = []
+
+    user = None
+
+    if value_is_set(data.get("id")):
+        user = User.objects.filter(client_profile__id=data["id"]).first()
```
lena, line 234:
```suggestion
        user = User.objects.get(client_profile__id=data["id"])
```
and can wrap that in a try block if you want a cuter error but figure it should fail explicitly if an id that's not tied to a user is passed

apps/betterangels-backend/clients/schema.py
```diff
@@ -130,46 +192,68 @@ def _validate_hmis_profiles(hmis_profiles: list[dict[str, Any]]) -> list[dict[st
...
 
-    if data["user"] is not strawberry.UNSET:
-        user_id = data["user"].get("id", None)
-        user = User.objects.filter(id=user_id).first() if user_id else None
+    if user_data := data.get("user"):
+        errors += validate_client_name(user_data, user_data.get("nickname"), user)
```
lena, line 240: i have a slight preference for putting all of these functions into a validator class, like:
```
class ClientProfileValidator:
    def __init__(self, data):
        self.data = data
        self.errors = []
    def validate_client_profile_data(data: dict) -> None:
        if ...:
            self.validate_...

    ...
```


then all of these validate functions can just append to self.errors directly. instead of the `errors +=` pattern, but lemme know what you think.

apps/betterangels-backend/clients/schema.py
```diff
@@ -67,60 +65,124 @@ def _format_graphql_error(error: Exception) -> str:
...
+                user.last_name and user_data.get("last_name") is strawberry.UNSET,
+                user.middle_name and user_data.get("middle_name") is strawberry.UNSET,
+                user.client_profile.nickname and nickname is strawberry.UNSET,
+            )
+        )
+
+        if client_has_name:
+            return errors
+
+    errors.append({"field": "nickname", "location": None, "errorCode": ErrorCodeEnum.NAME_NOT_PROVIDED.name})
```
tglaz, line 102: thinking more about the error appended here, it'd be difficult to present it to the user correctly, since it'd just show up under the `nickname` field.

perhaps could append the error for each of the 4 fields with an specific error like:
AT_LEAST_ONE_NAME_REQUIRED
"requires one or more of these fields: first name, last name, middle name or nickname"

NAME_NOT_PROVIDED seems like specific to a single field

apps/betterangels-backend/common/constants.py
```diff
@@ -0,0 +1,3 @@
+CALIFORNIA_ID_REGEX = r"^[A-Z]\d{7}$"
+EMAIL_REGEX = r"^[a-zA-Z0-9._%+-]+@[a-zA-Z0-9.-]+\.[a-zA-Z]{2,}$"
```
tglaz, line 2: I think I may prefer `CALIFORNIA_ID_REGEX` with `a-zA-Z`, but if not then should `EMAIL_REGEX` just accept lowercase? Personally I'd expect a regex to accept both upper/lower almost always.

apps/betterangels-backend/clients/schema.py
mikefeldberg: > if `user_data.get("first_name") === " "`? Wouldn't the first line evaluate to `True`?

if I remove strip from `value_is_set`

> Is the `UNSET` check necessary here? Seems like if `value_is_set(user_data.get("first_name"))` then this code would be unreachable

this code doesn't need to be reachable in that case. This check ensures that, if the mutation proceeds, the client will have at least one name field populated. If at least one name field on the payload is a non-empty string, this requirement is satisfied.

apps/betterangels-backend/clients/schema.py
tglaz: shouldn't we validate regardless? 
don't believe that how front-end validates should not be taken into account.

apps/betterangels-backend/clients/tests/test_utils.py
```diff
@@ -0,0 +1,246 @@
...
+        self.assertEqual(value_is_set(name), expected_result)
+
+    @parametrize(
+        "first_name, middle_name, last_name, nickname, operation, should_return_error",
+        [
+            (strawberry.UNSET, strawberry.UNSET, strawberry.UNSET, strawberry.UNSET, "create", True),
+            (None, None, None, None, "create", True),
+            (" ", " ", " ", " ", "create", True),
+            ("", None, " ", strawberry.UNSET, "create", True),
+            (strawberry.UNSET, strawberry.UNSET, strawberry.UNSET, strawberry.UNSET, "update", False),
```
tglaz, line 44: perhaps add a case for `create` that returns `False`?

apps/betterangels-backend/clients/schema.py
mikefeldberg: The goal of structuring the errors like this was so they could be parsed and applied to their respective FE fields. The "name" error doesn't fit that pattern because it doesn't belong to any single field. And, by design, we display it at the top of the "full name" section, above all the fields. I can change the value of `field` in this response and have the fe parse function handle it separately

apps/betterangels-backend/clients/tests/test_utils.py
```diff
@@ -0,0 +1,246 @@
...
+        else:
+            self.assertEqual(len(errors), 0)
+
+    @parametrize(
+        "email, expected_error_code",
+        [
+            (strawberry.UNSET, None),
+            (None, None),
+            ("", None),
+            (" ", ErrorCodeEnum.EMAIL_INVALID.name),
```
tglaz, line 78: perhaps add case with a valid email

apps/betterangels-backend/clients/tests/test_utils.py
```diff
@@ -0,0 +1,246 @@
...
+        self.assertEqual(len(validate_user_email(email, user)), 0)
+
+    @parametrize(
+        "california_id, expected_error_code",
+        [
+            (strawberry.UNSET, None),
+            (None, None),
+            ("", None),
+            ("  ", ErrorCodeEnum.CA_ID_INVALID.name),
+            ("l1234567", ErrorCodeEnum.CA_ID_INVALID.name),
```
tglaz, line 107: let's ad a valid id case

apps/betterangels-backend/clients/schema.py
tglaz: What I meant was appending 4 errors, one for each of the 4 fields, which would seem a valid representation of the errors. The FE can then decide whether to collapse them into a single message at top or render per each individual field.
Not a deal breaker, but appending it to only `nickname` becomes part of tribal knowledge

apps/betterangels-backend/clients/schema.py
tglaz: so my main feedback would be that such redundancy makes it a bit tough to read and comprehend the logic, which I feel is important. 

if the `if user:` block is meant to be fully independent then it may be more clear to split it off into a separate function, and then can still call it from `validate_client_name`

tglaz (approved):

apps/betterangels-backend/clients/schema.py
mikefeldberg: I don't like the idea of splitting the two steps above into discrete functions, because that creates other logic and organization problems. What do you think of combining the steps? I can see this being more or less readable.

```python
def validate_client_name(user_data: dict, nickname: Optional[str], user: Optional[User] = None) -> list[dict[str, Any]]:
    """Verify that either:
        1. The incoming data contains at least one name field OR
        2. The existing user has at least one name field and the incoming data isn't clearing it.
    """
    errors: list = []

    if any(
        (
            value_is_set(user_data.get("first_name"))
            or (user and user.first_name and user_data.get("first_name") is strawberry.UNSET),
            value_is_set(user_data.get("last_name"))
            or (user and user.last_name and user_data.get("last_name") is strawberry.UNSET),
            value_is_set(user_data.get("middle_name"))
            or (user and user.middle_name and user_data.get("middle_name") is strawberry.UNSET),
            value_is_set(nickname) or (user and user.client_profile.nickname and nickname is strawberry.UNSET),
        )
    ):
        return errors

    errors.append({"field": "name", "location": None, "errorCode": ErrorCodeEnum.NAME_NOT_PROVIDED.name})

    return errors
```

apps/betterangels-backend/clients/schema.py
mikefeldberg: the requirement is enforced at the db level. it feels a bit overkill to add validation for something that's nominally prevented on the frontend and, if all else fails, blocked by the db

```python
class HmisProfile(models.Model):
    client_profile = models.ForeignKey("ClientProfile", on_delete=models.CASCADE, related_name="hmis_profiles")
    hmis_id = models.CharField(max_length=50)
    agency = TextChoicesField(choices_enum=HmisAgencyEnum)

    class Meta:
        constraints = [models.UniqueConstraint(fields=["hmis_id", "agency"], name="unique_hmis_id_agency")]
```

PR #1080, opened 2025-02-26
Edit Screen Redesign (DEV-1220)

Davit-BetterAngels:
DEV-1220

## Summary by Sourcery

Implements a new screen for editing client profiles, which includes routing and UI components for displaying and managing client information.

New Features:
- Introduces a new screen for editing client profiles, allowing users to modify various client details through separate forms.
- Adds new components for individual client profile forms, including ContactInfo, DemographicInfo, Fullname, HmisId, Household, ImportantNote, PersonalInfo, and RelevantContact.


apps/betterangels/src/app/(private-screens)/_layout.tsx
```diff
@@ -198,6 +198,25 @@ export default function PrivateLayout() {
           ),
         }}
       />
+      <Stack.Screen
+        name="edit-client-v2/[id]"
+        options={{
+          headerTitleAlign: 'center',
```
tglaz, line 204: i think we'd want to set `title: ''` in options.
otherwise the screen will render `edit-client-v2/[id]` as the title before it's updated. this is already visible sometimes in the app

tglaz (approved):

apps/betterangels/src/app/(private-screens)/_layout.tsx
Davit-BetterAngels: ah right

PR #1082, opened 2025-02-26
remove veteran status cruft from be (DEV-1496)

mikefeldberg:
DEV-1496 step 5 of 5
step 1: #1044
step 2: #1051
step 3: #1068
step 4: #1081

✅ NOTE: do not merge until #1081 goes out

## Summary by Sourcery

Remove the temporary veteran status fields and related logic from the backend and frontend, completing the veteran status migration.


apps/betterangels-backend/accounts/migrations/0021_clientprofile_address_clientprofile_nickname_and_more.py
```diff
@@ -68,7 +68,7 @@ class Migration(migrations.Migration):
             field=django_choices_field.fields.TextChoicesField(
                 blank=True,
                 choices=[("yes", "Yes"), ("no", "No"), ("prefer_not_to_say", "Prefer not to say")],
-                choices_enum=clients.enums.YesNoPreferNotToSayEnum,
+                choices_enum=clients.enums.VeteranStatusEnum,
```
tglaz, line 71: I'm assuming this migration already run? If so then probably shouldn't update file?

apps/betterangels-backend/clients/migrations/0001_initial.py
```diff
@@ -220,7 +220,7 @@ class Migration(migrations.Migration):
                     django_choices_field.fields.TextChoicesField(
                         blank=True,
                         choices=[("yes", "Yes"), ("no", "No"), ("prefer_not_to_say", "Prefer not to say")],
-                        choices_enum=clients.enums.YesNoPreferNotToSayEnum,
+                        choices_enum=clients.enums.VeteranStatusEnum,
```
tglaz, line 223: same as above, should we be updating migration files or create new ones to update?

tglaz (changes requested):

apps/betterangels-backend/accounts/migrations/0021_clientprofile_address_clientprofile_nickname_and_more.py
mikefeldberg: the migration doesn't need to run again, but its imports still need to be valid. So we either have to update the import or define the enum within the migration file (in which case we'd still need to remove the old import, since `YesNoPreferNotToSayEnum` no longer exists)

tglaz: so not sure when `choices` are used, but do they need to be updated?

mikefeldberg: those got updated in the most recent client migration, [here](https://github.com/BetterAngelsLA/monorepo/pull/1044/files#diff-d96e35805d72c7efa80bd8457ac4bce1e7134393bf16bc3e6b7d4177ebdfb4c7R14-R26)

tglaz (approved):

PR #1088, opened 2025-03-04
Shelter db requested changes

jluna2000:
Made the requested changes to the Shelter DB Model.


apps/betterangels-backend/shelters/enums.py
```diff
@@ -126,8 +133,13 @@ class TrainingServiceChoices(models.TextChoices):
 class EntryRequirementChoices(models.TextChoices):
     MEDICAID_OR_MEDICARE = "medicaid_or_medicare", _("Medicaid or Medicare")
     PHOTO_ID = "photo_id", _("Photo ID")
-    REFERRAL = "referral", _("Referral")
```
vecchp, line 129: Do we migrate the old data here or just drop it?

apps/betterangels-backend/shelters/enums.py
```diff
@@ -41,9 +42,11 @@ class ShelterChoices(models.TextChoices):
 # Sleeping Details
 @strawberry.enum
 class RoomStyleChoices(models.TextChoices):
-    CONGREGANT = "congregant", _("Congregant (Open)")
```
vecchp, line 44: This is going to require a data migration.

You'll likely need to do it in RAW sql.

For old migrations take the enum and put the enum class directly in the migration and update the references

vecchp: take a look at `apps/betterangels-backend/shelters/migrations/0005_spa_alter_generalservice_name_alter_population_name_and_more.py` on how to do a raw sql migration to update the existing data to get around the validation checks.

apps/betterangels-backend/shelters/deprecated/deprecated_enums.py
```diff
@@ -33,3 +33,13 @@ class SleepingChoices(models.TextChoices):
...
+
+
+class RoomStyleChoicesV14(models.TextChoices):
+    CONGREGANT = "congregant", _("Congregant (Open)")
+    CUBICLE_LOW_WALLS = "cubicle_low_walls", _("Cubicle (Low Walls)")
+    CUBICLE_HIGH_WALLS = "cubicle_high_walls", _("Cubicle (High Walls)")
+    SHARED_ROOMS = "shared_rooms", _("Shared Rooms")
+    SINGLE_ROOM = "single_room", _("Single Room")
+    MOTEL_ROOM = "motel_room", _("Motel Room")
+    OTHER = "other", _("Other")
```
vecchp, line 45: if we put this in the actual migration then we don't have to worry about removing the enum in the future.  It will automatically be removed when we squash migrations.

apps/betterangels-backend/shelters/enums.py
vecchp: We're still going to have to do a data migration here.

apps/betterangels-backend/shelters/templates/shelters/time_range_widget.html
```diff
@@ -0,0 +1,270 @@
+{% comment %} Render the standard text input {% endcomment %}
```
vecchp, line 1: Have you tested this on mobile? On desktop, I can drag and drop to select a range of times, but on mobile I have to tap each one individually. Also, I think not all shelters may have intake times scheduled on the hour or half-hour, there may be shelters which have intervals that end on the 10's or 15's i.e 10:50pm or 10:45pm.

apps/betterangels-backend/shelters/widgets.py
```diff
@@ -39,3 +42,130 @@ def compress(self, data_list: List[float]) -> Optional[str]:
...
+        raise forms.ValidationError("Invalid input type for MultiTimeRangeFormField")
+
+    def validate(self, value: Any) -> None:
+        super().validate(value)
+        for start, end in value:
+            if start >= end:
+                raise forms.ValidationError("Each time range must have a start time before the end time.")
+
+
+class TimeRangeField(models.Field):
```
vecchp, line 102: If we use a JSONField this will prevent us from using the built in Postgres' built in date/time features. I.e being able to efficiently look up times, overlapping date ranges, etc.

vecchp (changes requested):

PR #1092, opened 2025-03-05
single picker component (DEV-1540)

Davit-BetterAngels:
DEV-1540


tglaz (changes requested):

libs/expo/shared/ui-components/src/lib/SingleSelect/Picker.ios.tsx
```diff
@@ -0,0 +1,138 @@
...
+        isVisible={isModalVisible}
+        onBackdropPress={() => setIsModalVisible(false)}
+        useNativeDriverForBackdrop={true}
+      >
+        <SafeAreaView
+          style={{
+            borderTopLeftRadius: Radiuses.xs,
+            borderTopRightRadius: Radiuses.xs,
+            paddingBottom: bottomOffset,
+            backgroundColor: '#d1d3da',
```
tglaz, line 76: We seem to reuse this color, so I'd suggest either:
1. set it as a constant in the file
2. add it to our color palette in `shared/static` as something like `IOS_GRAY_100` or something. I'm assuming this and `#007aff` are colors specific to iOS UI.

libs/expo/shared/ui-components/src/lib/SingleSelect/Radio.tsx
```diff
@@ -0,0 +1,82 @@
...
+
+interface IRadioProps {
+  displayValue?: string;
+  onPress: (value: string) => void;
+  value?: string;
+  selectedItem?: string;
+  error?: string;
+}
+
+export function Radio(props: IRadioProps) {
```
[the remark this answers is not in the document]
Davit-BetterAngels: idea was to remove Radio after redesign is finalized.

libs/expo/shared/ui-components/src/lib/SingleSelect/Picker.tsx
```diff
@@ -0,0 +1,94 @@
...
+  label?: string;
+  mb?: TSpacing;
+  mt?: TSpacing;
+  my?: TSpacing;
+  mx?: TSpacing;
+  ml?: TSpacing;
+  mr?: TSpacing;
+}
+
+export default function Picker(props: IPickerProps) {
```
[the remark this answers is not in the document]
Davit-BetterAngels: same as for radio. idea is to remove picker as separate lib and use only single select everywhere.

libs/expo/shared/ui-components/src/lib/SingleSelect/Radio.tsx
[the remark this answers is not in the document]
tglaz: We should still keep Radio component. May use it in other ways.

libs/expo/shared/ui-components/src/lib/SingleSelect/Picker.tsx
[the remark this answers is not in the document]
tglaz: We should still keep Picker component. Having SingleSelect does not mean we won't need to use Picker in a separate context. Overall just good to decouple them.

libs/expo/shared/ui-components/src/lib/Picker_V2/Picker.tsx
```diff
@@ -0,0 +1,99 @@
...
+  mx?: TSpacing;
+  ml?: TSpacing;
+  mr?: TSpacing;
+}
+
+export default function Picker(props: IPickerProps) {
+  const {
+    onChange,
+    error,
+    displayValue,
```
tglaz, line 31: shouldn't this just be `value` or `selectedValue`?

libs/expo/shared/ui-components/src/lib/Picker_V2/Picker.tsx
```diff
@@ -0,0 +1,99 @@
...
+  TSpacing,
+} from '@monorepo/expo/shared/static';
+import { Picker as RNPicker } from '@react-native-picker/picker';
+import { StyleSheet, View } from 'react-native';
+import TextRegular from '../TextRegular';
+
+export interface IPickerProps {
+  onChange: (value: string) => void;
+  error?: string;
+  displayValue?: string | null;
```
tglaz, line 14: maybe ok, but could also get `displayValue` by finding it in the `items` array by the selected `value`

libs/expo/shared/ui-components/src/lib/Picker/Picker.tsx
```diff
@@ -1,32 +1,67 @@
...
+  Colors,
+  Radiuses,
+  Spacings,
+  TSpacing,
+} from '@monorepo/expo/shared/static';
 import { Picker as RNPicker } from '@react-native-picker/picker';
 import { StyleSheet, View } from 'react-native';
+import TextRegular from '../TextRegular';
 
 export interface IPickerProps {
```
tglaz, line 11: if you're adding a V2 Picker, are these changes needed? Ok if they are, but just checking

tglaz (changes requested):

libs/expo/shared/ui-components/src/lib/SingleSelect/SingleSelect.tsx
```diff
@@ -0,0 +1,96 @@
...
+  mx?: TSpacing;
+  ml?: TSpacing;
+  mr?: TSpacing;
+  label?: string;
+  placeholder?: string;
+  onChange: (value: string) => void;
+  items: { displayValue: string; value: string }[];
+  selectedValue?: string;
+  labelMarginLeft?: TSpacing;
+  boldLabel?: boolean;
```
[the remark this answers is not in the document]
Davit-BetterAngels: i removed them I remember. don't get why are they here. must be accidentally reverted wrong change.

libs/expo/shared/ui-components/src/lib/Radio_V2/Radio.tsx
```diff
@@ -0,0 +1,78 @@
...
+  Spacings,
+} from '@monorepo/expo/shared/static';
+import { Keyboard, Pressable, StyleSheet, View } from 'react-native';
+import TextRegular from '../TextRegular';
+
+interface IRadioProps {
+  displayValue?: string;
+  onPress: (value: string) => void;
+  value?: string;
+  selectedItem?: string;
```
tglaz, line 14: since we're comparing selectedItem to value, can we call it selectedValue?

libs/expo/shared/ui-components/src/lib/Radio_V2/Radio.tsx
```diff
@@ -0,0 +1,78 @@
...
+  Radiuses,
+  Spacings,
+} from '@monorepo/expo/shared/static';
+import { Keyboard, Pressable, StyleSheet, View } from 'react-native';
+import TextRegular from '../TextRegular';
+
+interface IRadioProps {
+  displayValue?: string;
+  onPress: (value: string) => void;
+  value?: string;
```
tglaz, line 13: Should either displayValue or value be required? I'd suggest we require value and leave displayValue as optional.
And use displayValue for rendering only if present, basically as a label, and otherwise render value

tglaz (approved):

PR #1098, opened 2025-03-07
[DRAFT] CPR views (DEV-1199)

tglaz:
**DRAFT PR: View Profile**
https://betterangels.atlassian.net/browse/DEV-1199

#### For Product Review purposes only

* this is not the PR that will get merged
  * I am progressively opening smaller PR's for all of the items
  * I'll ask for a final Product Review with the final PR
    * this is behind a Feature Flag so the individual PR's should be safe to merge 
* includes no `edit` button (to be implemented)

#### UI implementation:
There are only a couple main components:
* `ClientProfileCard`
  * Renders properties for individual section, such as:
    * "Full Name"
    * "Personal Info"
    * individual "Household member" or "HMIS Id profile"
    * etc...
  * Has multiple states:
    * Standard: with existing content (all fields complete)
    * Standard: with existing content (partial fields complete)
    * Empty: with no content for any of the properties
* `EmptyState`
  * Renders card content when all properties are empty
* `DataTable`
  * Renders individual properties (key-value pairs), such as:
    * First Name
    * Last Name
    * etc...
  * In case of multi-value property, such as Contact Info phone numbers, it renders multiple values with a single header
    * These have their own `EmptyState` component, as they're `one-offs`

Multi-entity views, such as "Household member" or "HMIS Id profile"
* render a `ClientProfileCard` per each entity
* has a custom `EmptyState` component

### Regarding UI feedback
* I'd expect a fair amount of UI feedback as it wasn't always clear which mocks to implement, so there are probably multiple things to adjust
* For this review it may be better to leave feedback on the JIRA ticket, with 1 comment per each item
  * This would clarify requirements on the ticket
  * ticket link: https://betterangels.atlassian.net/browse/DEV-1199
* Please scope feedback to the components listed above, as those are reused for all cards
  * If a component needs to be adjusted, please provide a specific Figma link. 
  * I worked mostly off of the Lo-Fi mocks since it wasn't always clear which Hi-Def ones to use, and hence the discrepancies
* there may be some one-off item issues, such as social media list or others, so feel free to call them out as well, but please specify that they're `one-offs`. And there most thing should fall under the standard components.


tglaz:
closing. Feature merged via other PRs

PR #1126, opened 2025-03-18
Create Client Profile (DEV-1190)

Davit-BetterAngels:
DEV-1190

## Summary by Sourcery

Implements the ability to create and edit client profiles using new UI components and form handling with React Hook Form. It introduces a new screen for creating client profiles, enhances the existing client profile editing forms, and integrates these changes with the application's routing and data fetching/updating mechanisms.

New Features:
- Adds a new screen for creating client profiles.
- Implements client profile editing forms using React Hook Form.
- Introduces new UI components such as FormCard, ControlledInput, and KeyboardAwareScrollView.
- Integrates client profile creation and editing with the application's routing.
- Adds a loading view component.
- Adds form field label and error components


apps/betterangels/src/app/(private-screens)/_layout.tsx
```diff
@@ -218,6 +218,26 @@ export default function PrivateLayout() {
           ),
```
shiozakiyuko, line 218: Hey @Davit-BetterAngels 

It's almost perfect! 😄 The only thing we need to change is the color of * (asterisk).

shiozakiyuko (approved):

libs/expo/betterangels/src/lib/screens/CreateClientProfile/types.ts
```diff
@@ -0,0 +1,5 @@
+export type TValidationError = {
```
tglaz, line 1: we seem to use it in 3 places now. If this is a standard error, let's just keep one of them

libs/expo/betterangels/src/lib/screens/CreateClientProfile/index.tsx
```diff
@@ -0,0 +1,165 @@
...
+
+      if (result?.__typename === 'ClientProfileType') {
+        router.replace(`/client/${result.id}`);
+      } else {
+        console.log('Unexpected result: ', result);
+        showSnackbar({
+          message: `Something went wrong!`,
+          type: 'error',
+        });
+        router.replace(`/clients`);
```
tglaz, line 94: Probably can simplify the if/else here.
Below is an alternative version, though I'm not sure how possible it is to get `result?.__typename !== 'ClientProfileType'`.

And do we want to redirect the user back to `/clients` or just show snackbar message and ask to try again?
Not totally sure, but if need to redirect then the error block wouldn't throw but showSnackbar and reroute.

```javascript
     // something went wrong
     // would just checking "if result exists" do same thing?
      if (result?.__typename !== 'ClientProfileType') {
           throw('Unexpected CreateClientProfile  result: ', result)
      } 

       // success
       router.replace(`/client/${result.id}`);
      }
```

libs/expo/betterangels/src/lib/screens/CreateClientProfile/index.tsx
```diff
@@ -0,0 +1,165 @@
...
+  | 'user.lastName'
+  | 'nickname';
+
+interface FormField {
+  label: string;
+  name: AllowedFieldNames;
+  placeholder: string;
+}
+
+const FORM_FIELDS: FormField[] = [
```
tglaz, line 30: Technically this can be different from the `Fullname` form, so probably ok.
But otherwise could import from there, or just reuse that form and just handle the submit here.

tglaz (changes requested):

kimchoi-jjiggae (approved):

libs/expo/betterangels/src/lib/screens/CreateClientProfile/index.tsx
Davit-BetterAngels: typescript will not recognize result.id if we are not checking 

result?.__typename === 'ClientProfileType'

tglaz (approved):

PR #1132, opened 2025-03-20
Redis Connection Pool Settings (DEV-1553)

vecchp:
Setting a default of 100 connections per instance as a starting point.  Will be configurable in production / dev.

## Summary by Sourcery

Enhancements:
- Configures Redis connection pool settings, including the connection pool class, maximum number of connections, and blocking timeout.


apps/betterangels-backend/betterangels_backend/settings.py
```diff
@@ -45,6 +45,8 @@
     DEBUG=(bool, False),
     DEFAULT_FROM_EMAIL=(str, ""),
     DJANGO_CACHE_URL=(str, ""),
+    DJANGO_CACHE_BLOCKING_TIMEOUT=(float, 1.0),
```
tglaz, line 48: probably not a big deal, but would 1 sec be a bit aggressive?

tglaz (approved):

PR #1134, opened 2025-03-20
document deletion confirmation modal

aaronipollock:
Added a confirmation modal when deleting documents.

1. Changed Delete Flow:
- Instead of deleting immediately on button press, shows a confirmation modal first
- Original delete button now triggers the confirmation modal
- Final deletion happens only after user confirms in the second modal

2. UX Improvements:
- Added explanatory text: "All data associated with this [file/image] will be deleted"
- Provides Cancel option to back out of deletion
- Maintains consistent UI pattern with other modal confirmations

3. Technical Implementation:
- I chose to use the BasicModal over the DeleteModal in order to keep the code more consistent with how other actions work and keep all the visibility logic in one place, which I thought worked better with the existing ACTIONS structure without requiring extra modifications the way DeleteModal would seem to.

## Summary by Sourcery

Enhancements:
- Improves UX by adding a confirmation modal with explanatory text and a cancel option before deleting documents.


tglaz:
The issue here is that we can't really have a custom modal for each confirmation modal, except in super special cases. Otherwise the code becomes unmaintainable very quickly.

Is there a good reason why the `DeleteModal` wouldn't work here? I'm not 100% familiar with this and so perhaps there is. However, if it's a discrepancy with the mocks provided, you could perhaps comment on the ticket and ask if the default design DeleteModal design is acceptable. If there is a functional issue, feel free to let me know, as perhaps it's possible to adjust the existing solution perhaps.

tglaz (changes requested):
The issue here is that we can't really have a custom modal for each confirmation modal, except in super special cases. Otherwise the code becomes unmaintainable very quickly.

Is there a good reason why the DeleteModal wouldn't work here? I'm not 100% familiar with this and so perhaps there is. However, if it's a discrepancy with the mocks provided, you could perhaps comment on the ticket and ask if the default design DeleteModal design is acceptable. If there is a functional issue, feel free to let me know, as perhaps it's possible to adjust the existing solution perhaps.

aaronipollock:
Thanks, @tglaz! The issue I found with the DeleteModal was integrating it with the ACTIONS array structure:

  ```
const ACTIONS = [
    {
      title: `View this ${fileTypeText}`,
      Icon: ViewIcon,
      route: `/file/${document?.id}`,
    },
    {
      title: `Download this ${fileTypeText}`,
      Icon: DownloadIcon,
      onPress: downloadFile,
    },
    {
      title: `Delete this ${fileTypeText}`,
      Icon: DeleteIcon,
      onPress: handleDelete,
    },
  ];
```

DeleteModal requires a button to wrap with its own click handler, while the ACTIONS array expects simple function handlers. I did work on a custom action renderer that would accommodate DeleteModal, but there also seems to be a visibility handling conflict between the ACTIONS array and DeleteModal component. 

The changes required seemed to be outside the scope of the ticket, so I went with the BasicModal. Should I have commented on the ticket instead? Suggestions appreciated!

tglaz:
@aaronipollock Ok, I see. Perhaps we can have a further chat about this Monday, or whenever you're available. Think the actual issue may be related to rendering a nested modal, which you solved with a custom modal to handle that. 

This works, but I'd prefer to avoid it if possible, and figure out a reusable way to handle this scenario (as it'll happen again). We may include @jluna2000 in the conversation since he's been doing research on different modal imlementations. There are options such as using the [gorhom/react-native-bottom-sheet](https://github.com/gorhom/react-native-bottom-sheet) lib or [modal-screen-using-expo-router](https://docs.expo.dev/router/advanced/modals/#modal-screen-using-expo-router) that would help (though TBD).

aaronipollock:
@tglaz These are the changes to use the DeleteModal

tglaz (approved):

PR #1138, opened 2025-03-24
CPR Personal Info form (DEV-1187)

tglaz:
### Edit Personal Info
https://betterangels.atlassian.net/browse/DEV-1187

* adds `livingSituation` field
* adds `veteranStatus` field
* adds `preferredLanguage` field
* adds `californiaId` field
    * uses `ffClientDedupe` feature flag for the in-progress `dedupe modal`
* updates `FormStateMapping` to use `ClientProfileCardEnum` as keys
  * allows for consistently passing the keys as openCard param etc...
* after form submit opens the card corresponding to updated form

## Summary by Sourcery

Adds fields for living situation, veteran status, preferred language and California ID to the personal info form. Implements a feature flag for a dedupe modal when a duplicate California ID is detected.

New Features:
- Adds fields for living situation, veteran status, and preferred language to the personal info form.
- Adds a California ID field to the personal info form with validation and a dedupe modal when a duplicate is detected, controlled by a feature flag.


libs/expo/betterangels/src/lib/screens/ClientProfileForms-V2/index.tsx
```diff
@@ -140,7 +142,8 @@ export default function ClientProfileForms(props: IClientProfileForms) {
       }
 
       refetch();
-      router.replace(`/client/${id}`);
+
+      router.replace(`/client/${id}?openCard=${validComponentName}`);
```
tglaz, line 146: Technically doing `router.back()` would do the same thing. But this may also need to change when we add the HmisId multi-step form (we'll see).

shiozakiyuko (commented):
Hi @tglaz 

I noticed one thing.

1. **Discrepancy between iOS and Android**
I didn't see the default label for dropdown menu on Android. 
(iOS shows the default text e.g., Select veteran status, Select language, etc)
Starting with Housing for Living Situation, Yes for Veteran Status, and Arabic for Language can be very confusing to someone who doesn't know about the unhoused clients. Please modify this dropdown placeholder/default text.

Davit-BetterAngels (approved):

PR #1146, opened 2025-03-26
HmisProfile endpoints (DEV-1602)

mikefeldberg:
DEV-1602

- Adds `HmisProfile` endpoints, permissions, admin
- Fixes a unrelated few typos, missing tests, etc.
- Adds `NonBlankString` scalar to strip white space on string inputs. If the output is an empty string, this will cause a db error that will be returned on the mutation payload. graphql inspector flags it as a breaking change, since the type and input are changing from `str` to `NonBlankString`, but I don't think it will break anything since the scalar takes a string. 
  - If anything breaks, the quickest fix would be to change `hmis_id` back to a `str` and remove the validation test I added. Alternatively, I could cut new PR with just the new scalar and validation test, and deploy it after this one.

## Summary by Sourcery

Add HMIS Profile endpoints and permissions to the client management system

New Features:
- Implement GraphQL endpoints for HMIS Profile operations including create, read, update, and delete
- Add permission checks for HMIS Profile interactions

Enhancements:
- Refactor test utilities to support HMIS Profile testing
- Improve permission handling for client-related operations

Tests:
- Add comprehensive test cases for HMIS Profile permissions
- Create test utilities for HMIS Profile mutations and queries

Chores:
- Update GraphQL schema to include HMIS Profile type and operations
- Create migration to add HMIS Profile permissions


apps/betterangels-backend/clients/schema.py
```diff
@@ -355,50 +366,10 @@ def bulk_client_profile_import_records(
...
-                assign_perm(perm, permission_group.group, client_document)
-
-            return cast(ClientDocumentType, client_document)
-
-    delete_client_document: ClientDocumentType = mutations.delete(
-        DeleteDjangoObjectInput,
-        extensions=[
-            HasRetvalPerm(perms=AttachmentPermissions.DELETE),
-        ],
-    )
```
mikefeldberg, line 393: just moving this lower in the file

apps/betterangels-backend/clients/schema.py
```diff
@@ -355,50 +366,10 @@ def bulk_client_profile_import_records(
...
-
     @strawberry_django.mutation(extensions=[HasPerm(perms=[ClientProfilePermissions.ADD])])
     def create_client_profile(self, info: Info, data: CreateClientProfileInput) -> ClientProfileType:
         with transaction.atomic():
-            user = get_current_user(info)
-            get_user_permission_group(user)
```
mikefeldberg, line 399: this isn't part of the ticket but this shouldn't be here. probably copypasta

apps/betterangels-backend/clients/schema.py
```diff
@@ -487,25 +449,6 @@ def update_client_profile(self, info: Info, data: UpdateClientProfileInput) -> C
...
-                    [ClientProfilePermissions.CHANGE],
-                ).get(id=data.client_profile)
-
-                client_profile.profile_photo = data.photo
-                client_profile.save()
-
-            except ClientProfile.DoesNotExist:
-                raise PermissionError("You do not have permission to modify this client.")
-
-            return cast(ClientProfileType, client_profile)
```
mikefeldberg, line 507: just moving this lower in the file

apps/betterangels-backend/clients/schema.py
```diff
@@ -528,6 +471,77 @@ def delete_client_profile(self, info: Info, data: DeleteDjangoObjectInput) -> De
...
+                    [ClientProfilePermissions.CHANGE],
+                ).get(id=data.client_profile)
+
+                client_profile.profile_photo = data.photo
+                client_profile.save()
+
+            except ClientProfile.DoesNotExist:
+                raise PermissionError("You do not have permission to modify this client.")
+
+            return cast(ClientProfileType, client_profile)
```
mikefeldberg, line 543: relocated from above. no change

apps/betterangels-backend/clients/tests/test_permissions.py
```diff
@@ -48,6 +61,38 @@ def test_create_client_profile_permission(self, user_label: str, should_succeed:
...
+        "user_label, should_succeed",
+        [
+            ("org_1_case_manager_1", True),
+            ("org_1_case_manager_2", True),
+            ("org_2_case_manager_1", True),
+            ("client_user_1", False),  # Non CM should not succeed
+            (None, False),  # Anonymous user should not succeed
+        ],
+    )
+    def test_update_client_profile_permission(self, user_label: str, should_succeed: bool) -> None:
```
mikefeldberg, line 74: also not part of the ticket but I noticed we didn't have this test

apps/betterangels-backend/common/graphql/types.py
```diff
@@ -14,6 +14,12 @@
     parse_value=lambda v: parse(v, "US"),
 )
 
+NonBlankString = strawberry.scalar(
+    NewType("NonBlankString", str),
+    serialize=lambda v: str(v),
+    parse_value=lambda v: v.strip(),
```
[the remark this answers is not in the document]
mikefeldberg: an empty string will return a validation error because the db field is `blank=False`

tglaz (approved):
lgtm. 
not super familiar with the perm system, but seemed ok and tested

apps/betterangels-backend/clients/schema.py
```diff
@@ -429,18 +400,9 @@ def create_client_profile(self, info: Info, data: CreateClientProfileInput) -> C
...
-            user = get_current_user(info)
-            try:
-                client_profile = filter_for_user(
-                    ClientProfile.objects.all(),
-                    user,
-                    [ClientProfilePermissions.CHANGE],
-                ).get(id=data.id)
-                client_user = client_profile.user
-            except ClientProfile.DoesNotExist:
-                raise PermissionError("You do not have permission to modify this client.")
```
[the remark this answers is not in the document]
vecchp: Quick question — when exactly do the permission checks happen here?
Since we’re directly fetching the ClientProfile with .get(), are we certain the checks run before any updates are applied?

Also, is it guaranteed that this mutation always runs inside a transaction?
Just want to make sure we’re not at risk of applying updates and then failing the permission check afterward.

From what I can tell, as long as this line is executed under transaction.atomic(), we should be safe:
https://github.com/strawberry-graphql/strawberry-django/blob/6fe362b66c8ad44b3dbfc60b87b5e082ff26fe4a/strawberry_django/mutations/fields.py#L333

apps/betterangels-backend/clients/tests/test_permissions.py
```diff
@@ -48,6 +61,38 @@ def test_create_client_profile_permission(self, user_label: str, should_succeed:
...
+        if should_succeed:
+            self.assertIsNotNone(response["data"]["updateClientProfile"]["id"])
+            self.assertNotEqual(original_nickname, updated_nickname)
+        else:
+            self.assertEqual(len(response["errors"]), 1)
+            self.assertIn(
+                response["errors"][0]["message"],
+                {
+                    "You do not have permission to modify this client.",
+                    "Cannot filter against a non-conditional expression.",
```
mikefeldberg, line 91: this is the `TypeError` raised when running `filter_for_user` on a django guardian anonymoususer

PR #1149, opened 2025-03-27
reset notes and offset on interaction filterchange (DEV-1588)

tglaz:
### Interaction Search returns results that do not match search terms
https://betterangels.atlassian.net/browse/DEV-1588

* issue seemed to be related to not resetting Notes and Offset data in state, so items from different queries could just get added to existing list.
  * hopefully this resolves the issues, although there is perhaps a chance for discrepancies multiple queries are sent and return in different order. We debounce requests by 500ms which should mitigate that.
* also added `dedupe` by Interaction ID before merging results

Note: 
There is a lot of opportunity for refactoring here, such as:
* move debounce into Search field component to minimize state changes and renders
* consolidate duplicate code between the Interaction screen and Clients Interactions tab
* rename Notes to Interactions

## Summary by Sourcery

Fix interaction search results by resetting notes and offset when filter changes

Bug Fixes:
- Prevent duplicate notes from being added to search results by resetting notes and offset when filter search changes
- Ensure unique notes are displayed by using a uniqueBy function when appending new search results

Enhancements:
- Added useEffect to reset notes and offset state when filter search changes
- Implemented unique note filtering to prevent duplicate entries


Davit-BetterAngels (approved):

shiozakiyuko (approved):
Hi @tglaz approved!

kimchoi-jjiggae (approved):

PR #1158, opened 2025-03-29
dual write client name and email fields (DEV-1609)

mikefeldberg:
DEV-1609

**This is step 2 of (part 1) of 7<sup>*</sup> toward removing Users from ClientProfiles (see epic: DEV-1607)**

<sup>*</sup>*One of the steps is an epic* 😬. *We need to update `Note.client` to point to `ClientProfile` rather than `User`. See: DEV-1615*

_**NOTE: do not merge before:**_

Step 1: #1157

## Summary by Sourcery

Migrate client name and email fields from User to ClientProfile, preparing for decoupling Users from ClientProfiles

New Features:
- Add first_name, last_name, middle_name, and email fields directly to ClientProfile model

Enhancements:
- Implement dual write mechanism to populate new ClientProfile fields from User data
- Add full_name property to ClientProfile
- Update search functionality to use new name fields

Tests:
- Add tests for new name and email fields
- Add test for full_name property
- Add test for dual write functionality

Chores:
- Create database migration to add new fields
- Update GraphQL schema to include new fields

## Summary by Sourcery

Migrate client name and email fields from User to ClientProfile, preparing for decoupling Users from ClientProfiles

New Features:
- Add first_name, last_name, middle_name, and email fields directly to ClientProfile model

Enhancements:
- Implement dual write mechanism to populate new ClientProfile fields from User data
- Add full_name property to ClientProfile
- Update search functionality to use new name fields
- Make user field optional in ClientProfile

Tests:
- Add tests for new name and email fields
- Add test for full_name property
- Add test for dual write functionality

Chores:
- Create database migration to add new fields
- Update GraphQL schema to include new fields

PR #1159, opened 2025-03-29
fe cutover to clientprofile name and email fields (DEV-1610)

mikefeldberg:
DEV-1610

**This is step 3 of 7<sup>*</sup> toward removing Users from ClientProfiles (see epic: DEV-1607)**
<sup>*</sup>*There are 7 tickets in this epic but in order to complete them, we also have to update `Note.client` to point to `ClientProfile` rather than `User`, which is a separate epic (DEV-1615)*

_**NOTE: do not merge before:**_

Step 1: #1157
Step 2: #1158, #1160

I also threw in a few unrelated type fixes that weren't caught by CI

---

**_we should thoroughly test all name and email fields to make sure everything works as expected_**


mikefeldberg:
released in #1183

PR #1184, opened 2025-04-03
shelter back button fix (DEV-1631)

Davit-BetterAngels:
DEV-1631

## Summary by Sourcery

Implement map center persistence using session storage to maintain map view state when navigating back to the map page

Bug Fixes:
- Fix map view state loss when navigating between pages by saving and restoring map center coordinates

Enhancements:
- Add ability to preserve map center location using browser session storage


apps/shelter-web/src/app/shared/components/map/map.tsx
```diff
@@ -59,9 +59,40 @@ export function Map(props: TMap) {
     console.info(`[map] loading status: ${mapApiStatus}`);
   }, [mapApiStatus]);
 
+  useEffect(() => {
+    const savedCenter = sessionStorage.getItem('mapCenter');
```
tglaz, line 63: I think this should be passed in as a prop (defaultCenter?). 
A shared Map probably shouldn't know anything about sessions.

apps/shelter-web/src/app/shared/components/map/map.tsx
```diff
@@ -59,9 +59,40 @@ export function Map(props: TMap) {
...
+    }
+  }, []);
+
   const handleCameraChange = useCallback(
     (event: MapCameraChangedEvent) => {
       setCameraProps(event.detail);
+      const { center } = event.detail;
+
+      if (center) {
+        sessionStorage.setItem(
```
tglaz, line 88: We already use atoms in shelter-web, so may be worth sticking to a single state management system.
If we want to have the data persist, can use [atomwithstorage](https://jotai.org/docs/utilities/storage), or implement custom

also, I think `handleCameraChange` fires a ton of events, kind of like onScroll, so should be careful

tglaz (changes requested):

kimchoi-jjiggae:
Approved for this iteration, although need to address these edge cases later: [DEV-1658: Edge cases for red pin behavior
Draft](https://betterangels.atlassian.net/browse/DEV-1658)

kimchoi-jjiggae (approved):
See comment, works for now but needs future work

apps/shelter-web/src/app/shared/components/map/map.tsx
Davit-BetterAngels: I want specifically store it in session storage so to use atom to create custom function to store it in this case makes no sense. that is why I preferred sessionStorage. 

I think its safe to track the pin in this way. Maybe I'm missing something we can discuss.

apps/shelter-web/src/app/pages/home/home.tsx
```diff
@@ -60,10 +62,25 @@ export function Home() {
     });
   }
 
+  useEffect(() => {
+    const savedCenter = sessionStorage.getItem('mapCenter');
+
+    if (savedCenter) {
```
tglaz, line 68: could get rid of if/else if do `center = savedCenter || LA_COUNTY_CENTER`?

tglaz (approved):
left a small comment, but aprvd

apps/shelter-web/src/app/shared/components/map/map.tsx
tglaz: i think it's safe, and not end of the world, but more a question of having multiple state storage systems within the app. Ideally we just use one, but won't block on it

PR #1198, opened 2025-04-07
don't set associated_with on doc mutation (DEV-1681)

mikefeldberg:
DEV-1681

## Summary by Sourcery

Remove unnecessary association of document with user during client document creation

Bug Fixes:
- Prevent unnecessary setting of 'associated_with' field when creating a client document

Tests:
- Adjusted test case to reflect reduced query count after removing unnecessary user association


mikefeldberg:
this will get covered by the be cleanup work

PR #1199, opened 2025-04-07
Added a View Details button to the pop up cards for shelters on maps

jluna2000:
Added a View Details button to the map pop up card that leads to the same route as the ShelterCard. [DEV-1622](https://betterangels.atlassian.net/browse/DEV-1622)

## Summary by Sourcery

Add a "View Details" button to shelter map pop-up cards to enable navigation to individual shelter pages

New Features:
- Added a clickable 'View Details' button in shelter map pop-up cards that navigates users to the specific shelter's detailed page

Enhancements:
- Improved user navigation by providing a direct link from map markers to shelter details

[DEV-1622]: https://betterangels.atlassian.net/browse/DEV-1622?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


vecchp (changes requested):
image
I think the CSS is a bit off compared to the mock

kimchoi-jjiggae (approved):

shiozakiyuko (commented):
Hi @jluna2000 

Thanks for working on this.

It looks good but I don't see the pressed state on the View Details button.
Can you add Hex 1650C1 for the pressed state?

apps/shelter-web/src/app/pages/home/home.tsx
```diff
@@ -43,19 +44,25 @@ export function Home() {
...
-          className="mt-4"
-          shelter={
-            shelters.find((shelter) => shelter.id === markerId) as TShelter
-          }
-        />
+        <>
+          <ShelterCard
+            className="mt-4"
+            shelter={
+              shelters.find((shelter) => shelter.id === markerId) as TShelter
```
tglaz, line 59: please remove the `as TShelter` type casting as it's dangerous.

`shelters` is already of type `TShelter[]`, and the reason there is a typing issue is that `shelters.find(shelter) => ` can return `undefined`, in which case everything breaks here. So we need to handle the case where it's not found, which you can do at the top of the function where we check for `!markerId`

also, let's rename `markerId` to `shelterId`, as that's what it is

apps/shelter-web/src/app/pages/home/home.tsx
```diff
@@ -43,19 +44,25 @@ export function Home() {
...
+            }
+          />
+          <div onClick={() => navigate(`/shelter/${markerId}`)} className="font-semibold text-sm text-center cursor-pointer pt-6 md:pt-10 text-[#216BF8] active:text-[#1650C1]">
```
tglaz, line 62: couple things here:
1. We shouldn't be using hex codes (#216BF8) for color within the app. All colors should be defined within the shelter-web `global.css` and  `tailwind.config.js` 
2. If <ShelterCard> already has an `onClick` event then we don't want 2 different click handlers here. Seems like we probably just want to optionally render `View Details` button inside the ShelterCard. There are few ways to do this:
* add a `footer` prop to ShelterCard that accepts a ReactNode and renders it at bottom if present
* perhaps add a `footerButtonText` prop to ShelterCard and render the div with text at bottom. The card itself already has an `onClick` event, so it's basically one big button already, so if we render `View Details` within the cards it can just be a div

tglaz (changes requested):

apps/shelter-web/src/app/shared/modal/modal.tsx
```diff
@@ -70,7 +70,6 @@ export function Modal(props: IModal): ReactElement | null {
     'max-h-[calc(100vh - 88)]',
     'overflow-hidden',
     'overflow-y-auto',
-    type === 'default' ? 'pb-12' : '',
```
tglaz, line 73: This may be ok, but just wanted to double check as this is a default style for the `modal` and not specific to this feature. If this change makes sense globally then ok, but otherwise we'd need to pass in the custom styles as an option somehow.

apps/shelter-web/src/app/shared/components/shelter/shelterCard.tsx
```diff
@@ -42,6 +45,7 @@ export function ShelterCard(props: TShelterCard) {
     'md:flex-row',
     'cursor-pointer',
     className,
+    footer? 'pb-6':'',
```
tglaz, line 48: let's remove this and define footer styles either:
1. The `footer` that's passed in could just have its own margin styles
2. Create a `ShelterCardFooter` component (basically a wrapper). If go this route, then:
* define it in a separate file
* it should accept `children` to render inside
* it should have its own default className  (probably just default padding)
* `TShelterCard` should also accept `footerStyle` prop, which should be passed to `ShelterCardFooter` and merged with the default styling (idea here is to allow someone to override the default css)
* render it if `footer` is passed in
  * `footer` should be rendered inside of ShelterCardFooter

so in the component you may have:

```javascript
{footer & <ShelterCardFooter className={footerStyles}>{footer}</ShelterCardFooter>
```

and inside `ShelterCardFooter` it could look something like:
```javascript
const parentCss = ['default css classes', className];

  return (
    <div className={mergeCss(parentCss)}>
    ...
```

tglaz, line 48: let's remove this and define footer styles either:
1. The `footer` that's passed in could just have its own margin styles
2. Create a `ShelterCardFooter` component (basically a wrapper). If go this route, then:
* define it in a separate file
* it should accept `children` to render inside
* it should have its own default className  (probably just default padding)
* `TShelterCard` should also accept `footerStyle` prop, which should be passed to `ShelterCardFooter` and merged with the default styling (idea here is to allow someone to override the default css)
* render it if `footer` is passed in
  * `footer` should be rendered inside of ShelterCardFooter

so in the component you may have:

```javascript
{footer & <ShelterCardFooter className={footerStyles}>{footer}</ShelterCardFooter>
```

and inside `ShelterCardFooter` it could look something like:
```javascript
const parentCss = ['default css classes', className];

  return (
    <div className={mergeCss(parentCss)}>
    ...
```

shiozakiyuko (commented):
Hi @jluna2000 

Sorry for taking so long to review this.

The View Details button is very close to the location line above. 
Can you add or make sure the button has 0.5rem vertical padding and 1 rem spacing between the location and the button?



Thanks!

shiozakiyuko (commented):
Hi @jluna2000 

Here is the screen shot for the desktop for now.

shiozakiyuko (approved):

apps/shelter-web/src/app/pages/home/home.tsx
```diff
@@ -44,6 +45,15 @@ export function Home() {
     setShelterMarkers(markers);
   }, [shelters]);
 
+  const footer_style = [
```
tglaz, line 48: Pls change it to `footerStyle`.
By convention javascript uses `camelCase` for variables.

apps/shelter-web/src/app/shared/components/shelter/shelterCard.tsx
```diff
@@ -23,13 +24,17 @@ type TShelterCard = {
...
+  footerName?: string;
 };
 
 export function ShelterCard(props: TShelterCard) {
   const {
     shelter: { id, name, heroImage, distanceInMiles, location },
     originCoordinates,
     className,
+    footer,
+    footerName,
```
tglaz, line 37: pls change this to `footerClassName`

apps/shelter-web/src/app/shared/components/shelter/shelterCard.tsx
```diff
@@ -39,11 +44,25 @@ export function ShelterCard(props: TShelterCard) {
...
+  const bodyCss = [
+    'flex',
+    'flex-col',
+    'md:flex-row',
+    'cursor-pointer',
+  ];
+
+  const footerCss = [
+    'mt-4',
+    'py-2',
```
tglaz, line 60: We probably don't want both padding and margin here.
To separate the footer vertically from top can either use `mt` or `pt`, and just adjust that property for `md` responsive style.
I'd suggest just using `mt`.

`py` sets both padding top and bottom, and I'd guess we only need vertical separation above, so `mt` would be enough. I don't think we'd need extra spacing below the footer, as the card padding should take care of that, but in this case don't care.

tglaz (changes requested):

PR #1203, opened 2025-04-07
bump version for release

mikefeldberg:
this release will include a few big changes (tho none of them client facing):
1. cutover to paginated `clientProfiles` and `notes` queries
3. removal of `user` from `ClientProfile` (first, last, middle name and email fields now live directly on `ClientProfile`)
4. replacing `note.client` (fk `User`) with `note.client_profile` (fk `ClientProfile`)

these updates should be thoroughly tested before publishing to app stores

## Summary by Sourcery

Bump application version for release, including version increments for mobile platforms

New Features:
- Prepare for release with version increments across iOS, Android, and app configuration

Chores:
- Increment app version from 1.0.39 to 1.0.40
- Update iOS build number from 1.0.50 to 1.0.51
- Increment Android version code from 49 to 50

PR #1210, opened 2025-04-08
fix interaction filters & add tests (DEV-1693)

mikefeldberg:
DEV-1693

old bug. caught while testing fe for release, [here](https://github.com/BetterAngelsLA/monorepo/pull/1203#discussion_r2033912346)

## Summary by Sourcery

Fix interaction filters for notes query to handle empty filter lists correctly

Bug Fixes:
- Modify filter logic to correctly handle empty lists for authors and teams filters in notes query, ensuring all notes are returned when no specific filter is applied

Tests:
- Add comprehensive test cases for notes query filters, including scenarios with empty author and team filter lists

### DIFF
--- a/apps/betterangels-backend/notes/types.py
+++ b/apps/betterangels-backend/notes/types.py
@@ -158,7 +158,7 @@ class NoteFilter:
     def authors(
         self, queryset: QuerySet, info: Info, value: Optional[List[ID]], prefix: str
     ) -> Tuple[QuerySet[models.Note], Q]:
-        if value in [None, []]:
+        if not value:
             return queryset, Q()
 
         return queryset.filter(created_by__in=value), Q()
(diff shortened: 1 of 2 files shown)


apps/betterangels-backend/notes/types.py
```diff
@@ -158,7 +158,7 @@ class NoteFilter:
     def authors(
         self, queryset: QuerySet, info: Info, value: Optional[List[ID]], prefix: str
     ) -> Tuple[QuerySet[models.Note], Q]:
-        if value is None:
+        if value in [None, []]:
```
vecchp, line 161:
```python
t = None

if not t:
    print("t is None")

x = []

if not x:
  print('x is empty')
```
```
t is None
x is empty
```

I think we can simply do if not value here.

apps/betterangels-backend/notes/types.py
```diff
@@ -191,7 +191,7 @@ def search(
     def teams(
         self, queryset: QuerySet, value: Optional[List[SelahTeamEnum]], prefix: str
     ) -> Tuple[QuerySet[models.Note], Q]:
-        if value is None:
+        if value in [None, []]:
```
vecchp, line 194: Same as above for here.

vecchp (changes requested):

vecchp (approved):

PR #1218, opened 2025-04-10
Mutation to rename client document

CruiseDevice:
DEV-1674

## Summary by Sourcery

Add mutation to rename client documents with permission-based access control

New Features:
- Introduce a new mutation to update the filename of an existing client document

Enhancements:
- Create a permission-based update mechanism for client documents using Strawberry GraphQL


apps/betterangels-backend/clients/tests/test_mutations.py
```diff
@@ -643,6 +646,12 @@ class ClientDocumentMutationTestCase(ClientProfileGraphQLBaseTestCase):
     def setUp(self) -> None:
         super().setUp()
         self._handle_user_login("org_1_case_manager_1")
+        self._setup_client_documents()
+
+        # Ensure permissions are set correctly
+        permission_group = get_user_permission_group(self.org_1_case_manager_1)
+        document = Attachment.objects.get(id=self.client_profile_1_document_1["id"])
+        assign_perm(AttachmentPermissions.CHANGE, permission_group.group, document)
```
mikefeldberg, line 654: This would hide potential errors. The updated `create_client_document` should grant `AttachmentPermissions.CHANGE` permission when a document is created. We want to verify this functionality with a test where: (1) a user logs in; (2) creates a document; (3) attempts to update the document. That test would fail if the create mutation isn't assigning the permission correctly. In this test, on line 654, you're adding an extra step where you manually assign the `CHANGE` permission. This means means this test will always pass, even if the `create_client_document` mutation isn't working as expected.

apps/betterangels-backend/clients/tests/test_mutations.py
```diff
@@ -681,6 +690,77 @@ def test_delete_client_document(self) -> None:
             "The document should have been deleted from the database.",
         )
 
+    def test_update_client_document_success(self) -> None:
```
mikefeldberg, line 693: you can call this 
```suggestion
    def test_update_client_document(self) -> None:
```
 by default, we assume that a test is verifying the "happy" path

apps/betterangels-backend/clients/tests/test_mutations.py
```diff
@@ -681,6 +690,77 @@ def test_delete_client_document(self) -> None:
...
+
+        self.assertIsNotNone(response.get("data"))
+        self.assertIsNotNone(response["data"].get("updateClientDocument"))
+        self.assertEqual(response["data"]["updateClientDocument"]["originalFilename"], new_filename)
+
+        # Verify database was updated
+        document = Attachment.objects.get(id=document_id)
+        self.assertEqual(document.original_filename, new_filename)
+
+    def test_update_client_document_empty_filename(self) -> None:
```
mikefeldberg, line 728: I don't think we need this test. Technically, I don't think there's anything in the requirements stating that a filename can't be empty

apps/betterangels-backend/clients/types.py
```diff
@@ -371,3 +371,10 @@ class ImportClientProfileInput:
     source_name: auto
     raw_data: auto
     client_profile: CreateClientProfileInput
+
+
+# Input for updating a client document
+@strawberry_django.input(Attachment, fields=["id", "original_filename"])
+class UpdateClientDocumentInput:
+    id: strawberry.ID
```
mikefeldberg, line 379: `ID` is imported at the top of the file, so this can just be
```suggestion
    id: ID
```

apps/betterangels-backend/clients/types.py
```diff
@@ -371,3 +371,10 @@ class ImportClientProfileInput:
...
+    original_filename: str
```
mikefeldberg, line 380: I think you can make this `auto` instead of `str`

apps/betterangels-backend/clients/tests/test_mutations.py
```diff
@@ -681,6 +690,77 @@ def test_delete_client_document(self) -> None:
...
+                }
+            """,
+            variables={
+                "data": {
+                    "id": document_id,
+                    "originalFilename": "",
+                }
+            },
+        )
+        print("\n Full GraphQL Response: ", response)
```
mikefeldberg, line 751: don't need this since you have assertions below. We can also meet to go over some debugging tools you could use so you don't need print statements

mikefeldberg (changes requested):
thanks! I left some comments below. I also noticed that there's no permission test ([AC4](https://betterangels.atlassian.net/browse/DEV-1674)) and the types haven't been updated (`ynx-generate`). You can take a look at [my branch](https://github.com/BetterAngelsLA/monorepo/compare/main...DEV-1674/update-document-filename) to see the permission test I added. I also included an `_update_client_document_fixture` in the utils file. 

Let me know if you have any questions or want to review anything together!

apps/betterangels-backend/clients/tests/test_permissions.py
```diff
@@ -314,6 +320,39 @@ def test_view_client_documents_permission(self, user_label: str, expected_docume
         else:
             self.assertTrue("errors" in response)
 
+    @parametrize(
+        "user_label, should_succeed",
+        [
+            ("org_1_case_manager_1", True),  # Owner should succeed
+            ("org_1_case_manager_2", True),  # Other CM in owner's org should succeed
+            ("org_2_case_manager_1", False),  # CM in different org should not succeed
+            ("client_user_1", False),  # Client modifying client profile should not succeed
```
mikefeldberg, line 329: you can change this to
```suggestion
            ("non_case_manager_user", False),  # Client modifying client profile should not succeed
```
and undo the updates to `apps/betterangels-backend/common/tests/utils.py` below. We used to use `client_user_1` in these tests but we removed client users. Now we use `non_case_manager_user`. The [example in my branch](https://github.com/BetterAngelsLA/monorepo/compare/main...DEV-1674/update-document-filename#diff-511f398a3a6759dab6480621ca80e41db06205b2ceebe8aa864c66a6e65a8a76R242) was written before we made this change, which is why it's still in there

apps/betterangels-backend/common/tests/utils.py
```diff
@@ -30,6 +30,7 @@ def _setup_users(self) -> None:
             "org_1_case_manager_2",
             "org_2_case_manager_1",
             "non_case_manager_user",
+            "client_user_1",
```
mikefeldberg, line 33: you can remove `client_user_1` from this file

mikefeldberg (changes requested):

mikefeldberg:
replaced with #1276

PR #1220, opened 2025-04-11
 note & client profile dual write cleanup (DEV-1611, DEV-1652)

mikefeldberg:
DEV-1611, DEV-1652

## Summary by Sourcery

Migrate client profile and note models to remove direct user references, focusing on using client profile as the primary entity

Bug Fixes:
- Fixed potential data integrity issues by decoupling client profile from user model
- Improved data validation by centralizing client information in client profile

Enhancements:
- Refactored client profile and note models to remove direct user dependencies
- Updated validation and query methods to work with client profile instead of user

Chores:
- Removed user-related fields from various models
- Updated GraphQL schema to remove user-related fields
- Modified test cases to work with new client profile structure


apps/betterangels-backend/clients/models.py
```diff
@@ -96,7 +96,7 @@ class Meta:
     pghistory.DeleteEvent("client_profile.remove"),
 )
 class ClientProfile(BaseModel):
-    user = models.OneToOneField(User, on_delete=models.CASCADE, related_name="client_profile")
+    user = models.OneToOneField(User, on_delete=models.SET_NULL, blank=True, null=True, related_name="client_profile")
```
vecchp, line 99: Makes sense, this is optional so we can associate users if we want in the future.

Do we have a ticket to delete client user accounts?

vecchp (approved):

PR #1221, opened 2025-04-11
filter shelters by map bounds (DEV-1699)

mikefeldberg:
DEV-1699

not sure if "bounds" is the best name for the filter param. It was just the first thing I thought of. Open to suggestion

## Summary by Sourcery

Add support for filtering shelters by a polygon-defined map bounds

New Features:
- Implement a new GraphQL filter that allows filtering shelters based on a geographical polygon boundary

Enhancements:
- Extend the shelter query filtering capabilities to support geographical polygon-based filtering

Tests:
- Add a test case to verify shelter filtering by polygon bounds, checking that shelters within the specified geographical area are correctly returned


apps/betterangels-backend/shelters/types.py
```diff
@@ -195,6 +196,18 @@ def properties(
 
         return queryset.filter(**filters).distinct(), Q()
 
+    @strawberry_django.filter_field
+    def bounds(
+        self,
+        queryset: QuerySet,
+        value: Optional[Polygon],
```
tglaz, line 203: think it may be preferable to pass in just the bounds (not Polygon) from the FE and convert to a Polygon  on the BE via `Polygon.from_bbox`?


Can't be certain until start on it, but from google maps we get bounds by using `map.getBounds()`, and bounds is typically defined as the 4 edges: N (lat), S (lat), E (lng), W (lng)

apps/betterangels-backend/schema.graphql
```diff
@@ -1384,6 +1389,7 @@ input ShelterFilter {
   NOT: ShelterFilter
   DISTINCT: Boolean
   properties: ShelterPropertyInput
+  bounds: Polygon
```
tglaz, line 1392: left related comment below, but we'd probably want to pass in a `bounds` object from the FE, but that's a bit different from a Polygon.

it'd basically be like this:
https://developers.google.com/maps/documentation/javascript/reference/coordinates#LatLngBoundsLiteral

tglaz (changes requested):

apps/betterangels-backend/schema.graphql
tglaz: Technically we could deduce the 4 points on FE to pass to the API, so I guess it's more of a question about which interface is easer/better. Could perhaps leave it as is and then address it when it's implemented.

But if `bounds` accepts 4 points (with lat/lng), then perhaps may rename the prop to `boundsCoordinates` or `boundsCoords` or `boundsPoints`?  Or maybe not. I just know that in the FE, with the google maps API, the word `bounds` comes up and just wouldn't want to confuse them if they differ.

apps/betterangels-backend/shelters/types.py
mikefeldberg: I'm down to change the input to fit the maps data. Just want to make sure I'm reading this correctly:
> Object literals are accepted in place of LatLngBounds objects throughout the API. These are automatically converted to LatLngBounds objects
`north`, `south`, `east`, `west` are numbers, but they're used to create a `LatLngBounds` bbox, right?

tglaz: Yes, in this scenario FE passes just 4 numbers defining the boundary

But I’d be probably ok keeping it as is and changing it if it’s an issue, cause shouldn’t be too complicated (i think). And until we work with it I can't be sure which approach is better or if it's even an issue.

PR #1222, opened 2025-04-12
[DRAFT] appUpdate component

tglaz:
### Periodic EAS Update Checks & User Prompt
https://betterangels.atlassian.net/browse/DEV-1584

* feature behind `ffAppUpdatePrompt` flag
  * already added in DEV and PRD
  * turned on for superusers and staff
* show update prompt if:
  * app came into foreground
  * user has not dismissed the prompt in the last X days 
    * current value set low on purpose to allow testing in PRD (behind feature flag)
  * an updated check has not happened in last X minutes
    * just a safety thing to prevent spamming Updates
    * current value set low on purpose to allow testing in PRD (behind feature flag)
* added `ffShowDebugInfo` feature flag
  * allows viewing certain debug info in app (settings/about)
  * already added in DEV and PRD
  * turned on for superusers and staff

## Summary by Sourcery

Implement an app update mechanism using Expo Updates with feature flag control

New Features:
- Add an AppUpdate component to check for and manage app updates using Expo Updates

Enhancements:
- Create a custom hook to track app state changes
- Add a feature flag for controlling app update prompts

Chores:
- Add expo-updates package to project dependencies


libs/expo/betterangels/src/lib/providers/featureControls/constants.ts
```diff
@@ -1,6 +1,8 @@
 export const FeatureFlags = {
   PROFILE_REDESIGN_FF: 'ffClientProfileRedesign',
   CLIENT_DEDUPE_FF: 'ffClientDedupe',
```
vecchp, line 3: If we don't we should create tech debt tickets for each of these flags to make sure we keep track of their removal.

libs/expo/betterangels/src/lib/ui-components/AppUpdate/checkForUpdate.ts
```diff
@@ -0,0 +1,19 @@
+import * as Updates from 'expo-updates';
+
+const isLocalEnv = process.env.NODE_ENV === 'development';
```
vecchp, line 3: It may be better to have a env which turns the feature on vs off instead of checking for node env.

libs/expo/betterangels/src/lib/ui-components/AppUpdate/checkForUpdate.ts
```diff
@@ -0,0 +1,19 @@
...
+  try {
+    const update = await Updates.checkForUpdateAsync();
+
+    return update.isAvailable;
+  } catch (e) {
+    console.log(e);
+
+    return false;
+  }
+}
```
vecchp, line 19: Whats the difference between this and here?

https://github.com/BetterAngelsLA/monorepo/pull/1222/files#diff-a6788637fd7fd6d8467fa311bc67fd47d35c0be0de38c5b0b2c5711c9004814aR18-R31

apps/betterangels/src/app/_layout.tsx
```diff
@@ -38,6 +41,12 @@ export default function RootLayout() {
               <UserProvider>
                 <SnackbarProvider>
                   <StatusBar style="light" />
+                  <FeatureFlagControlled
+                    flag={FeatureFlags.APP_UPDATE_PROMPT_FF}
```
vecchp, line 45: Good improvement, is this pattern from other feature flag sdk's you've used?

libs/expo/betterangels/src/lib/hooks/appState/useAppState.ts
```diff
@@ -0,0 +1,44 @@
...
+
+  const handleAppStateChange = (newState: AppStateStatus) => {
+    setCurrentState((prevState) => {
+      const becameActive = prevState !== 'active' && newState === 'active';
+
+      setBecameActive(becameActive);
+
+      return newState;
+    });
+  };
```
vecchp, line 38: Won't this redefine the function each render?  Should we consider moving it above the effect and wrapping it in useCallback help ensure it’s created only once per hook instance and prevent unnecessary re-subscriptions?

libs/expo/betterangels/src/lib/hooks/appState/useAppState.ts
```diff
@@ -0,0 +1,44 @@
+import { useEffect, useState } from 'react';
+import {
+  AppState,
+  AppStateStatus,
+  NativeEventSubscription,
+} from 'react-native';
+
+let appStateSubscription: NativeEventSubscription | null = null;
```
vecchp, line 8: I’m not very good with React, but I think we might want to use a local ref (via useRef) for the subscription instead of sharing it, to avoid issues when multiple instances of the hook are mounted.

libs/expo/betterangels/src/lib/hooks/appState/useAppState.ts
```diff
@@ -0,0 +1,44 @@
...
+
+export default function useAppState() {
+  const [currentState, setCurrentState] = useState<AppStateStatus>(
```
vecchp, line 11: Should we consider using a reducer here to combine them into a single atomic update? This might help ensure consistency and reduce the chance of race conditions when the app state changes.

libs/expo/betterangels/src/lib/screens/AppSettings/AboutApp/UpdatesDebugInfo.tsx
```diff
@@ -0,0 +1,72 @@
...
+      return;
+    }
+
+    try {
+      const newUpdateStatus: Updates.UpdateCheckResult =
+        await Updates.checkForUpdateAsync();
+
+      setUpdateAvailable(newUpdateStatus.isAvailable);
+    } catch (e) {
+      console.log(e);
```
vecchp, line 29: Should we instead use console.error or another error tracking mechanism, especially since this is a debug info component?

libs/expo/betterangels/src/lib/ui-components/AppUpdate/AppUpdate.tsx
```diff
@@ -0,0 +1,67 @@
...
+      // reset storage data before reload
+      await Promise.all([
+        AsyncStorage.setItem(UPDATE_DISMISSED_TS_KEY, ''),
+        AsyncStorage.setItem(LAST_UPDATE_CHECK_TS_KEY, String(Date.now())),
+      ]);
+
+      // reload app with update
+      await Updates.reloadAsync();
+    } catch (e) {
+      console.log('Error updating app:', e);
```
vecchp, line 51: Console.error?

libs/expo/betterangels/src/lib/ui-components/AppUpdate/checkForUpdate.ts
```diff
@@ -0,0 +1,19 @@
...
+  if (isLocalEnv) {
+    return false;
+  }
+
+  try {
+    const update = await Updates.checkForUpdateAsync();
+
+    return update.isAvailable;
+  } catch (e) {
+    console.log(e);
```
vecchp, line 15: console.error?

vecchp (commented):
Good stuff here are some comments.

libs/expo/betterangels/src/lib/hooks/appState/useAppState.ts
tglaz: i'll check it out.
but this already ensures (or is meant to ensure) that only a single subscription exists when multiple components use the hook.

libs/expo/betterangels/src/lib/hooks/appState/useAppState.ts
tglaz: may need to ask you about cause not 100% sure I'm understanding correctly

libs/expo/betterangels/src/lib/providers/featureControls/constants.ts
tglaz: for `PROFILE_REDESIGN_FF` it's noted on this placeholder cleanup ticket- https://betterangels.atlassian.net/browse/DEV-1547

Regarding others I'm ok either way.

libs/expo/betterangels/src/lib/ui-components/AppUpdate/checkForUpdate.ts
tglaz: Yeah, meant to ask about this.
Ideally would be able to access `channel` from `eas.json` via expo-constants, but I don't think it was there. It may be defined in PRD though - i've added it to a debug section to check after deploy.

libs/expo/betterangels/src/lib/ui-components/AppUpdate/checkForUpdate.ts
tglaz: yeah, they're same. I'll update.

The other is in the `debugInfo` component and originally did it to be standalone.

apps/betterangels/src/app/_layout.tsx
tglaz: I've seen this before, but the main reasons were:
1. In `_layout.tsx` function the `FeatureControlContext` is not available
2. It's much more explicit as to which feature is behind a featureFlag. And if you do the check inside the component then that code still actually runs, even if it may return null

libs/expo/betterangels/src/lib/hooks/appState/useAppState.ts
tglaz: yeah, and I can look at it again.
but small function like that is quite cheap and react is pretty efficient with that.

libs/expo/betterangels/src/lib/screens/AppSettings/AboutApp/UpdatesDebugInfo.tsx
tglaz: yup, i'll update. 
I typically use console.error and just missed these

libs/expo/betterangels/src/lib/ui-components/AppUpdate/checkForUpdate.ts
tglaz: I switched to using `channel` from `Updates`. In theory should work.

```javascript
export async function checkForUpdate() {
  const { channel, checkForUpdateAsync } = Updates;

  if (channel !== 'production') {
    return {} as Updates.UpdateCheckResult;
  }
  ```

File: libs/expo/betterangels/src/lib/hooks/appState/useAppState.ts (hunk 6 above)
Comment (tglaz) [reply libs/expo/betterangels/src/lib/hooks/appState/useAppState.ts]: updated

File: libs/expo/betterangels/src/lib/hooks/appState/useAppState.ts (hunk 5 above)
Comment (tglaz) [reply libs/expo/betterangels/src/lib/hooks/appState/useAppState.ts]: updated

File: libs/expo/betterangels/src/lib/hooks/appState/useAppState.ts (hunk 7 above)
Comment (tglaz) [reply libs/expo/betterangels/src/lib/hooks/appState/useAppState.ts]: made an update that may address this:

```javascript
  const handleAppStateChange = useCallback((newState: AppStateStatus) => {
    setAppState((prevState) => {
      const { currentAppState: prevCurrent } = prevState;
      const becameActive = prevCurrent !== 'active' && newState === 'active';

      return {
        currentAppState: newState,
        appBecameActive: becameActive,
      };
    });
  }, []);
```

tglaz:
closing. feature merged as PR #1229

PR #1224, opened 2025-04-15
Shelter db multifile upload with django-s3file

jluna2000:
Implemented multi file upload fields for each of the media upload fields. Used the django ClearableFIleInput widget and corresponding fields to do so instead of AsyncFileField. Removed progress bars from each of the upload fields after having added them to avoid confusion, as I learned that these progress bars would only start upon form submission when file upload begins. Left one progress bar next to save button once form submission happens. [django-s3file](https://github.com/codingjoe/django-s3file) module was added for direct upload to s3 capability. Selected this module in particular as it integrates with django's ClearableFileInput widget directly, and has options for a dummy upload django route while in development environment.

## Summary by Sourcery

Implement multi-file upload functionality for shelter media using django-s3file and ClearableFileInput

New Features:
- Added support for multiple file uploads for interior photos, exterior photos, and videos in the shelter admin interface

Enhancements:
- Replaced single file upload with multi-file upload capability
- Added a form-level progress bar for file uploads
- Integrated django-s3file for direct S3 uploads

Chores:
- Updated admin form to handle multiple file uploads
- Modified inline admin views to remove extra empty fields

### DIFF
--- a/apps/betterangels-backend/betterangels_backend/settings.py
+++ b/apps/betterangels-backend/betterangels_backend/settings.py
@@ -128,6 +128,7 @@ INSTALLED_APPS = [
     "django_select2",
     "import_export",
     "rangefilter",
+    "s3file",
     "strawberry_django",
     "waffle",
     # Our Apps
@@ -153,6 +154,7 @@ MIDDLEWARE = [
     "django.middleware.clickjacking.XFrameOptionsMiddleware",
     "allauth.account.middleware.AccountMiddleware",
     "pghistory.middleware.HistoryMiddleware",
+    "s3file.middleware.S3FileMiddleware",
     # Our Middleware
     "common.middleware.TimezoneMiddleware",
 ]
@@ -330,6 +332,7 @@ if env("AWS_S3_MEDIA_STORAGE_ENABLED"):
         },
     }
 
+AWS_STORAGE_BUCKET_NAME = ""
 
 ADMIN_RESUMABLE_CHUNKSIZE = 10 * 1024 * 1024
 ADMIN_RESUMABLE_SHOW_THUMB = True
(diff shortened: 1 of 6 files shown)


apps/betterangels-backend/betterangels_backend/settings.py
```diff
@@ -330,6 +332,7 @@
         },
     }
 
+AWS_STORAGE_BUCKET_NAME = ""
```
vecchp, line 335: I don't think this should be set up here, i think we just need to use django storages which should already be configured by this block

```
if env("AWS_S3_MEDIA_STORAGE_ENABLED"):
    STORAGES = {
        "default": {
            "BACKEND": "storages.backends.s3.S3Storage",
            "OPTIONS": {
                "bucket_name": env("AWS_S3_STORAGE_BUCKET_NAME"),
                "cloudfront_key": env("AWS_CLOUDFRONT_KEY").encode("ascii"),
                "cloudfront_key_id": env("AWS_CLOUDFRONT_KEY_ID"),
                "custom_domain": env("AWS_S3_CUSTOM_DOMAIN"),
                "location": env("AWS_CLOUDFRONT_MEDIA_LOCATION"),
                "signature_version": "s3v4",
            },
        },
        "staticfiles": {
            "BACKEND": "django.contrib.staticfiles.storage.StaticFilesStorage",
        },
    }
```

vecchp (changes requested):
good stuff here is some feedback

apps/betterangels-backend/shelters/admin.py
```diff
@@ -401,6 +405,23 @@ def _clean_choices(self, field_name: str, model_class: Type[T]) -> list[T]:
 
         return existing_objects
 
+    def save(self, commit: bool = True) -> Any:
```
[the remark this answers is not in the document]
vecchp: @jluna2000 this might be a useful recommendation.

apps/betterangels-backend/betterangels_backend/settings.py
jluna2000: Removing that line causes the following error: `AttributeError: 'Settings' object has no attribute 'AWS_STORAGE_BUCKET_NAME'"}`
Maybe bucket_name/AWS_STORAGE_BUCKET_NAME isn't set up by default in a dev environment?

PR #1229, opened 2025-04-16
implement AppUpdatePrompt component (DEV-1584)

tglaz:
### Periodic EAS Update Checks & User Prompt
https://betterangels.atlassian.net/browse/DEV-1584

* feature behind `ffAppUpdatePrompt` flag
  * already added in DEV and PRD
  * turned on for superusers and staff
* show update prompt if:
  * app came into foreground
  * user has not dismissed the prompt in the last X days 
    * current value set low on purpose to allow testing in PRD (behind feature flag)
  * an updated check has not happened in last X minutes
    * just a safety thing to prevent spamming Updates
    * current value set low on purpose to allow testing in PRD (behind feature flag)
* added `ffShowDebugInfo` feature flag
  * allows viewing certain debug info in app (settings/about)
  * already added in DEV and PRD
  * turned on for superusers and staff

**Also**
* update`ActionModal` to accept optional `body` 
* update `GoogleSignIn` to use `useAppState` hook
* add icon

image

## Summary by Sourcery

Implement an app update prompt feature with periodic EAS update checks and user notifications, controlled by feature flags

New Features:
- Add AppUpdatePrompt component to check and prompt users about available app updates
- Introduce a debug info view for updates in the About App screen

Enhancements:
- Update ActionModal to support optional body content
- Refactor GoogleSignIn to use new useAppState hook
- Add useAppState hook for tracking app state changes

Tests:
- Implement logic to prevent update prompt spamming
- Add conditions for showing update prompt based on app state and time intervals

Chores:
- Add new feature flags for update prompts and debug info
- Update package dependencies to include expo-updates


Davit-BetterAngels (approved):

libs/expo/betterangels/src/lib/ui-components/AppUpdatePrompt/AppUpdatePrompt.tsx
```diff
@@ -0,0 +1,67 @@
...
+      const { isAvailable } = await checkForUpdate();
+
+      if (isAvailable) {
+        setPromptModalVisible(true);
+      }
+    };
+
+    if (!appBecameActive) {
+      return;
+    }
```
vecchp, line 31: Is this more appropriate logic?

```suggestion
if (!appBecameActive) return;

const canShowAgain = await canShowPromptAgain();
if (!canShowAgain) return;

const { isAvailable } = await checkForUpdate();
if (isAvailable) {
  setPromptModalVisible(true);
}

```

libs/expo/betterangels/src/lib/ui-components/AppUpdatePrompt/AppUpdatePrompt.tsx
```diff
@@ -0,0 +1,67 @@
...
+    setPromptModalVisible(false);
+
+    try {
+      await Updates.fetchUpdateAsync();
+
+      // reset storage data before reload
+      await Promise.all([
+        AsyncStorage.setItem(UPDATE_DISMISSED_TS_KEY, ''),
+        AsyncStorage.setItem(LAST_UPDATE_CHECK_TS_KEY, String(Date.now())),
+      ]);
```
vecchp, line 46: Wouldn't it make more sense to clear these after a successful update?

libs/expo/betterangels/src/lib/ui-components/AppUpdatePrompt/checkForUpdate.ts
```diff
@@ -0,0 +1,15 @@
...
+  if (channel !== 'production') {
+    return {} as Updates.UpdateCheckResult;
+  }
+
+  try {
+    return await checkForUpdateAsync();
+  } catch (e) {
+    return {} as Updates.UpdateCheckResult;
+  }
+}
```
vecchp, line 15: we have preview builds as well, those could have eas updates.

vecchp: Might make sense to have a env var which enables eas updates vs checking channels.

tglaz: yeah, I had the `preview` question saved for our meeting later. 

Regarding env var, we already have `APP_VARIANT` it seems but only defined for `production`
So I'm ok with Env Var, but using `channel` may also be appropriate as it's specific to Updates. Either way, probably may want to settle one one.

libs/expo/betterangels/src/lib/ui-components/AppUpdatePrompt/AppUpdatePrompt.tsx
tglaz: I put it here cause an update may fail for various reasons and they don't recommend putting any logic after.
This would essentially block the next Prompt for the period `MIN_UPDATE_CHECK_INTERVAL_MS`, which would be a minute or two.

But honestly can be convinced either way

libs/expo/betterangels/src/lib/ui-components/AppUpdatePrompt/AppUpdatePrompt.tsx
tglaz: can you compare it to this version the version in PR #1231 ?
I updated it there to allow bypassing the `canShowPromptAgain` for the Crash scenario

libs/expo/betterangels/src/lib/ui-components/AppUpdatePrompt/AppUpdatePrompt.tsx
tglaz: will add comment

PR #1242, opened 2025-04-21
website label shelter (DEV-1711)

Davit-BetterAngels:
DEV-1711

## Summary by Sourcery

Enhance website label rendering for shelter information by adding website detection and formatting

New Features:
- Add intelligent website label rendering that converts website strings into clickable links

Enhancements:
- Create utility functions to detect and format website URLs
- Improve label display with automatic website link generation


libs/react/components/src/utils/formatWebsite.ts
```diff
@@ -0,0 +1,4 @@
+export default function formatWebsite(value: string): string {
+  if (value.startsWith('http')) return value;
+  return `https://${value}`;
```
tglaz, line 3: if this is for external sites, then we may not want to always convert to https. It may actually be `http` only.
if we don't want to handle insecure domains then I suppose that's differnt.

libs/react/components/src/utils/isWebsite.ts
```diff
@@ -0,0 +1,7 @@
+export default function isWebsite(value: string | undefined): boolean {
+  if (!value) return false;
+
+  const isDomainLike = value.includes('.') && !value.includes(' ');
+
+  return value.startsWith('http') || value.startsWith('www.') || isDomainLike;
```
tglaz, line 6: this mostly works, but may want to consider using a lib to get a more robust check

libs/react/components/src/utils/isWebsite.ts
```diff
@@ -1,7 +1,9 @@
-export default function isWebsite(value: string | undefined): boolean {
-  if (!value) return false;
+import validator from 'validator';
 
-  const isDomainLike = value.includes('.') && !value.includes(' ');
-
-  return value.startsWith('http') || value.startsWith('www.') || isDomainLike;
+export default function isWebsite(value: string) {
```
tglaz, line 3: how about using `isWebURL` for the name?

Also, may want to consider wrapping it in a try/catch just to be safe.
Even though it only accept a string, in practice it'll be used values we get from the database, and we can't always guarantee it's not null or undefined etc.

libs/react/components/src/utils/formatWebsite.ts
```diff
@@ -1,4 +1,7 @@
 export default function formatWebsite(value: string): string {
-  if (value.startsWith('http')) return value;
+  if (value.startsWith('http')) {
```
tglaz, line 2: couple possible edits:
* name: `toWebURL`, `toValidWebURL`, `formatWebURL`, `withURLProtocol`
* validate value with `isWebsite` first to avoid an invalid `http://random-string`

tglaz (changes requested):

libs/react/components/src/utils/isWebsite.ts
```diff
@@ -1,7 +1,9 @@
-export default function isWebsite(value: string | undefined): boolean {
-  if (!value) return false;
+import validator from 'validator';
```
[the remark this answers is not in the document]
Davit-BetterAngels: added one more line for safety but ignored try catch couse it is pure function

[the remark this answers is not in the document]
tglaz: yup. that'll work also

libs/react/components/src/utils/toValidWebURL.ts
```diff
@@ -0,0 +1,9 @@
+import isWebsite from './isWebsite';
+
+export default function toValidWebURL(value: string): string {
+  if (isWebsite(value)) {
+    return value;
+  }
+
+  return `https://${value}`;
```
tglaz, line 8: couple small comments:

1. Given that this is called `toValidWebURL`, would it be more consistent if it called `isValidWebURL` instead of `isWebsite`?
2. The return value here is called when `isWebsite` returns false, so technically it may not be a valid URL, and it'll just add `hrttps` protocol to any string, even if invalid.
* should it first validate if `isValidWebURL` and return false if not?
* and should it assume that https is valid if protocol is not provided, or just return the `value` as is if it passes the URL check? The `Website` field in admin validates URL and adds protocol, but can be either http or https, but this is a generic util, so should work with everything.

libs/react/shelter/src/lib/pages/shelter/GeneralInfo/index.tsx
```diff
@@ -11,6 +11,26 @@ import parsePhoneNumber from 'libphonenumber-js';
...
 
+function renderLabel(label?: string | null) {
+  if (!label) {
+    return 'Not Available';
+  }
+
+  if (isWebsite(label)) {
+    return (
+      <a
+        href={toValidWebURL(label)}
```
tglaz, line 22: calling `isWebsite` and then `toValidWebURL` is fine, though could also just do 
`const webURL = toValidWebURL(lablel)` and then check `if (webURL) { ...
though this assumes toValidWebURL returns false if invalid, which it does not right now

libs/react/components/src/lib/index.ts
```diff
@@ -1,3 +1,5 @@
+export { default as isWebsite } from '../utils/isWebsite';
+export { default as toValidWebURL } from '../utils/toValidWebURL';
 export { default as Button } from './Button';
```
tglaz, line 3: btw, `utils` has its own `index` file, so can probably export from there.

And at some point we may want to break out `utils` out of `components`, though not urgent
-- react
 |-- components
 |-- icons
 |-- shelter
 |-- utils

tglaz (approved):

shiozakiyuko (approved):

PR #1259, opened 2025-04-24
default map behavior shelter (DEV-1633)

Davit-BetterAngels:
DEV-1633

## Summary by Sourcery

Enhancements:
- Improve map initialization logic to consistently set both default center and current location based on saved or default coordinates


tglaz (approved):
approving, but I think we may need to revisit how we're tracking and naming all of the location stuff, because it may be getting a bit confusing.

PR #1260, opened 2025-04-24
hide location button if permission denied (DEV-1761)

Davit-BetterAngels:
DEV-1761

## Summary by Sourcery

Improve geolocation handling by hiding the current location button when location permission is denied

Bug Fixes:
- Add handling for geolocation permission denial by checking permission state before showing location button

Enhancements:
- Implement permission check for geolocation to provide better user experience
- Add dynamic visibility for current location button based on location permissions

### DIFF
--- a/apps/shelter-web/src/app/shared/components/map/controls/currentLocationBtn.tsx
+++ b/apps/shelter-web/src/app/shared/components/map/controls/currentLocationBtn.tsx
@@ -7,12 +7,12 @@ import { ControlBtnWrapper } from './controlBtnWrapper';
 
 type TProps = {
   className?: string;
-  onLocationSucccess?: (location: TLatLng) => void;
+  onLocationSuccess?: (location: TLatLng) => void;
   onError?: (location: TLocationError) => void;
 };
 
 export function CurrentLocationBtn(props: TProps) {
-  const { className = '', onError, onLocationSucccess } = props;
+  const { className = '', onError, onLocationSuccess } = props;
 
   const map = useMap();
 
@@ -21,7 +21,7 @@ export function CurrentLocationBtn(props: TProps) {
   }
 
   function onLocationChange(location: TLatLng) {
-    onLocationSucccess && onLocationSucccess(location);
+    onLocationSuccess && onLocationSuccess(location);
   }
 
   function onLocationError(error: TLocationError) {
--- a/apps/shelter-web/src/app/shared/components/map/map.tsx
+++ b/apps/shelter-web/src/app/shared/components/map/map.tsx
@@ -118,9 +118,14 @@ export function Map(props: TMap) {
 
   useEffect(() => {
     let permissionStatus: PermissionStatus;
+    const permissionQuery = navigator.permissoms;
 
-    navigator.permissions
-      ?.query({ name: 'geolocation' as PermissionName })
+    if (!permissionQuery) {
+      return;
+    }
+
+    permissionQuery
+      .query({ name: 'geolocation' as PermissionName })
       .then((result) => {
         setGeolocationPermission(result.state);
         permissionStatus = result;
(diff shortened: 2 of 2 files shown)


kimchoi-jjiggae (approved):

apps/shelter-web/src/app/shared/components/map/controls/currentLocationBtn.tsx
```diff
@@ -7,12 +7,12 @@ import { ControlBtnWrapper } from './controlBtnWrapper';
```
shiozakiyuko, line 7: Hey @Davit-BetterAngels 

I tested it on iOS and Android. iOS looks good.

On Android, when the location is turned off, I see a weird position of the +/- buttons (until you refresh the page) and also in the drawer-ish section. Please see the attachments.

apps/shelter-web/src/app/shared/components/map/map.tsx
```diff
@@ -117,19 +118,33 @@ export function Map(props: TMap) {
...
+            setUserLocation(newCenter);
+          },
+          (error) => {
+            console.error('Geolocation error', error);
+            setLocationAllowed(false);
+          },
+          { enableHighAccuracy: true }
+        );
+
+        result.onchange = () => {
```
tglaz, line 142: There are couple ways to do this, but I'd suggest splitting up the useEffect into the separate concerns. Track permissions changes first and then react to them.

1. a useEffect to track `geolocationPermission` in state
  * run single useEffect on map change, with the result.onChange
  * clean up the onChange handler on unmount
 
2. another useEffect that reacts to `geolocationPermission change`
  * here can run `getCurrentPosition`

apps/shelter-web/src/app/shared/components/map/map.tsx
```diff
@@ -68,6 +68,7 @@ export function Map(props: TMap) {
   });
   const [userLocation, setUserLocation] =
     useState<google.maps.LatLngLiteral | null>(null);
+  const [locationAllowed, setLocationAllowed] = useState(true);
```
tglaz, line 71: would defaulting to `true` cause the button to be visible and then possibly hide?
the permissions check should be quick, so think that defaulting to `false` and then showing button only if changes to `true` should be ok

tglaz (changes requested):

apps/shelter-web/src/app/shared/components/map/map.tsx
tglaz: I think the issue may be that `locationAllowed` actually tracks if permission !== `Denied`, which is not the same as allowed. There is also the "unknown" or "prompt" state.

Could just store the actual permission.state or change it to `locationDenied`?

apps/shelter-web/src/app/shared/components/map/map.tsx
```diff
@@ -115,7 +117,28 @@ export function Map(props: TMap) {
...
+
+    return () => {
+      if (permissionStatus) {
+        permissionStatus.onchange = null;
+      }
+    };
+  }, []);
+
+  useEffect(() => {
+    if (!map || geolocationPermission !== 'granted') return;
```
tglaz, line 141: or should it be  `geolocationPermission === 'denied'`?
just double checking the logic, but if it's `null` | `prompt` then we should still call `getCurrentPosition` and it'll ask for permissions if needed,

I'd also return if `!navigator.geolocation`

apps/shelter-web/src/app/shared/components/map/map.tsx
tglaz: final small thing I noticed:

I think with the `permissions?.query`, you can actually call `.then` on undefined, so may be safer to check if `query` exists and only then call it

apps/shelter-web/src/app/shared/components/map/map.tsx
Davit-BetterAngels: add !navigator.geolocation

but to check denied no need because if denied you will not see the current button to prompt

tglaz (approved):

PR #1276, opened 2025-05-05
add update document name mutation (DEV-1674)

mikefeldberg:
DEV-1674

## Summary by Sourcery

Add a mutation to update the name of a client document with appropriate permission checks

New Features:
- Implement update client document name mutation

Tests:
- Add test cases for updating client document name with various user permission scenarios


tglaz (approved):
lgtm

The only comment would be our usage of the `original_filename`, which makes it confusing as that's not what it is.
Perhaps could add a tech-debt ticket to rename it, or track original_filename and display_filename separately, but ok for now.

PR #1311, opened 2025-05-21
cluster pin icon (DEV-1684)

Davit-BetterAngels:
DEV-1684

## Summary by Sourcery

Introduce a new MapClusterMarker component for displaying styled cluster pins on maps and extend the shared theme with an orange color

New Features:
- Add MapClusterMarker component with configurable size, text, variant, subscript, and optional house icon
- Expose SIZES and variantStyleMap constants for cluster marker styling

Enhancements:
- Add ORANGE to the shared Colors enum


libs/expo/shared/ui-components/src/lib/MapClusterMarker/MapClusterMarker.tsx
```diff
@@ -0,0 +1,129 @@
...
+          <View
+            style={[
+              styles.house,
+              {
+                height: SIZES[size].houseContainer,
+                width: SIZES[size].houseContainer,
+              },
+            ]}
+          >
+            {/* TODO: We don't have the icon which is on FIGMA (AC said ignore the icon for now) */}
```
tglaz, line 84: @Davit-BetterAngels  I'd just remove everything related to `house`. We may never use that feature, and if so, we'd do a new ticket.
If wanna keep the code then I'd suggest moving to another file and just not importing it yet.

tglaz (approved):

PR #1318, opened 2025-05-22
center to current location (DEV-1734)

Davit-BetterAngels:
DEV-1734

## Summary by Sourcery

Enable a new Locations screen with map centering functionality, introduce a reusable MapView and location hook, and refactor client tab management to support the new feature via an enum and feature flag.

New Features:
- Add a Locations tab in the client view that shows a map and a button to center on the user’s current location
- Introduce a MapView component in shared UI components with support for react-native-maps (including web fallback)
- Add a useUserLocation hook to request permissions and retrieve the user’s current location

Enhancements:
- Refactor ClientTabs to use a ClientViewTabEnum, externalize styles with StyleSheet, and toggle the Locations tab via feature flag
- Update the Client screen to integrate the new Locations tab, hide the header when on Locations, and use the enum for tab state

Chores:
- Add shared map library files (constants, types, mapLib) and Google Maps typings


libs/expo/betterangels/src/lib/screens/Client/ClientTabs.tsx
```diff
@@ -1,83 +1,61 @@
 import { Colors, FontSizes, Spacings } from '@monorepo/expo/shared/static';
```
shiozakiyuko, line 1: Hi @Davit-BetterAngels 

When the location setting is ON, it works beautifully on both iOS and Android. However, when the location setting is OFF, Android shows the dialog to ask to turn the Location ON at the initial view (but it doesn't happen every time). Can iOS side have that as well because otherwise, there is no feedback when I tapped the current location button.
Let me know if you can turn on the native notification dialog or what you think.

Thanks!


**iOS**
https://github.com/user-attachments/assets/6b6bab49-1166-4c23-b2c4-a1bff42c087e

**Android** (sorry it's long but you will see what I mean by the "native dialog")
https://github.com/user-attachments/assets/2b37ea45-5b2a-4d5d-8eb2-f5c71ac032ad

libs/expo/betterangels/src/lib/screens/Client/Locations/Locations.tsx
```diff
@@ -1,12 +1,53 @@
...
+  };
+
+  return (
+    <View>
+      <MapView
+        ref={mapRef}
+        showsUserLocation={userLocation ? true : false}
+        style={styles.map}
+      />
+      <LocateMeButton
```
tglaz, line 32: wouldn't we want to place the `LocateMeButton` directly inside the Map.tsx `RNMapView` component and enable it based off of `showsUserLocation` prop?

I already approved your `track-user-location` PR, but instead of passing a `showsUserLocation` we could pass `tracUserLocation` Boolean, and then run all of this logic inside the MapView component.

tglaz (changes requested):

libs/expo/betterangels/src/lib/screens/Client/ClientTabs.tsx
shiozakiyuko: If it takes some time or you think it's out of scope, I can create a ticket. Let me know.

libs/expo/betterangels/src/lib/screens/Client/Locations/Locations.tsx
Davit-BetterAngels: good point

tglaz (approved):

libs/expo/betterangels/src/lib/screens/Client/ClientTabs.tsx
Davit-BetterAngels: please lets create another ticket for this one. its not fast to do

shiozakiyuko (approved):
Thanks, @Davit-BetterAngels . I will create another ticket for the issue then. Thanks, and approved!

PR #1321, opened 2025-05-22
fix android oauth (DEV-1840)

mikefeldberg:
DEV-1840

removed the hook from google signin because including it in useEffect dependencies broke android oauth

## Summary by Sourcery

Fix Android OAuth flow by replacing the custom useAppState hook with AppState.currentState and adjusting deep link effect dependencies

Bug Fixes:
- Fix Android OAuth flow by replacing the custom useAppState hook with AppState.currentState to avoid broken redirects

Enhancements:
- Simplify deep link handling effect to only depend on handleDeepLinking

### DIFF
--- a/libs/expo/betterangels/src/lib/ui-components/SocialLogin/GoogleSignIn.tsx
+++ b/libs/expo/betterangels/src/lib/ui-components/SocialLogin/GoogleSignIn.tsx
@@ -12,8 +12,7 @@ import {
   useEffect,
   useState,
 } from 'react';
-import { Linking, View } from 'react-native';
-import { useAppState } from '../../hooks';
+import { AppState, Linking, View } from 'react-native';
 import useSignIn from '../../hooks/user/useSignIn';
 
 export const GOOGLE_AUTH_MUTATION = gql`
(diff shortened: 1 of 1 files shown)


libs/expo/betterangels/src/lib/ui-components/SocialLogin/GoogleSignIn.tsx
```diff
@@ -157,7 +155,7 @@ export function GoogleSignIn({
       https://github.com/expo/expo/issues/12044#issuecomment-1431310529
     */
 
-    if (currentAppState === 'active') {
```
tglaz, line 160: I'd just suggest adding a comment such as 
```
Not using currentAppState from useAppState as it triggers a bug ....
```

I could possibly see someone coming in and reverting it again, as I did. Not sure why using the `useAppState` hook breaks here, but seems like there are special circumstances for Android, so it is what it is for now.
Adding `maestro` test for this could also help, but not sure when that'd happen, and may not be super simple to do in this case.

tglaz (approved):

PR #1322, opened 2025-05-22
Clear button should be in each filter (DEV-1551)

vecchp:


shiozakiyuko (approved):
Thanks, @0nayan! Approved!

PR #1323, opened 2025-05-27
fix uploaded file name & extension (DEV-1778, DEV-1779)

mikefeldberg:
DEV-1778, DEV-1779

### DIFF
--- a/apps/betterangels-backend/common/utils.py
+++ b/apps/betterangels-backend/common/utils.py
@@ -1,7 +1,26 @@
+import mimetypes
 import os
 import uuid
+from typing import TYPE_CHECKING
 
 from django.db.models import Model
+
+
+def get_filename_with_extension(mime_type: str, filename: str) -> str:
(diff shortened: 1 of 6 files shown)


apps/betterangels-backend/common/utils.py
```diff
@@ -1,9 +1,28 @@
...
+
+    if f".{provided_extension}" == expected_extension or (
+        expected_extension and expected_extension == alt_extension_map.get(provided_extension)
+    ):
+        return filename
+
+    if type(expected_extension) is str:
+        return f"{filename}{expected_extension}"
+
+    raise ValueError("Invalid file type")
```
mikefeldberg, line 23: not part of a the ticket and maybe too aggressive. My thinking was that, if we can't tell what the file is, it likely couldn't be opened and shouldn't be uploaded. Happy to take this out if we want to address it another way later

vecchp (approved):

PR #1336, opened 2025-05-30
update interactioncard (DEV-1796)

mikefeldberg:
DEV-1796

### DIFF
--- a/libs/expo/betterangels/src/lib/ui-components/NoteCard/NoteCardByline.tsx
+++ b/libs/expo/betterangels/src/lib/ui-components/NoteCard/NoteCardByline.tsx
@@ -2,13 +2,15 @@ import { Colors, Spacings } from '@monorepo/expo/shared/static';
 import { TextRegular } from '@monorepo/expo/shared/ui-components';
 import { format } from 'date-fns';
 import { View } from 'react-native';
+import { NoteType } from '../../apollo';
 
-interface INoteCardFooterProps {
-  isSubmitted: boolean;
-  interactedAt: string;
+interface INoteCardBylineProps {
+  createdBy: NoteType['createdBy'];
+  organization: NoteType['organization']['name'];
+  team: NoteType['team'];
 }
 
-export default function NoteCardFooter(props: INoteCardFooterProps) {
+export default function NoteCardByline(props: INoteCardBylineProps) {
   const { isSubmitted, interactedAt } = props;
   return (
     <View
--- a/libs/expo/betterangels/src/lib/ui-components/NoteCard/NoteCardByline.tsx
+++ b/libs/expo/betterangels/src/lib/ui-components/NoteCard/NoteCardByline.tsx
@@ -1,43 +1,49 @@
 import { Colors, Spacings } from '@monorepo/expo/shared/static';
-import { TextRegular } from '@monorepo/expo/shared/ui-components';
-import { format } from 'date-fns';
+import { Avatar, TextRegular } from '@monorepo/expo/shared/ui-components';
 import { View } from 'react-native';
 import { NoteType } from '../../apollo';
+import { enumDisplaySelahTeam } from '../../static/enumDisplayMapping';
 
 interface INoteCardBylineProps {
   createdBy: NoteType['createdBy'];
-  organization: NoteType['organization']['name'];
+  organization: NoteType['organization'];
   team: NoteType['team'];
 }
 
 export default function NoteCardByline(props: INoteCardBylineProps) {
-  const { isSubmitted, interactedAt } = props;
+  const { createdBy, organization, team } = props;
   return (
     <View
       style={{
-        justifyContent: 'space-between',
         flexDirection: 'row',
-        alignItems: 'center',
+        justifyContent: 'space-between',
+        marginBottom: Spacings.xs,
       }}
     >
-      <View style={{ flex: 1 }}>
-        <TextRegular size="sm" color={Colors.NEUTRAL_DARK}>
-          {format(new Date(interactedAt), 'MM/dd/yyyy')}
+      <Avatar
+        mr="xs"
+        size="sm"
+        accessibilityLabel={createdBy?.email || 'unknown user'}
+        accessibilityHint={
+          `${createdBy.email} user's avatar` || `user's avatar`
+        }
+      />
+      <View
+        style={{
+          flex: 1,
+          justifyContent: 'space-between',
+          marginBottom: Spacings.xs,
+          marginRight: Spacings.md,
+        }}
+      >
+        <TextRegular size="md" color={Colors.PRIMARY_EXTRA_DARK}>
+          {createdBy.firstName} {createdBy.lastName}
+        </TextRegular>
+        <TextRegular size="md" color={Colors.PRIMARY_EXTRA_DARK}>
+          {organization.name}{' '}
+          {team && <TextRegular>({enumDisplaySelahTeam[team]})</TextRegular>}
         </TextRegular>
       </View>
-      {!isSubmitted && (
-        <View
-          style={{
-            backgroundColor: Colors.PRIMARY_EXTRA_LIGHT,
-            paddingHorizontal: Spacings.xxs,
-            borderRadius: 20,
-          }}
-        >
-          <TextRegular size="sm" color={Colors.NEUTRAL_EXTRA_DARK}>
-            Draft
-          </TextRegular>
-        </View>
-      )}
     </View>
   );
 }
--- a/libs/expo/betterangels/src/lib/ui-components/NoteCard/NoteCardClient.tsx
+++ b/libs/expo/betterangels/src/lib/ui-components/NoteCard/NoteCardClient.tsx
@@ -1,4 +1,4 @@
-import { Colors } from '@monorepo/expo/shared/static';
+import { Colors, Spacings } from '@monorepo/expo/shared/static';
 import { Avatar, TextRegular } from '@monorepo/expo/shared/ui-components';
 import { View } from 'react-native';
 import { NoteType } from '../../apollo';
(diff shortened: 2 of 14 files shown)


libs/expo/betterangels/src/lib/ui-components/NoteCard/NoteCardByline.tsx
```diff
@@ -0,0 +1,57 @@
...
+    <View
+      style={{
+        flexDirection: 'row',
+        marginBottom: Spacings.xs,
+      }}
+    >
+      <Avatar
+        mr="xs"
+        size="sm"
+        accessibilityLabel={createdBy?.email || 'unknown user'}
```
tglaz, line 29: Just a comment, but I'm not sure including email in in any accessibility markup is helpful, especially when it's a non-actionable element and we render the user name.
But mostly just feels weird to proliferate PII.
Having said that, not 100% what the correct way is.

libs/expo/betterangels/src/lib/ui-components/NoteCard/NoteCardClient.tsx
```diff
@@ -1,58 +1,34 @@
...
-            Draft
-          </TextRegular>
-        </View>
-      )}
+      <Avatar
+        mr="xs"
+        size="sm"
+        accessibilityLabel={clientProfile?.email || 'unknown user'}
+        accessibilityHint={
+          `${clientProfile?.email} client's avatar` || `client's avatar`
```
tglaz, line 26: `${clientProfile?.email} client's avatar` is a string, so will always evaluate to `True`.

`accessibilityHint='client's avatar'` may perhaps suffice.

libs/expo/betterangels/src/lib/ui-components/NoteCard/NoteCardPills.tsx
```diff
@@ -15,28 +16,74 @@ export default function NoteCardPills({
...
   type: 'success' | 'primary' | 'warning';
 }) {
+  const [containerWidth, setContainerWidth] = useState(0);
+  const [pillWidths, setPillWidths] = useState<number[]>(
+    Array(services.length).fill(0)
+  );
+  const [visibleCount, setVisibleCount] = useState(services.length);
+
+  useEffect(() => {
+    if (containerWidth > 0 && pillWidths.every((w) => w > 0)) {
```
tglaz, line 26: We have a `<PillContainer>` component which may or may not do the same thing. 
Would preferable to store this functionality there unless there's some reason not to.

libs/expo/shared/ui-components/src/lib/TextBold/TextBold.tsx
```diff
@@ -24,7 +24,7 @@ interface ITextBoldProps extends TextProps {
    * size:
    * - xxs: 10 xs: 12 sm: 14 md: 16 lg: 20 xl: 24 2xl: 36
    */
-  size?: 'xxs' | 'xs' | 'sm' | 'ms' | 'md' | 'lg' | 'xl' | '2xl';
+  size?: 'xxs' | 'xs' | 'xsm' | 'sm' | 'md' | 'ms' | 'lg' | 'xl' | '2xl';
```
tglaz, line 27: Think I understand the why, but let's not change Text Size options for a single ticket as it has global implications. Product/UX needs to address this globally, and there's a reason we have only those sizes.

Otherwise, if it's very custom, then can just use `<Text>` and not `<TextBold>`.

Also, don't both of these indicate `extra small`?:  `xsm` and `xs`
And `ms` is just weird.

libs/expo/shared/ui-components/src/lib/TextMedium/TextMedium.tsx
```diff
@@ -24,7 +24,7 @@ interface ITextMediumProps extends TextProps {
    * size:
    * - xxs: 10 xs: 12 sm: 14 md: 16 lg: 20 xl: 24 2xl: 36
    */
-  size?: 'xxs' | 'xs' | 'sm' | 'md' | 'lg' | 'xl' | '2xl';
+  size?: 'xxs' | 'xs' | 'xsm' | 'sm' | 'md' | 'ms' | 'lg' | 'xl' | '2xl';
```
tglaz, line 27: same comments as for `TextBold`

libs/expo/shared/ui-components/src/lib/TextRegular/TextRegular.tsx
```diff
@@ -24,7 +24,7 @@ interface ITextRegularProps extends TextProps {
    * size:
    * - xxs: 10 xs: 12 sm: 14 md: 16 lg: 20 xl: 24 2xl: 36
    */
-  size?: 'xxs' | 'xs' | 'sm' | 'md' | 'lg' | 'xl' | '2xl';
+  size?: 'xxs' | 'xs' | 'xsm' | 'sm' | 'md' | 'ms' | 'lg' | 'xl' | '2xl';
```
tglaz, line 27: same comments as for `TextBold`

tglaz (changes requested):

libs/expo/betterangels/src/lib/ui-components/NoteCard/NoteCardByline.tsx
mikefeldberg: I was following the pattern from another component, but I agree. I think it would make more sense for this to be "author profile photo", since the name will be displayed and captured by screen readers. And the email isn't displayed anywhere so it doesn't make sense to suddenly include in accessibility

libs/expo/betterangels/src/lib/ui-components/NoteCard/NoteCardClient.tsx
mikefeldberg: I'm not sure why we require `accessibilityHint`, since the avatar isn't interactive. What do you think of `accessibilityLabel={'interaction author photo'}` is fine for both?

libs/expo/shared/ui-components/src/lib/TextBold/TextBold.tsx
mikefeldberg: What are the global implications of adding these sizes as acceptable values? Happy to use `<Text>`, I just actually don't know the potential downstream effects of this change. wrt naming, I was using the labels in our [fontSizes](https://github.com/BetterAngelsLA/monorepo/blob/main/libs/expo/shared/static/src/lib/fontSizes.ts) file

tglaz: my bad. didn't realize we actually had them declared.

libs/expo/betterangels/src/lib/ui-components/NoteCard/NoteCardClient.tsx
tglaz: unless there's a good reason to be more specific, that'd work for me

libs/expo/betterangels/src/lib/ui-components/NoteCard/NoteCardPills.tsx
mikefeldberg: afaict, they have different homes and serve unique functions. `PillContainer` is used to display services in the note [summary](https://github.com/BetterAngelsLA/monorepo/blob/main/libs/expo/betterangels/src/lib/screens/Note/NoteServices.tsx); it truncates results behind an interactive collapsible that can be expanded to view the full list. This lives in `NoteCard` and has no interactivity.

tglaz: PillContainer should just render a group of Pills, regardless of location. It may have different ways to display them, but that would just be a display option. There'll probably ever be 1 or 2 ways to render them, apart from rendering all.

libs/expo/shared/ui-components/src/lib/PillContainer/PillContainer.tsx
```diff
@@ -1,45 +1,136 @@
...
+  services,
+  serviceType,
   variant,
-  maxVisible,
 }: {
-  data: string[];
-  variant: 'primary' | 'success' | 'warning';
-  maxVisible: number;
+  services: string[];
+  serviceType: 'providedServices' | 'requestedServices';
```
tglaz, line 15: I think we can just accept pill `variant` prop and pass it on to <Pill>.

libs/expo/shared/ui-components/src/lib/PillContainer/PillContainer.tsx
```diff
@@ -1,45 +1,136 @@
...
+  variant: 'singleRow' | 'expandable';
 }) {
+  const maxVisiblePills = 5;
```
tglaz, line 18: Let's keep `maxVisible` as a prop and just set a default.

libs/expo/shared/ui-components/src/lib/PillContainer/PillContainer.tsx
```diff
@@ -1,45 +1,136 @@
...
-  return (
-    <View>
-      <View style={styles.pillContainer}>
-        {servicesToDisplay.map((item, idx) => (
-          <Pill variant={variant} label={item} key={idx} />
+      setVisibleCount(count);
+    }
+  }, [containerWidth, pillWidths, services.length]);
+
+  if (variant === 'singleRow') {
```
tglaz, line 54: may help to split this out into its own component, otherwise we're always managing a lot of state even if unused.

```typescript
if (variant === 'singleRow') {
   return <SinglePillRow pills={pills} pillVariant={.....   />
}

... other variants here
```

libs/expo/shared/ui-components/src/lib/PillContainer/PillContainer.tsx
```diff
@@ -1,45 +1,122 @@
...
+  variant: 'singleRow' | 'expandable';
 }) {
   const [showAll, setShowAll] = useState(false);
-  const servicesToDisplay = showAll ? data : data.slice(0, maxVisible);
+  const pillsToDisplay = showAll ? pills : pills.slice(0, maxVisible);
+  const [containerWidth, setContainerWidth] = useState(0);
+  const [pillWidths, setPillWidths] = useState<number[]>(
+    Array(pills.length).fill(0)
+  );
+  const [visibleCount, setVisibleCount] = useState(pills.length);
```
tglaz, line 26: perhaps this is unnecessary, as can just reference `pillsToDisplay.length`?

libs/expo/shared/ui-components/src/lib/PillContainer/ExpandablePillRow.tsx
```diff
@@ -0,0 +1,59 @@
...
+import TextBold from '../TextBold';
+
+export function ExpandablePillRow({
+  maxVisible = 5,
+  pills,
+  pillVariant,
+}: {
+  maxVisible?: number;
+  pills: string[];
+  pillVariant: 'primary' | 'success' | 'warning';
```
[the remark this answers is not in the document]
tglaz: can update `Pill` to something like:

```typescript
export const VARIANTS = {
  primary: {
    bg: Colors.PRIMARY_EXTRA_LIGHT,
    border: Colors.PRIMARY_DARK,
  },
  success: {
    bg: Colors.SUCCESS_EXTRA_LIGHT,
    border: Colors.SUCCESS,
  },
  warning: {
    bg: Colors.WARNING_EXTRA_LIGHT,
    border: Colors.WARNING,
  },
} as const;

export type TVariants = typeof VARIANTS;
export type TVariant = keyof TVariants;

interface IPillProps {
  variant?: TVariant;
  label: string;
}

...
```

PR #1347, opened 2025-06-03
fix pill row overflow (DEV-1869)

mikefeldberg:
DEV-1869

## Summary by Sourcery

Refactor SinglePillRow to improve width measurement and overflow handling by caching dimensions in refs, debouncing layout calculations with requestAnimationFrame, and ensuring accurate visible pill counts.

Bug Fixes:
- Fix pill row overflow by recalculating visibility based on actual measured widths

Enhancements:
- Refactor SinglePillRow to use React.memo and useRef for container and pill widths
- Replace state-based width tracking with ref-based storage and recalculate visibility on layout events
- Debounce layout measurements using requestAnimationFrame to reduce unnecessary renders
- Key pill elements by label instead of index to ensure consistent rendering

### DIFF
--- a/libs/expo/betterangels/src/lib/apollo/graphql/__generated__/queries.generated.ts
+++ b/libs/expo/betterangels/src/lib/apollo/graphql/__generated__/queries.generated.ts
@@ -10,14 +10,14 @@ export type NotesQueryVariables = Types.Exact<{
 }>;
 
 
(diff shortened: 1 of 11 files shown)


libs/expo/shared/ui-components/src/lib/PillContainer/SinglePillRow.tsx
```diff
@@ -59,7 +68,7 @@ export function SinglePillRow({
       {pills.slice(0, visibleCount).map((item, idx) => (
         <View
           key={idx}
-          style={{ marginLeft: idx === 0 ? 0 : Spacings.xs }}
+          style={{ marginLeft: idx === 0 ? 0 : Spacings.xxs }}
```
tglaz, line 71: this could just be a `gap` prop at the component level, and include it i the container <View> styles

tglaz (approved):

PR #1362, opened 2025-06-09
Bump python pacakges

vecchp:
## Summary by Sourcery

Bump Python package versions across both production and development dependencies, updating the poetry manifests and lockfile to the latest compatible releases.

Enhancements:
- Upgrade production dependencies (e.g., Django, psycopg, Celery, Strawberry GraphQL, Hiredis, and others) to their latest patch/minor versions.
- Upgrade development dependencies (e.g., flake8, django-stubs, boto3-stubs, ipython, type stubs, and testing tools) to their latest versions.

### DIFF
--- a/apps/betterangels-backend/accounts/backends.py
+++ b/apps/betterangels-backend/accounts/backends.py
@@ -1,9 +1,11 @@
 import uuid
-from typing import Any, Optional, cast
+from typing import Any, Optional
 
 from django.contrib.auth.models import AbstractBaseUser
+from django.core.mail import EmailMultiAlternatives
 from django.db import transaction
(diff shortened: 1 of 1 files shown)


apps/betterangels-backend/clients/models.py
```diff
@@ -175,7 +176,7 @@ def display_pronouns(self) -> Optional[str]:
         if self.pronouns == PronounEnum.OTHER:
             return self.pronouns_other
 
-        return self.pronouns.label
+        return force_str(self.pronouns.label)
```
mikefeldberg, line 179: why `force_str` vs `str`?

PR #1367, opened 2025-06-10
shelter directions (DEV-1862)

mikefeldberg:
DEV-1862

## Summary by Sourcery

Enable directions functionality by updating Actions to accept shelter location and launch Google Maps directions

New Features:
- Add Directions button to open Google Maps with destination coordinates

Enhancements:
- Define TShelterLocation type and accept optional location prop in Actions
- Pass shelter location from ShelterPage to Actions

Chores:
- Log location prop for debugging


Davit-BetterAngels (changes requested):

kimchoi-jjiggae:
Neither are launch blocking, let me know if you prefer to address now or for me to file tickets for a later milestone (clean up)
1. Destination should auto-populate as the shelter name rather than "dropped pin" (current implementation) - this makes it easier for the user to remember what the shelter is without having to revisit the app
2. clicking directions opens up a new empty tab. Slightly confusing since when you go back to your browser from the map, you have to close this page or find the shelter page within your tab. Ideal behavior is that map opens without opening a new empty tab

shiozakiyuko:
Hi @Davit-BetterAngels 
I have to add something to @kimchoi-jjiggae 's comment.

What she is noticing is happening in iOS. Android is fine that is showing the address of the destination instead of "Dropped Pin" and there is not blank page after you tapped the Direction button.

libs/react/shelter/src/lib/pages/shelter/index.tsx
```diff
@@ -85,7 +85,10 @@ export default function ShelterPage({ id }: { id: string }) {
       >
         See all photos
       </Button>
-      <Actions />
+      <Actions
+        location={shelter.location ? shelter.location : null}
```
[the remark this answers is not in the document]
mikefeldberg: 😳🤦

libs/react/shelter/src/lib/pages/shelter/Actions.tsx
```diff
@@ -4,12 +4,41 @@ import {
...
   phone?: string | null;
+  shelterName: string;
 };
 
-export default function Actions({ phone, shelterName }: TProps) {
+const ua = navigator.userAgent;
+const isIOS = /iPad|iPhone|iPod/i.test(ua);
+
+function openDirections(place?: string) {
+  if (place == null) {
```
tglaz, line 23: technically it could be `undefined` also

libs/react/shelter/src/lib/pages/shelter/Actions.tsx
```diff
@@ -4,12 +4,41 @@ import {
...
+    return;
+  }
+
+  let url: string;
+  let destination: string;
```
tglaz, line 28: think `destination` can probably be a `const` as it's same for both

tglaz (changes requested):

libs/react/shelter/src/lib/pages/shelter/Actions.tsx
```diff
@@ -4,12 +4,41 @@ import {
...
+  location?: TShelterLocation | null;
   phone?: string | null;
+  shelterName: string;
 };
 
-export default function Actions({ phone, shelterName }: TProps) {
+const ua = navigator.userAgent;
+const isIOS = /iPad|iPhone|iPod/i.test(ua);
+
+function openDirections(place?: string) {
```
tglaz, line 22: We'd probably want this to be reusable, so can go into `libs/react/components`.
Can name it something like `OpenInMaps` or something.
And should accept `className` to be fully customizable as a button.

libs/react/shelter/src/lib/pages/shelter/Actions.tsx
```diff
@@ -4,12 +4,41 @@ import {
...
+
+  let url: string;
+  let destination: string;
+
+  if (isIOS) {
+    destination = encodeURIComponent(place);
+    url = `maps://?q=${destination}`;
+  } else {
+    destination = encodeURIComponent(place);
+    url = `https://www.google.com/maps/dir/?api=1&destination=${destination}`;
```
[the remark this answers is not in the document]
mikefeldberg: `geo` is android specific

libs/react/shelter/src/lib/pages/shelter/Actions.tsx
mikefeldberg: the first draft of this used `lag + lng`. It didn't work as expected. The coords would get converted to an address and didn't always match the shelter's address. e.g., a shelter on a corner/intersection may be given the nearest cross-street address rather than the actual address. 

Address seems to be the safer approach to go for now. We should probably start storing `place_id` (returned by the Places API but we don't use it). It wouldn't help with apple maps, but it would guarantee that we always had the right place in google maps.

tglaz (approved):

PR #1368, opened 2025-06-10
share shelter (DEV-1863)

mikefeldberg:
DEV-1863

## Summary by Sourcery

Enable users to share the current shelter via native sharing or by copying the link, by extending the Actions component with a new share handler

New Features:
- Add share functionality to the shelter Actions component using the Web Share API with a clipboard fallback

Enhancements:
- Convert the Actions component to accept a shelterName prop and update its usage to pass the shelter name from ShelterPage


kimchoi-jjiggae:
Please remove the colon after the shelter name, since nothing is after it! This is not launch blocking, so I can either approve it and file a separate ticket, or if it's easier to fix now then we can address it now

shiozakiyuko:
Hi @mikefeldberg 

You missed the pressed state here too. :)

libs/react/shelter/src/lib/pages/shelter/Actions.tsx
```diff
@@ -1,6 +1,35 @@
...
-export default function Actions() {
+type TProps = {
+  shelterName: string;
+};
+
+export default function Actions({ shelterName }: TProps) {
+  const handleShare = async () => {
+    const shareData = {
+      title: shelterName,
+      text: `${shelterName} `,
```
tglaz, line 11: if title and text (description) are the same, then can probably omit the text

mikefeldberg: | `title` is what shows up in the share dialogue  | `text` is what gets copied with the URL |
|--------|--------|
|  | image |

libs/react/shelter/src/lib/pages/shelter/Actions.tsx
```diff
@@ -1,6 +1,35 @@
 import { CallIcon, LocationIcon, ShareIcon } from '@monorepo/react/icons';
```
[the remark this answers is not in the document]
shiozakiyuko: I know @Davit-BetterAngels knows how.

PR #1369, opened 2025-06-10
make email and phone number clickable (DEV-1537, DEV-1542)

Davit-BetterAngels:
DEV-1537
[DEV-1542](https://betterangels.atlassian.net/browse/DEV-1542)

[DEV-1542]: https://betterangels.atlassian.net/browse/DEV-1542?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libs/expo/shared/ui-components/src/lib/EmailBtn/EmailBtn.tsx
+++ b/libs/expo/shared/ui-components/src/lib/EmailBtn/EmailBtn.tsx
@@ -1,5 +1,6 @@
 import { Colors } from '@monorepo/expo/shared/static';
-import { Linking, Pressable, Text } from 'react-native';
+import { Linking, Pressable } from 'react-native';
+import TextBold from '../TextBold';
 
 interface IEmailBtnProps {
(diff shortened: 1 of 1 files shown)


libs/expo/shared/ui-components/src/lib/PhoneNumberBtn/PhoneNumberBtn.tsx
```diff
@@ -0,0 +1,30 @@
+import { Colors } from '@monorepo/expo/shared/static';
+import { Linking, Pressable } from 'react-native';
+import TextBold from '../TextBold';
+
+interface IPhoneNumberBtnProps {
+  text: string;
```
tglaz, line 6: may need to accept other props, like size, color etc if we use it elsewhere, but can customize later I suppose
same for `EmailBtn`

PR #1370, opened 2025-06-11
edit file name (DEV-1561)

mikefeldberg:
DEV-1561

## Summary by Sourcery

Enable in-place editing of client document filenames in the FileScreenComponent

New Features:
- Add an editing mode flag and BasicInput field to FileScreenComponent to allow users to modify file names.
- Introduce an UpdateClientDocument GraphQL mutation to persist filename changes and refetch the client profile.

Enhancements:
- Display bottom action buttons (Submit/Cancel) and show snackbar notifications for update success or errors.
- Update DocumentModal to include an “Edit file name” option navigating to the editing interface.

Chores:
- Propagate clientId and editing parameters through the file screen route and component props.


libs/expo/betterangels/src/lib/ui-components/DocumentModal.tsx
```diff
@@ -111,6 +111,11 @@ export default function DocumentModal(props: IDocumentModalProps) {
       Icon: ViewIcon,
       route: `/file/${document?.id}`,
     },
+    {
+      title: `Edit filename`,
+      Icon: ViewIcon,
```
mikefeldberg, line 116: [waiting](https://betterangels-group.slack.com/archives/C05ECN9FLGL/p1749601083001939?thread_ts=1749601027.500239&cid=C05ECN9FLGL) for direction from product

libs/expo/betterangels/src/lib/screens/FileScreenComponent/index.tsx
```diff
@@ -111,17 +162,48 @@ export default function FileScreenComponent(props: TFileScreenComponent) {
             />
           )}
 
-          <TextBold mt="sm" size="sm">
-            File Name
-          </TextBold>
-          <TextRegular size="sm" style={{ width: '100%' }}>
-            {originalFilename}
-          </TextRegular>
+          {editing === 'true' ? (
```
mikefeldberg, line 165: I don't love this str comparison but wasn't sure if there's a better standard approach. I considered using a variant but that would still result in a str comparison so 🤷

tglaz (approved):

apps/betterangels/src/app/(private-screens)/file/[id].tsx
```diff
@@ -2,13 +2,15 @@ import { FileScreenComponent } from '@monorepo/expo/betterangels';
 import { useLocalSearchParams } from 'expo-router';
```
Comment (shiozakiyuko) [inline apps/betterangels/src/app/(private-screens)/file/[id].tsx:2]: Hi @mikefeldberg 

The only issue I had is the textbox wasn't visible while I'm typing with the keyboard (it was hidden behind the keyboard).

When the keyboard is visible, the textbox should be pushed up so the user can see while/what they are typing. 
Thanks!

https://github.com/user-attachments/assets/385792bc-801b-4476-b283-4b64ba85bed6

shiozakiyuko (commented):
Hi @mikefeldberg 

So.. I confirmed it's fixed on iOS. Looks great so thank you! But on Android side, it doesn't load... 
I did get the latest build first and scanned the QR code afterwards. I tried to scan twice but I get the same result: black screen (see the attachment)

kimchoi-jjiggae (approved):

PR #1377, opened 2025-06-12
fix author filter name display (DEV-1563)

mikefeldberg:
DEV-1563

## Summary by Sourcery

Bug Fixes:
- Omit trailing space and only include last name if present in the AuthorsFilter label


libs/expo/betterangels/src/lib/screens/Interactions/InteractionsFilters/AuthorsFilter/index.tsx
```diff
@@ -106,7 +106,7 @@ export default function AuthorsFilter(props: IAuthorsFilterProps) {
       .filter((item) => item.id !== user?.id && item.firstName)
       .map((item) => ({
         id: item.id,
-        label: `${item.firstName} ${item.lastName}`,
+        label: `${item.firstName}${item.lastName ? ` ${item.lastName}` : ''}`,
```
tglaz, line 109: should that space be omitted?

and can probably omit the ternary
```javascript
label: `${item.firstName} ${item.lastName || ''}`
```

tglaz (changes requested):

PR #1389, opened 2025-06-19
Location Tab drawer behavior (DEV-1730)

Davit-BetterAngels:
DEV-1730

### DIFF
--- a/libs/expo/betterangels/src/lib/screens/Client/Locations/InteractionLocationsModal.tsx
+++ b/libs/expo/betterangels/src/lib/screens/Client/Locations/InteractionLocationsModal.tsx
@@ -1,8 +1,16 @@
-import { Colors, Spacings } from '@monorepo/expo/shared/static';
+import BottomSheet, { BottomSheetView } from '@gorhom/bottom-sheet';
+import {
+  Colors,
+  Radiuses,
+  Shadow,
+  Spacings,
+} from '@monorepo/expo/shared/static';
 import { TextBold } from '@monorepo/expo/shared/ui-components';
-import { View } from 'react-native';
+import { useMemo, useRef, useState } from 'react';
+import { StyleSheet, View } from 'react-native';
+import { useSafeAreaInsets } from 'react-native-safe-area-context';
 import { NotesQuery } from '../../../apollo';
-import { Modal, NoteCard } from '../../../ui-components';
+import { NoteCard } from '../../../ui-components';
 
 interface InteractionLocationsModalProps {
   selectedLocation: NotesQuery['notes']['results'][number] | null;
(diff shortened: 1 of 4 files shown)


libs/expo/betterangels/src/lib/screens/Client/Locations/InteractionLocationsModal.tsx
```diff
@@ -15,43 +23,72 @@ export function InteractionLocationsModal(
   props: InteractionLocationsModalProps
 ) {
   const { selectedLocation, setSelectedLocation } = props;
+  const [titleHeight, setTitleHeight] = useState<number>(1);
```
tglaz, line 26: It it possible to make it a lib component we can reuse?
Perhaps it could accept props such as header + content etc....

tglaz (changes requested):

libs/expo/betterangels/src/lib/screens/Client/Locations/InteractionLocationsModal.tsx
```diff
@@ -15,43 +23,72 @@ export function InteractionLocationsModal(
...
+        backgroundColor: Colors.NEUTRAL_LIGHT,
+        width: 54,
+        height: 5,
+        borderRadius: Radiuses.xxs,
+      }}
+      snapPoints={snapPoints}
+      enablePanDownToClose={false}
+      ref={bottomSheetRef}
+      enableDynamicSizing
+      style={styles.shadow}
```
tglaz, line 51: I see this warning in the logs a lot, which we may want to address. Though not sure if we'd need to wrap it in a View or handle in another way or choose to ignore.
```
 WARN  (ADVICE) View #1689 of type RCTView has a shadow set but cannot calculate shadow efficiently. Consider setting a solid background color to fix this, or apply the shadow to a more specific component.
 ```

I saw it with the ClusterIcon as well and removed it for now, though mostly cause we can render a ton of markers and that many shadows can get be expensive, especially when React complains.

**Comment by kimchoi-jjiggae** [comment]
1. Can you make the lower padding for the android version on the drawer header clearer (right underneath the address)? right now it's so small that it's easy to miss the drawer header entirely. You can just match to ios for now

Non-Blocking, only include if trivial to implement. Otherwise I can file new tickets:
2. When you click on an interaction in the drawer and then go back, the drawer is not visible at all. Could you keep the drawer open? 
3. I believe the swiped up version of the drawer is supposed to be taller, but we can wait to fix that until @shiozakiyuko returns and gives us an exact spec. As long as it is clear in the edge case (1844) where there are multiple cards, this should be fine for now

File: libs/expo/shared/ui-components/src/lib/BottomSheetModal/BottomSheetModal.tsx [hunk 3]
```diff
@@ -0,0 +1,57 @@
...
+  const bottomSheetRef = useRef<BottomSheet>(null);
+
+  const snapPoints = useMemo(() => [titleHeight], [titleHeight]);
+
+  return (
+    <BottomSheet
+      {...rest}
+      index={0}
+      handleIndicatorStyle={{
+        backgroundColor: Colors.NEUTRAL_LIGHT,
```
tglaz, line 30: minor thing,
since already defining styles with `StyleSheet.create`, can move these styles there

libs/expo/shared/ui-components/src/lib/BottomSheetModal/Content.tsx
```diff
@@ -0,0 +1,60 @@
...
+import TextBold from '../TextBold';
+
+export default function BottomSheetModalContent({
+  children,
+  setTitleHeight,
+  title,
+}: {
+  children: React.ReactNode;
+  style?: React.CSSProperties;
+  setTitleHeight: (height: number) => void;
```
tglaz, line 14: to make it more generic, instead of setTitleHeight could pass an `onMeasuredLayout` function that takes (title) dimensions as an arg. Then the consumer can do whatever it wants with it, such as `setTitleHeight`.

```
type TDims = {width: num, height: num} ;

onMeasuredLayout: (dims: TDims) => void;
```

libs/expo/shared/ui-components/src/lib/BottomSheetModal/Content.tsx
```diff
@@ -0,0 +1,60 @@
...
+  title?: string | null;
+}) {
+  const insets = useSafeAreaInsets();
+  const bottomOffset = insets.bottom;
+  return (
+    <BottomSheetView style={styles.container}>
+      {title && (
```
tglaz, line 21: seems like w/o `title` there'd be nothing visible?

tglaz (changes requested):

tglaz (approved):

PR #1402, opened 2025-06-25
address direction in interaction summary (DEV-1727)

Davit-BetterAngels:
DEV-1727

## Summary by Sourcery

Introduce address-based direction selection by adding a DirectionsPopup component and integrating it into NoteLocation to allow users to launch Apple or Google Maps from an address link.

New Features:
- Make NoteLocation address clickable to open a directions modal
- Add a shared DirectionsPopup component offering Apple and Google Maps routing options

### DIFF
--- a/libs/expo/shared/ui-components/src/lib/DirectionsPopup/DirectionsPopup.tsx
+++ b/libs/expo/shared/ui-components/src/lib/DirectionsPopup/DirectionsPopup.tsx
@@ -1,6 +1,7 @@
 import { Colors, Spacings } from '@monorepo/expo/shared/static';
 
-import { Modal, Pressable, StyleSheet, View } from 'react-native';
+import { useState } from 'react';
+import { Modal, Platform, Pressable, StyleSheet, View } from 'react-native';
 import openMap from 'react-native-open-maps';
 import TextBold from '../TextBold';
 import TextRegular from '../TextRegular';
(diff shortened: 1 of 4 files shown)


kimchoi-jjiggae (approved):

libs/expo/shared/ui-components/src/lib/DirectionsPopup/DirectionsPopup.tsx
```diff
@@ -0,0 +1,154 @@
...
+
+  if (!address || !address.street) {
+    return null;
+  }
+
+  return (
+    <>
+      <TextRegular
+        onPress={() => setChooseDirections(true)}
+        textDecorationLine="underline"
```
tglaz, line 43: May be better to pass in styles such as `mt` and `textDecorationLine` as props.
Alternatively `DirectionsPopup` could just accept `address` prop + children. And DirectionsPopup would just care about a click event, assuming address exists.

But nice abstraction.

libs/expo/shared/ui-components/src/lib/DirectionsPopup/DirectionsPopup.tsx
```diff
@@ -0,0 +1,154 @@
...
+    address?.city,
+    address?.state,
+    address?.zipCode,
+  ]
+    .filter(Boolean)
+    .join(', ');
+
+  const isDisabled = !address?.street;
+
+  if (!address || !address.street) {
```
tglaz, line 35: this looks same as `isDisabled`, so using it below probably not useful

PR #1438, opened 2025-07-08
handle missing extension (DEV-1952)

mikefeldberg:
DEV-1952

## Summary by Sourcery

Bug Fixes:
- Conditionally append the extension to the dial URL to avoid trailing commas when no extension is provided.

### DIFF
--- a/libs/expo/shared/ui-components/src/lib/PhoneNumberBtn/PhoneNumberBtn.tsx
+++ b/libs/expo/shared/ui-components/src/lib/PhoneNumberBtn/PhoneNumberBtn.tsx
@@ -10,14 +10,17 @@ interface IPhoneNumberBtnProps {
 export function PhoneNumberBtn(props: IPhoneNumberBtnProps) {
   const { number, label } = props;
   const [phoneNumber, extension] = number;
+
+  const phoneNumberUrl = extension
+    ? `${phoneNumber},${extension}`
+    : phoneNumber;
+
   return (
     <Pressable
       accessibilityRole="button"
       accessibilityHint="Opens your phone dialer to call the number"
-      accessibilityLabel={`Call ${phoneNumber}`}
-      onPress={() =>
-        Linking.openURL(`tel:${phoneNumber}${extension ? `,${extension}` : ''}`)
-      }
+      accessibilityLabel={`Call ${phoneNumberUrl}`}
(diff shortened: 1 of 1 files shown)


libs/expo/shared/ui-components/src/lib/PhoneNumberBtn/PhoneNumberBtn.tsx
```diff
@@ -15,7 +15,9 @@ export function PhoneNumberBtn(props: IPhoneNumberBtnProps) {
       accessibilityRole="button"
       accessibilityHint="Opens your phone dialer to call the number"
       accessibilityLabel={`Call ${phoneNumber}`}
-      onPress={() => Linking.openURL(`tel:${phoneNumber},${extension}`)}
+      onPress={() =>
+        Linking.openURL(`tel:${phoneNumber}${extension ? `,${extension}` : ''}`)
```
tglaz, line 19: I wouldn't mind breaking out the phone URL string logic

```javascript
const phoneURLNumber = extension
    ? `${phoneNumber},${extension}`
    : phoneNumber;

// and then
onPress={() => Linking.openURL(`tel:${phoneURLNumber}`)}
```

and maybe a line break after line 12 :)

tglaz: also technically `phoneNumber` can be undefined given this type from IPhoneNumberBtnProps:

```number: Array<string | undefined>;```

Comment (mikefeldberg) [reply libs/expo/shared/ui-components/src/lib/PhoneNumberBtn/PhoneNumberBtn.tsx]: added `phoneNumberUrl`. This component is only displayed when there's a number to display, so I don't want to address that here

File: libs/expo/shared/ui-components/src/lib/PhoneNumberBtn/PhoneNumberBtn.tsx [hunk 2]
```diff
@@ -10,12 +10,17 @@ interface IPhoneNumberBtnProps {
 export function PhoneNumberBtn(props: IPhoneNumberBtnProps) {
   const { number, label } = props;
   const [phoneNumber, extension] = number;
+
+  const phoneNumberUrl = extension
```
tglaz, line 14: since `phoneNumber` could be undefined, should probably add

```javascript
if (!phoneNumber) {
   return null;
}

```

PR #1457, opened 2025-07-16
Modal provider services (DEV-1009)

vecchp:


shiozakiyuko (approved):
Hi @aaronipollock!

I got some clarification from @vecchp that you are splitting your DEV-1009 to some parts. The previous comment about not seeing Done button in Interaction/Location - Search box should be disregarded in this PR. However, I am not sure which PR I should attach the comment to so please let me know with the right PR.

So to follow the title, I confirmed that the modal looks good in the Provided/Requested Services view. Thanks!

PR #1459, opened 2025-07-17
admin portal login (DEV-1903)

Davit-BetterAngels:
## Summary by Sourcery

Implement comprehensive admin portal authentication and routing infrastructure including OTP and password login, CSRF handling, and user context providers

New Features:
- Add SignIn component supporting email validation, OTP and password flows, and dynamic demo/production environment switching
- Introduce ApiConfigProvider for selecting API environment and performing CSRF-protected fetch requests
- Implement UserProvider to fetch and manage current user data and AuthProvider to enforce route access and redirects
- Configure ApolloClient with custom CSRF ApolloLink, RestLink, and uploadLink for GraphQL and REST integration

Enhancements:
- Restructure application routing to use a dynamic routeChildren array and remove generated NxWelcome routes
- Add cookie utility functions for extracting, deduplicating, and erasing CSRF tokens across subdomains

Chores:
- Add "admin" serve script in package.json
- Export hooks, pages, and providers from the library entry point


apps/betterangels-admin/src/app/clients/apollo/client.tsx
vecchp, line 1: This is effectively the same code from the shelter app so i think this should be moved to a lib that both can use.

apps/betterangels-admin/src/app/clients/apollo/csrf.ts
vecchp, line 1: This is the same code used in the shelter app so we should pull this out into a shared lib now given it is the same functionality across two apps.

apps/betterangels-admin/src/main.tsx
```diff
@@ -1,18 +1,38 @@
...
 import * as ReactDOM from 'react-dom/client';
 import { BrowserRouter } from 'react-router-dom';
+import { apiUrl, demoApiUrl } from '../config';
 import App from './app/app';
+import { createApolloClient } from './app/clients/apollo/client';
 
 const basename = import.meta.env.VITE_APP_BASE_PATH || '/';
 
+const apolloClient = createApolloClient({
+  apiUrl: import.meta.env.VITE_SHELTER_API_URL,
```
vecchp, line 17: `VITE_SHELTER_API_URL` I don't think this is the Shelter app

libs/react/betterangels-admin/src/lib/providers/api/ApiConfigProvider.tsx
vecchp, line 1: I don't think we need advanced environment switching here. In the mobile app, it was necessary because we didn't have separate versions to log into. But now that we’ll have distinct local, dev, and production pages, simpler environment handling should be sufficient.

Also, I don't think this approach would work properly in the browser anyway; we’re likely to run into CORS issues with the backend.

Is there something I'm missing?

package.json
```diff
@@ -10,7 +10,8 @@
     "ba": "nx start betterangels",
     "ba:build-dev": "nx build betterangels --profile development",
     "shelter": "nx serve shelter-web --port 8083",
-    "fires": "nx serve wildfires --port 8200"
+    "fires": "nx serve wildfires --port 8200",
+    "admin": "nx serve betterangels-admin --port 8083"
```
vecchp, line 14: I think we probably want to run this on 8084 instead of 8083

apps/betterangels-admin/src/app/clients/apollo/csrf.ts
```diff
@@ -0,0 +1,87 @@
+import { ApolloLink, Observable } from '@apollo/client';
+import { eraseCookie, getCookie } from '../../utils/storage/cookies';
+
+const csrfCookieName =
```
vecchp, line 4: We can make the constants optionally overridable to the fetchClient and appolloClient after we move to a shared library, rather than relying on environment overrides in a constants file.

apps/betterangels-admin/src/app/utils/storage/cookies.ts
vecchp, line 1: I think this is also duplicate across shelter and admin now which we should move into a shared lib.

vecchp (changes requested):
Good start! Here are a few comments

libs/react/betterangels-admin/src/lib/providers/user/UserProvider.tsx
```diff
@@ -0,0 +1,76 @@
...
+
+  const refetchUser = useCallback(async () => {
+    setIsLoading(true);
+    try {
+      const res = await refetch();
+      updateUser(res);
+      setIsLoading(false);
+    } catch (err) {
+      console.error('Error refetching user data:', err);
+      setUser(undefined);
```
tglaz, line 64: should this need `setIsLoading(false)` here, or with a `finally()`?

Davit-BetterAngels: updated with finally

tglaz (approved):

vecchp (approved):

PR #1481, opened 2025-07-30
Persist email on login

vecchp:


libs/expo/betterangels/src/lib/ui-components/LoginForm/index.tsx
```diff
@@ -5,12 +5,13 @@ import {
...
   Loading,
 } from '@monorepo/expo/shared/ui-components';
+import { useAsyncStorageState } from '@monorepo/expo/shared/utils';
 import { useCallback, useEffect, useState } from 'react';
 import { StyleSheet, Text, View } from 'react-native';
 import useUser from '../../hooks/user/useUser';
 
 export default function LoginForm() {
-  const [email, setEmail] = useState('');
+  const [email, setEmail] = useAsyncStorageState('userEmail', '');
```
tglaz, line 14: not sure how much we care, but typically there's a checkbox to ask user if they want their username/email persisted on the screen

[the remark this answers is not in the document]
tglaz: think this is valid. we're changing to an async state, which i'm not sure is necessary.

I'd think we can continue storing email in state as before.
And use asyncStore to:
1. persist email upon successful login
2. pre-populate email state when opening the form

vecchp: wouldn't an opt in to this defeat the purpose of what we discussed for debugging the login issues?

libs/expo/betterangels/src/lib/ui-components/LoginForm/index.tsx
```diff
@@ -157,21 +159,57 @@ export default function LoginForm() {
 
       {step === 'otp' && (
         <>
+          <BasicInput
```
tglaz, line 162: since now Email field is always present, could remove the duplication and  just move it to very top

```javascript

  function onEmailChange(email: string) {
    setEmail(email);

    if (step === 'otp') {
      setOtp('');
      setStep('initial');
    }
  }
  
  
return (
    <View style={styles.container}>
      <BasicInput
        label="Email"
        value={email}
        onChangeText={onEmailChange}
        ...
```

PR #1487, opened 2025-07-31
admin add useRequiredUser hook and enforce email and org (DEV-1906)

tglaz:
* add `useRequiredUser` hook
* throw if no user.email or no user.organizations[0]
* update `TUser` with required values for:
  * email: string;
  * organization: TOrganization;


tglaz:
closing. may bring back later under own ticket

PR #1497, opened 2025-08-06
create and implement last seen marker (DEV-81, DEV-1896)

Davit-BetterAngels:
DEV-81
DEV-1896


libs/expo/shared/ui-components/src/lib/Map/mapUi/LocationMarker.tsx
```diff
@@ -0,0 +1,43 @@
...
+            }}
+          >
+            <TextBold size="sm">{label}</TextBold>
+          </View>
+        </View>
+      )}
+      <LocationPinIcon width={25} height={36} />
+    </View>
+  );
+}
```
tglaz, line 43: Probably good idea to memoize it, either here, the way `MapClusterMarker` does it, or elsewhere.

Since it only takes 1 prop, a string, it's probably simplest to return a memo version here.

PR #1546, opened 2025-08-28
Move Delete Profile button closer to Option Menu button

aaronipollock:
Setting StatusBar per Platform in _layout moves the Delete Profile button to the correct position underneath the Option Menu button, improving UX.


shiozakiyuko (commented):
Hi @aaronipollock and @tglaz 

I checked with the latest build but I didn't see any change.

image

shiozakiyuko:
Hi @tglaz and @aaronipollock 

As suggested, I waited for a while till the next morning and I rescanned the build code and the PR code but now I get the error message saying "There was a problem loading the project". I tried again by removing my app and rebuilt, and scanned the PR/QR code but same result. 

Please advice. Thanks.

PR #1579, opened 2025-09-17
display shelter hero (DEV-1764)

mikefeldberg:
DEV-1764

## Summary by Sourcery

New Features:
- Return the hero_image.file.url if a hero image is set in the hero_image resolver of ShelterType

### DIFF
--- a/apps/betterangels-backend/shelters/tests/test_queries.py
+++ b/apps/betterangels-backend/shelters/tests/test_queries.py
@@ -334,7 +334,7 @@ class ShelterQueryTestCase(GraphQLTestCaseMixin, ParametrizedTestCase, TestCase)
             }}
         """
 
-        expected_query_count = 24
+        expected_query_count = 28
 
         variables = {"order": {"name": "ASC"}}
 


libs/react/shelter/src/lib/pages/shelter/HeroCarousel.tsx
```diff
@@ -9,21 +9,24 @@ type TProps = {
...
 
   if (!images.length) {
     return <ImagePlaceholder className={mergeCss(placeholderCss)} />;
   }
 
+  const rest = images
+    .map((i) => i.file?.url)
+    .filter((u): u is string => !!u && u !== shelter.heroImage);
+
+  const imageUrls = shelter.heroImage ? [shelter.heroImage, ...rest] : rest;
```
mikefeldberg, line 28: this keeps typescript happy. The way the backend is designed, if the shelter has any images, `heroImage` can't be null, even if no hero image was explicitly set. But typescript doesn't know that, since `heroImage` optional

tglaz (approved):

PR #1580, opened 2025-09-17
enable task clients filter on fe (DEV-2155)

tglaz:
### Enable task clients filter on front-end
DEV-2155

## Summary by Sourcery

Re-enable the client profiles filter in the TaskFilters UI and update related types and utilities to use the pluralized clientProfiles field

New Features:
- Add the client profiles filter back to the TaskFilters component

Enhancements:
- Rename filter properties from clientProfile to clientProfiles in types, constants, and props
- Update toTaskFilterValue to handle an array of clientProfiles instead of a single clientProfile ID


shiozakiyuko (approved):
Looks good!

PR #1999, opened 2026-03-30
Update Shelter Model

nathanaela2002:
## Summary                                                         
  - Adds shelter_layout field to the Shelter model to support the new congregate/motel/mixed shelter type concept.                
  - New enum ShelterLayoutChoices with values Congregate, Motel/Single-Occupancy, and Mixed
  - Model field added as null=True, blank=True no backfill needed, existing shelters default to null
  - GraphQL: Exposed in CreateShelterInput (optional) and ShelterTypeMixin output                                         
  - Django Admin: added to ShelterForm and the "Summary Info" fieldset                                                        
  - Migration: Adds field to both shelter and shelterevent (pghistory audit table) and regenerates the three audit triggers


mikefeldberg:
this appears to be covered by [room_type](https://github.com/BetterAngelsLA/monorepo/blob/252661e3f3b9918de683700065000e2fbc57b0ef/apps/betterangels-backend/shelters/models/shelter.py#L220)

PR #2015, opened 2026-04-06
add endpoints for direct s3 uploads (DEV-2381)

tglaz:
### Direct to S3 file upload
DEV-2381

* adds endpoints to enable direct uploads to S3 from client (app/browser etc)
  * `generate_client_document_uploads` + `resolve_client_document_uploads`
  * `generate_client_profile_photo_upload` + `resolve_client_profile_photo_upload`
* updates docker-compose to use `minio` for local/ci s3 interactions
 * can inspect BE uploads storage in browser at `http://localhost:9001/`
* adds `STORAGES` config for local env (in settings.py)
* adds common/services/ 
  * s3
  * upload_token
* adds clients/services/ 
  * client_document
  * client_profile_photo
* adds common/storage.py with `LocalS3Storage` class for local env overrides
* extends `Attachment` class
  * `save` method now accepts `direct_upload` arg for working without an actual file
  

## Summary by Sourcery

Add support for direct S3-based uploads for client documents and profile photos, including local MinIO-backed storage for development and associated token-based validation and permissions.

New Features:
- Introduce GraphQL mutations and types to generate and resolve presigned S3 uploads for client documents and client profile photos.
- Add services for generating presigned S3 upload URLs and validating one-time upload tokens for client documents and profile photos.
- Provide a custom S3 storage backend and configuration to support local MinIO-based media storage in development.

Bug Fixes:
- Adjust tests comparing profile photo URLs to ignore presigned query parameters that vary between requests.
- Update scheduled report task tests to assert email template formatting via the mail-sending helper rather than persisted Email records.

Enhancements:
- Refactor attachment handling to support a direct_upload code path that infers attachment type from a provided MIME type and canonicalizes original filenames.
- Extract file utility helpers (filename canonicalisation, unique file path generation, attachment type inference) into a dedicated common.files.utils module.
- Add comprehensive unit tests for S3 upload URL generation, upload token lifecycle, direct upload handling, and client document/profile photo upload services.
- Update test storage configuration overrides to use the new STORAGES settings structure.

Build:
- Extend docker-compose to run a MinIO server and bucket-initialization sidecar for local S3-compatible storage.
- Load shared compose environment variables for local development and wire new S3-related settings into Django storage configuration.


apps/betterangels-backend/betterangels_backend/settings.py
```diff
@@ -297,6 +306,22 @@
             "BACKEND": "django.contrib.staticfiles.storage.StaticFilesStorage",
         },
     }
+elif IS_LOCAL_DEV and LOCAL_S3_INTERNAL_ENDPOINT_URL:
```
mikefeldberg: I think we need `and LOCAL_S3_PUBLIC_ENDPOINT_URL` here. As is, storages can get set to `LocalS3Storage`, and `LocalS3Storage.url()` (gated on `LOCAL_S3_PUBLIC_ENDPOINT_URL`) will call `S3Storage.url()`

apps/betterangels-backend/clients/services/client_document.py
```diff
@@ -0,0 +1,123 @@
...
+from common.models import Attachment
+from common.permissions.enums import AttachmentPermissions
+from common.permissions.utils import assign_object_permissions
+from common.services.s3 import DEFAULT_UPLOAD_EXPIRATION_SECONDS, generate_s3_presigned_upload_urls
+from common.services.upload_token import create_upload_token, validate_upload_token
+from django.contrib.contenttypes.models import ContentType
+
+STORAGE_DIR = "media"
+CLIENT_DOCUMENT_RELATIVE_PATH = "attachments"
+S3_CLIENT_DOCUMENT_PREFIX = f"{STORAGE_DIR}/{CLIENT_DOCUMENT_RELATIVE_PATH}"
```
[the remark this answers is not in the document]
mikefeldberg: +1 on https://github.com/BetterAngelsLA/monorepo/pull/2015#discussion_r3042183748

apps/betterangels-backend/clients/services/client_document.py
```diff
@@ -0,0 +1,128 @@
...
+SERVICE_NAME = "client_document"
+
+
+class AuthorizedPresignedUpload(TypedDict):
```
mikefeldberg: since this is defined in two places, maybe consider moving both somewhere like common.services?

apps/betterangels-backend/clients/services/client_profile_photo.py
```diff
@@ -0,0 +1,83 @@
...
+from typing import TypedDict
+
+from accounts.models import User
+from clients.models import ClientProfile
+from clients.types import GenerateClientProfilePhotoUploadInput
+from common.services.s3 import DEFAULT_UPLOAD_EXPIRATION_SECONDS, generate_s3_presigned_upload_urls
+from common.services.upload_token import create_upload_token, validate_upload_token
+
+STORAGE_DIR = "media"
+CLIENT_PROFILE_PHOTO_PATH = f"{STORAGE_DIR}/client_profile_photos"
```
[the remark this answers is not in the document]
mikefeldberg: same

apps/betterangels-backend/common/services/s3.py
```diff
@@ -0,0 +1,192 @@
...
+        {"Content-Type": content_type},
+        ["content-length-range", 0, max_file_size],
+    ]
+
+    if normalized_path:
+        conditions.append(["starts-with", "$key", f"{normalized_path}/"])
+
+    # TODO - ohter conditions?
+    # conditions.append(["starts-with", "$key", ""]) # may be too loose
+    # Optional: restrict content-type prefix?
```
mikefeldberg, line 129: may consider implementing an allowlist at some point, but doesn't seem urgent

apps/betterangels-backend/common/services/s3.py
```diff
@@ -0,0 +1,192 @@
...
+            → attachments/drivers_license_front/<uuid>.jpg
+
+        upload_path=""
+            → <uuid>.jpg
+    """
+    # TODO: derive extension
+    # add mimetypes.guess_extension to common/files?
+    # ext = mimetypes.guess_extension(content_type) or Path(filename).suffix
+
+    ext = Path(filename).suffix
```
mikefeldberg: any reason not to use the existing util?
```suggestion
    try:
        ext = Path(canonicalise_filename(content_type, filename)).suffix
    except ValueError:
        ext = Path(filename).suffix
```

apps/betterangels-backend/common/services/s3.py
```diff
@@ -0,0 +1,192 @@
...
+    normalized_path = _normalize_upload_path(upload_path)
+    key = _build_s3_key(
+        filename=filename,
+        content_type=content_type,
+        upload_path=normalized_path,
+    )
+
+    conditions = [
+        {"Content-Type": content_type},
+        ["content-length-range", 0, max_file_size],
```
mikefeldberg: I know it's common to set the lower bound at `0` but doesn't that allow uploading empty files?

apps/betterangels-backend/common/services/s3.py
```diff
@@ -0,0 +1,192 @@
...
+import boto3
+from botocore.client import Config
+from botocore.exceptions import ClientError
+from django.core.files.storage import default_storage
+from mypy_boto3_s3 import S3Client
+from storages.backends.s3 import S3Storage
+
+# generate_presigned_post
+DEFAULT_UPLOAD_EXPIRATION_SECONDS = 300
+DEFAULT_MAX_FILE_SIZE = 10_000_000
```
mikefeldberg, line 12: did you check this or just a really good guesser?
```sh
$ aws s3api list-objects-v2 --bucket production-us-west-2-betterangels-backend --query "sort_by(Contents, &Size)[-1:]"
[
    {
        ...
        "Size": 91784660,
        ...
    }
]
```

tglaz: picked an arbitrary number for now.
we'd probably want to set it at something and can validate same size on FE as well. but set it at service level (client_document, client_profile_photo...)?
any thoughts?

currently `generate_s3_presigned_upload_urls` accepts any `max_file_size` as it's a common s3 service, but the consumers don't pass it so we just use the default.
do we want to set global `MAX_FILE_SIZE` somewhere?

apps/betterangels-backend/common/services/s3.py
tglaz: yup. i'll set to 1, unless there's a reason to set it higher

apps/betterangels-backend/common/services/s3.py
tglaz: think we technically for HMIS? will check.
at minimum i'll add some validation that will go along with your comment below regarding 
`ext = Path(filename).suffix`

apps/betterangels-backend/clients/services/client_document.py
tglaz: yup, in `common/services/types.py`?

apps/betterangels-backend/common/services/s3.py
tglaz: `canonicalise_filename` does more than necessary in s3.py. `generate_s3_presigned_upload_urls` simply builds `uuid+ext` for every file, where we can derive `ext` like this:

```
ext = mimetypes.guess_extension(content_type) or Path(filename).suffix
```

But my question would be whether we should raise if `mimetypes.guess_extension(content_type)` returns `None`, like in `canonicalise_filename`, or if file has not extension to use as fallback.
It'd guess it's safe to do so in our case.

But Python mimetype list may be missing some newer mimetypes. It's an edge case obviously.
And S3 doesn't really care, at least I know it does not require a file extension.

apps/betterangels-backend/common/services/s3.py
mikefeldberg: may be a product question. e.g., what do our users need to store/what do want to allow them to store?

apps/betterangels-backend/clients/services/client_document.py
```diff
@@ -0,0 +1,150 @@
...
+        "image/png",
+        "image/tiff",
+        "image/webp",
+    }
+)
+
+
+def _validate_content_type(content_type: str) -> None:
+    if content_type not in ALLOWED_CONTENT_TYPES:
+        raise ValueError(f"Unsupported content_type: {content_type}.")
```
mikefeldberg: since this is used in a batch op, maybe also include the filename in the error? I don't expect everyone to be able to identify which of their attachments was problematic if all they have to go on is content type

apps/betterangels-backend/clients/services/client_profile_photo.py
```diff
@@ -0,0 +1,91 @@
...
+from clients.types import GenerateClientProfilePhotoUploadInput
+from common.services.s3 import DEFAULT_UPLOAD_EXPIRATION_SECONDS, generate_s3_presigned_upload_urls, s3_key_exists
+from common.services.types import AuthorizedPresignedUpload
+from common.services.upload_token import create_upload_token, validate_upload_token
+
+STORAGE_DIR = "media"
+CLIENT_PROFILE_PHOTO_PATH = f"{STORAGE_DIR}/client_profile_photos"
+SERVICE_NAME = "client_profile_photo"
+
+ALLOWED_CONTENT_TYPES = frozenset(
```
mikefeldberg: why is this image list more restrictive than in doc service? 
```py
        # apps/betterangels-backend/clients/services/client_document.py
       ...
        "image/bmp",  # older windows/systems
        "image/gif",
        "image/heic",
        "image/heif",
        "image/jpeg",
        "image/png",
        "image/tiff",
        "image/webp",
        ...
```

apps/betterangels-backend/betterangels_backend/settings.py
```diff
@@ -280,6 +284,11 @@
 USE_TZ = True
 
 # Storage Settings
+AWS_REGION = env("AWS_REGION")
+AWS_S3_STORAGE_BUCKET_NAME = env("AWS_S3_STORAGE_BUCKET_NAME")
```
mikefeldberg: this is only used on line 314, so probably don't need to define it here. Unless you also add it to line 297, but I think doing `env("AWS_S3_STORAGE_BUCKET_NAME")`  in two places is fine

apps/betterangels-backend/clients/tests/test_mutations.py
```diff
@@ -867,3 +942,116 @@ def test_update_client_document(self) -> None:
...
+        self.assertEqual(result_docs[1]["mimeType"], "image/jpeg")
+        self.assertEqual(result_docs[1]["namespace"], ClientDocumentNamespaceEnum.OTHER_CLIENT_DOCUMENT.name)
+        self.assertIsNotNone(result_docs[1]["id"])
+
+        # Verify documents were created in the database
+        for doc in result_docs:
+            attachment = Attachment.objects.get(id=doc["id"])
+            self.assertEqual(attachment.object_id, int(self.client_profile_1["id"]))
+
+    def test_resolve_client_document_uploads_invalid_token(self) -> None:
```
mikefeldberg, line 1045: need a test where the second of two docs has a bad token. These mutations are atomic, so if any docs raise, the whole batch gets rolled back

apps/betterangels-backend/clients/services/client_document.py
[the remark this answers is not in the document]
mikefeldberg: @tglaz ☝️

apps/betterangels-backend/clients/tests/test_mutations.py
mikefeldberg, line 1045: maybe add unhappy path tests, to confirm we're returning the right error to the fe?

tglaz: yeah, good point. will do

apps/betterangels-backend/betterangels_backend/settings.py
tglaz: sounds good.

did it as generally I figured we'd want vars be available via settings, so never use env("MY_VAR") except to define the actual settings VAR. but works either way

apps/betterangels-backend/clients/services/client_profile_photo.py
tglaz: oh yeah. just forgot to update.
looked a bit more into some older image types that people may use and didn't update the doc service. will consolidate and try to centralize the list.

apps/betterangels-backend/clients/services/client_document.py
tglaz: yeah, i had actually started on it but got lazy. good call

apps/betterangels-backend/common/tests/test_models.py
```diff
@@ -427,3 +428,35 @@ def test_save(self) -> None:
...
+    def test_save_direct_upload(self) -> None:
+        file_name = "test%20file%20name.jpg"
+        file = SimpleUploadedFile(name=file_name, content=b"irrelevant")
+
+        content_type = ContentType.objects.get_for_model(Address)
+
+        attachment = Attachment(
+            content_type=content_type,
+            object_id=1,
+            file=file,
```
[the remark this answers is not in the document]
mikefeldberg: I added test infer_attachment_type (DEV-2145), which is more to the point but technically out of scope since the util was added earlier

apps/betterangels-backend/common/tests/test_storage.py
```diff
@@ -0,0 +1,45 @@
+from unittest.mock import Mock, patch
+
+from common.storage import LocalS3Storage
+from django.test import SimpleTestCase, override_settings
+
+
+@override_settings(LOCAL_S3_PUBLIC_ENDPOINT_URL="http://localhost:9000", AWS_REGION="us-west-2")
+class LocalS3StorageTest(SimpleTestCase):
```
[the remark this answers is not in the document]
mikefeldberg: +1 [robot comment](https://github.com/BetterAngelsLA/monorepo/pull/2015/changes#r3042183760)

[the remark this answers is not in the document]
mikefeldberg: +1 [robot comment](https://github.com/BetterAngelsLA/monorepo/pull/2015/changes#r3042183760)

apps/betterangels-backend/clients/tests/test_client_document_service.py
```diff
@@ -0,0 +1,390 @@
...
+                _validate_content_type(content_type)
+            except ValueError:
+                self.fail(f"_validate_content_type raised ValueError for allowed type: {content_type}")
+
+    def test_rejects_invalid_content_type(self) -> None:
+        with self.assertRaises(ValueError, msg="Unsupported content_type: application/zip."):
+            _validate_content_type("application/zip")
+
+    def test_rejects_empty_content_type(self) -> None:
+        with self.assertRaises(ValueError, msg="Unsupported content_type: ."):
```
mikefeldberg: sorry, should have caught this sooner

global comment: this isn't doing what it looks like it's doing. `msg` is _not_ the error message `ValueError` is expected to return, it's what will be printed by the test runner when the test fails, as additional context for the user. To test the actual error message, use

```py
        with self.assertRaisesRegex(ValueError, r"Unsupported content_type: application/zip."):
```

or

```py
        with self.assertRaisesMessage(ValueError, "Unsupported content_type: application/zip."):
```

apps/betterangels-backend/common/services/s3.py
```diff
@@ -0,0 +1,192 @@
...
+
+
+def s3_key_exists(*, key: str) -> bool:
+    """Check whether an object exists in the default S3 bucket."""
+    s3_client, bucket_name = _get_s3_client_and_bucket()
+    try:
+        s3_client.head_object(Bucket=bucket_name, Key=key)
+        return True
+    except ClientError as exc:
+        if exc.response["Error"]["Code"] == "404":
```
mikefeldberg, line 183: s3 doesn't always return a `404`. Based on these [error codes](https://docs.aws.amazon.com/AmazonS3/latest/API/ErrorResponses.html#ErrorCodeList), it looks like we should include `NoSuchKey` and `NotFoundException`

apps/betterangels-backend/clients/tests/test_mutations.py
tglaz: you think it's worth it to move the `transaction.atomic` wrapper inside `client_document.resolve_upload`?
keeping it at resolver level is fine, but if something fails then service itself should probably communicate what got saved (it's not guaranteed that a caller wraps it in a transaction).
I'll move the validation out of the `save` loop at least

apps/betterangels-backend/common/services/s3.py
tglaz: both of these return a 404 http code
also, here we're just calling [head_object](https://docs.aws.amazon.com/boto3/latest/reference/services/s3/client/head_object.html) which doesn't return error codes

apps/betterangels-backend/clients/services/client_profile_photo.py
```diff
@@ -0,0 +1,83 @@
...
+STORAGE_DIR = "media"
+CLIENT_PROFILE_PHOTO_PATH = f"{STORAGE_DIR}/client_profile_photos"
+SERVICE_NAME = "client_profile_photo"
+
+ALLOWED_CONTENT_TYPES = DEFAULT_IMAGE_CONTENT_TYPES
+
+
+def _validate_content_type(content_type: str) -> None:
+    if content_type not in ALLOWED_CONTENT_TYPES:
+        raise ValueError(f"Content type not allowed: {content_type}")
```
mikefeldberg, line 22: can this and the
```
def _validate_content_type(content_type: str, filename: str) -> None:
    if content_type not in ALLOWED_CONTENT_TYPES:
        raise ValueError(f"Unsupported content_type: {content_type} for filename={filename}.")
```
in `apps/betterangels-backend/clients/services/client_document.py` be consolidated in a `services/utils.py`?

apps/betterangels-backend/common/tests/test_storage.py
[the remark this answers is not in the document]
mikefeldberg: @tglaz sorry, this got missed because sourcery has an annoying habit of resolving comments that I'm actually trying to flag

apps/betterangels-backend/clients/services/client_document.py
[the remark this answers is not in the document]
mikefeldberg: @tglaz another one that sourcery resolved prematurely

apps/betterangels-backend/clients/services/client_profile_photo.py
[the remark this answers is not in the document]
mikefeldberg: @tglaz and another 😫

apps/betterangels-backend/clients/tests/test_mutations.py
mikefeldberg: > you think it's worth it to move the transaction.atomic wrapper inside client_document.resolve_upload

ya, I think that's a good call

mikefeldberg (commented):
one new request and a couple old ones that got missed because sourcery marked them resolved 😩

apps/betterangels-backend/clients/services/client_profile_photo.py
tglaz: yeah, but not worth it for me in this case. they're almost same right now but only by coincidence. they could easily have different validation messages (or even slightly different rules, apart from ALLOWED_CONTENT_TYPES)

apps/betterangels-backend/clients/tests/test_mutations.py
tglaz: done

apps/betterangels-backend/clients/services/client_document.py
tglaz: updated

mikefeldberg (approved):

PR #2035, opened 2026-04-13
base button update (DEV-2402)

Davit-BetterAngels:
DEV-2402

## Summary by Sourcery

Adjust BaseButton layout and press behavior to better control background feedback and height.

Enhancements:
- Move pressed background-color handling from an inner View to the Pressable container to simplify button structure and visual feedback.
- Set a default button height with optional override via style props and remove vertical padding from the container to refine spacing and layout.

### DIFF
--- a/libs/expo/betterangels/src/lib/ui-components/NavModal/components/BaseButton.tsx
+++ b/libs/expo/betterangels/src/lib/ui-components/NavModal/components/BaseButton.tsx
@@ -27,23 +27,16 @@ function BaseButton(props: BaseButtonProps) {
     <Pressable
       onPress={onPress}
       accessibilityRole="button"
-      style={[styles.container, style]}
+      style={({ pressed }) => [
+        {
+          backgroundColor: pressed ? Colors.NEUTRAL_EXTRA_LIGHT : Colors.WHITE,
+          height: style?.height || 72,
+        },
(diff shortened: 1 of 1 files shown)


shiozakiyuko (approved):

libs/expo/betterangels/src/lib/ui-components/NavModal/components/BaseButton.tsx
```diff
@@ -27,23 +27,16 @@ function BaseButton(props: BaseButtonProps) {
     <Pressable
       onPress={onPress}
       accessibilityRole="button"
-      style={[styles.container, style]}
+      style={({ pressed }) => [
+        {
+          backgroundColor: pressed ? Colors.NEUTRAL_EXTRA_LIGHT : Colors.WHITE,
+          height: style?.height || 72,
```
tglaz: won't `style` object below should automatically override this `height`? 
if so then can probably just set it directly as `72` // default

tglaz (approved):

libs/expo/betterangels/src/lib/ui-components/NavModal/components/BaseButton.tsx
Davit-BetterAngels: fixed

PR #2037, opened 2026-04-13
shelter max stay (SDB-149)

mikefeldberg:
SDB-149

- adds `max_stay` filter with optional "include null"
- adds `shelter_max_stay` query that returns the max `max_stay` to limit filter input

## Summary by Sourcery

Add support for querying and filtering shelters by maximum stay duration, including a dedicated query to expose the current maximum allowed value.

New Features:
- Introduce a maxStay filter on shelters that supports an optional includeNull flag to include shelters without a max_stay value.
- Expose a shelterMaxStay GraphQL query that returns the maximum max_stay among approved shelters for use in client filter limits.

Tests:
- Add GraphQL tests covering the shelter maxStay filter behavior with and without null values.
- Add GraphQL tests verifying shelterMaxStay returns the correct maximum and null when no max_stay values are set.


apps/betterangels-backend/shelters/schema.py
```diff
@@ -36,6 +39,10 @@ class Query:
         permission_classes=[IsAuthenticated],
     )
 
+    @strawberry.field()
+    def shelter_max_stay(self, info: Info) -> Optional[int]:
```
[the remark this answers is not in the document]
mikefeldberg: > while other shelter fields on Query enforce permission_classes=[IsAuthenticated]

no they don't
```
    shelter: ShelterType = strawberry_django.field()
    shelters: OffsetPaginated[ShelterType] = strawberry_django.offset_paginated()
```

apps/betterangels-backend/shelters/tests/test_shelter_queries.py
```diff
@@ -369,3 +369,34 @@ def test_shelters_query(self, mock_build_imgproxy_url: Mock) -> None:
         self.assertEqual(Shelter.objects.count(), shelter_count + 1)
         self.assertEqual(shelters[0]["heroImage"], exterior_photo_0.file.url)
         self.assertEqual(shelters[1]["heroImage"], interior_photo_1.file.url)
+
+
+class ShelterMaxStayQueryTestCase(ShelterGraphQLFixtureMixin, GraphQLBaseTestCase):
+    def test_shelter_max_stay_query(self) -> None:
+        shelter_recipe.make(max_stay=None, status=StatusChoices.APPROVED)
+        shelter_recipe.make(max_stay=0, status=StatusChoices.APPROVED)
```
tglaz, line 377: what would `max_stay=0` mean?

apps/betterangels-backend/shelters/types/filters.py
```diff
@@ -52,6 +52,12 @@ class ShelterPropertyInput:
     parking: Optional[List[ParkingChoices]] = None
 
 
+@strawberry.input
+class MaxStayInput:
+    days: int
```
tglaz, line 57: do we/should we validate for non-zero ints?

apps/betterangels-backend/shelters/tests/test_shelter_queries.py
mikefeldberg: `Shelter.max_stay` min value is `0` and we have shelters with `max_stay` of `0`. Seems to me that the floor should be `1`. I don't know if it was set at `0` intentionally or because we didn't have enough information/direction to choose a positive value. It's a product question but probably too low priority for anyone to have addressed yet

apps/betterangels-backend/shelters/types/filters.py
mikefeldberg: not sure what you mean. Negative values?

tglaz: where `days = -1`
i guess kind of related to the max_stay floor question above, but also not super urgent

tglaz (approved):

apps/betterangels-backend/shelters/types/filters.py
mikefeldberg: [max_stay is a PositiveIntegerField](https://github.com/BetterAngelsLA/monorepo/blob/b703b2198e493902db2dde6f13eee05c04c5f20a/apps/betterangels-backend/shelters/models/shelter.py#L102) so the db will reject negative values

PR #2039, opened 2026-04-14
DEV-2199 [BA - Doc Library] Upload progress panel, direct multi-doc upload & empty state

nparmar1:
## Summary

Improves the **BA - Doc Library** client screen: an empty state, a direct multi-document upload flow, and an app-wide upload progress panel that shows live progress with cancel, retry, and swipe-to-dismiss. Also fixes a cancel-upload race so **Cancel** reliably aborts the upload — including the persist step, which previously could still save the file.

Part of [DEV-2199](https://betterangels.atlassian.net/browse/DEV-2199).

## What's included

- **Empty state** — shows a "No files yet" illustration (native icon + text, no text-in-SVG) when a client has no documents.
- **Direct multi-document upload (AC2)** — tapping a doc type opens the media picker and uploads immediately; the form stays open so several documents can be uploaded in one session and is dismissed explicitly with **Done**.
- **Upload progress panel** — an app-wide bottom-sheet panel (the app's `BottomSheetPanel`, so it's a real swipeable sheet) shows the latest upload's stage, x-of-y counter, overall progress bar, and per-file status. Active uploads can't be dismissed (Cancel is the control); complete/failed sessions are dismissed by swiping the panel down. Failed sessions offer **Retry**.
- **Upload feedback** — success/failure states surface in the progress panel (toast/snackbar feedback removed).
- **Prevents overwriting** — doc types that already have a file are marked complete and disabled.
- **Reliable Cancel** — Cancel aborts the S3 upload *and* the persist step (previously, cancelling during "Saving…" still saved the file).

## Under the hood

- Pure, testable presigned-upload pipeline (`runPresignedUpload`) with typed errors, manifest mapping, throttled byte progress, and abort support across every stage — including the save step.
- Cross-platform S3 transport (`expo-file-system` on native, XHR on web).
- Module-scoped upload session store shared by the root and modal-screen providers, so progress survives navigation and modal close.
- HMIS upload flows reuse the same progress panel.
- Shared `toPresignedUploads` mapper and `abortableContext` helper keep the Docs and profile-photo upload adapters DRY.

## Cleanup included in this PR

- Deleted the dead tab-based upload flow (`ClientDocumentUploads`, `UploadSection`, `UploadsPreview`, the `TABS` map).
- Consolidated duplicate upload logic into a single path in `UploadModal`.
- Removed a no-op `onDismiss` prop chain in `MediaPicker` / `BottomSheetModalControlled`.
- Replaced the text-bearing `no-files-yet.svg` with a native `EmptyState` component using an existing icon + design tokens.
- Removed dead `IUploadModalProps` fields and narrowed the `DocUploads` type to the namespaces the modal exposes.
- `s3/types.ts` is now type-only (runtime exports moved to the barrel) and the `EXPO_PUBLIC_SIMULATE_UPLOAD_DELAY_MS` test tooling was stripped from the shipped uploader.
- Merged latest `main`.

## CI / maintenance included

- **Unrelated to DEV-2199:** bumped `react-native-gesture-handler` `^2.32.0` → `~3.1.0` to match Expo SDK 57. This fixes a pre-existing `expo-doctor` CI failure that trips any PR touching `apps/betterangels`, and is required for this PR's checks to pass. Gorhom bottom-sheet's peer range is satisfied by 3.1.0.

## Tests

Added/updated unit tests (vitest):
- `EmptyState` copy
- `Docs` empty vs. populated branching
- `UploadModal` namespace selection, error handling, keep-open + Done
- `UploadProgressDrawer` / `UploadProgressProvider` / `useUploadSession`
- `runPresignedUpload` pipeline (stages, byte progress, throttling, abort incl. cancel-during-save)
- `toPresignedUploads` mapper
- S3 web transport

## Follow-ups

- The HMIS clone (DEV-2354) should apply the same cleanup (HMIS upload flows already share the progress panel).

[DEV-2199]: https://betterangels.atlassian.net/browse/DEV-2199


shiozakiyuko (commented):
Hey @nparmar1 

The flow looks good, but since the redundant Upload page is still visible in the background, AC2 isn't quite fulfilled yet. Please check. Thanks!

Image

shiozakiyuko (commented):
Hey @nparmar1 

I think we are almost there! 
So when Michelle and I tested, this is what we saw. 

Image

Instead, we would like to have the 4 buttons on this page so the AC2 will be completed. 

Image

And that's it. Let me know if anything is unclear. Thanks so much!

shiozakiyuko (commented):
Hey @nparmar1 !

Ok, AC2 looks good! 

A couple of things I noticed:

- It takes a while to upload to complete the process after the upload. Can you do anything about that?
- The width of notification bar is shorter than the other ones (please see the screen shots). Please be consistent on UI. Thank you so much!

Image

Image

shiozakiyuko (approved):
Beautiful! Thanks so much. It looks great!

libs/expo/shared/icons/src/lib/components/NoFilesYet.tsx
```diff
@@ -0,0 +1,6 @@
+import NoFilesYetSVG from '../../assets/better_angels/no-files-yet.svg';
+import createSvgIcon from '../createSvgIcon';
+
+const NoFilesYet = createSvgIcon(NoFilesYetSVG);
```
tglaz: Don't think we need this SVG (we want to avoid text within svgs unless it's some unique style that cannot be replicated with styles/css).
Pretty sure we already have the Icon, so can just create a component that renders the existing icon with text.

libs/expo/betterangels/src/lib/screens/Client/Docs/UploadModal/ClientDocumentUploads/ClientDocumentUploads.tsx
```diff
@@ -26,44 +27,58 @@ export function ClientDocumentUploads(props: IClientDocUploadsProps) {
...
-      return;
-    }
+  useEffect(() => {
+    setIsModalVisible(true);
+  }, []);
+
+  const handleMediaPickerClose = () => {
+    setIsModalVisible(false);
+
+    setTimeout(() => {
```
tglaz: is it possible to avoid this and rely on callbacks?
think I understand the motivation here, but fixed timeouts can be brittle and hopefully we'd have a callback that tells us when the modal close has completed.

libs/expo/betterangels/src/lib/screens/Client/Docs/UploadModal/ClientDocumentUploads/ClientDocumentUploads.tsx
```diff
@@ -26,44 +27,58 @@ export function ClientDocumentUploads(props: IClientDocUploadsProps) {
...
-        message: `Sorry, there was an error with the file upload.`,
-        type: 'error',
-      });
+      onUploadError?.();
     } finally {
       setProcessing(false);
     }
   };
 
+  const handleUpload = async () => {
```
tglaz: any reason to not just call `uploadSelectedFiles` directly?
handleUpload may be redundant here

libs/expo/betterangels/src/lib/screens/Client/Docs/UploadModal/ClientDocumentUploads/ClientDocumentUploads.tsx
```diff
@@ -78,6 +93,25 @@ export function ClientDocumentUploads(props: IClientDocUploadsProps) {
     return !!file.name && !!file.type && !!file.uri;
   });
 
+  if (processing) {
+    return (
+      <View
```
tglaz: This is small enough, but can either move it out to separate component inside `ClientDocumentUploads` directory.
Better yet, this seems pretty generic, so can we reuse a component like `LoadingView` from our Lib?

libs/expo/betterangels/src/lib/screens/Client/Docs/UploadModal/ClientDocumentUploads/ClientDocumentUploads.tsx
```diff
@@ -78,6 +93,25 @@ export function ClientDocumentUploads(props: IClientDocUploadsProps) {
...
+          color={Colors.PRIMARY}
+          style={{ transform: [{ translateY: -40 }] }}
+        />
+      </View>
+    );
+  }
+
   return (
     <>
       <UploadSection
```
tglaz: is this ever used/rendered? seems like this flow is meant to remove the submit/cancel buttons we used and upload directly? haven't fully tested but wanted to confirm

tglaz (dismissed):

vecchp:
Hi @tglaz — following up on your review from May. All five points are now addressed (the last two via a cleanup pass on top of this PR):

1. **`NoFilesYet` text-SVG** — removed. The SVG (which embedded "No files yet" as vector text) is deleted. Replaced with a native `EmptyState` component that reuses an existing icon (`mdi:file-outline` → `FileOutlineIcon`) with a `PlusIcon` badge, plus real `TextBold`/`TextRegular` copy, all using shared design tokens.
2. **Fixed timeouts for modal close** — replaced with callbacks (`onSelectionComplete`); no timeouts remain.
3. **`handleUpload` redundancy** — removed; the flow calls `uploadSelectedFiles` directly.
4. **Loading UI** — now reuses the shared `LoadingView` from `@monorepo/expo/shared/ui-components`.
5. **Submit/cancel flow "ever used?"** — you were right. The whole tab-based flow (`ClientDocumentUploads`, `UploadSection`, `UploadsPreview`, the `TABS` map) was dead code and has been deleted. `UploadModal` is now a single direct-upload path.

Also fixed along the way:
- Removed the duplicate error toast (upload errors now go through one callback path).
- Removed the no-op `onDismiss` prop chain in `MediaPicker`/`BottomSheetModalControlled`.
- Modal stays open on upload failure so the user can retry.
- Merged latest `main` and added unit tests for the empty state, docs branching, and upload wiring.

Could you take another look? Thanks!

shiozakiyuko (commented):
Hey @vecchp 

I noticed one thing after I tested it.

The Cancel Upload button doesn't really cancel uploading an image sometimes. 
This doesn't happen every time so I don't know how I can reproduce this but luckily,
I was able to capture when it happened. Check it out.
When the cancel doesn't work, it goes back to the Doc Library front page, instead of staying at Upload Files page. 

https://github.com/user-attachments/assets/170a01e3-a3fc-4dd2-823d-f34620ddc741

vecchp:
Temporarily closing to force a fresh pull_request CI event (workflow dispatch missed on d8ac36f0/1df61ec1)

vecchp (approved):

PR #2043, opened 2026-04-15
shelter contact is_claimant field & filter (SDB-129)

mikefeldberg:
SDB-129

## Summary by Sourcery

Add support for marking shelter contact records as claimants and expose this in the admin interface and event tracking.

New Features:
- Introduce an is_claimant flag on shelter contact records to indicate claimant status.
- Add an admin list filter on shelters to filter by presence of a claimant contact.
- Display the claimant flag on additional contact entries in the shelter admin inline form.

Enhancements:
- Extend contact info event triggers to capture and persist the claimant flag for auditing.


apps/betterangels-backend/shelters/admin.py
```diff
@@ -868,6 +869,25 @@ def queryset(self, request: HttpRequest, queryset: QuerySet) -> QuerySet:
...
+        ]
+
+    def queryset(self, request: HttpRequest, queryset: QuerySet) -> QuerySet:
+        value = self.value()
+        if value == "yes":
+            return queryset.filter(additional_contacts__is_claimant=True).distinct()
+        if value == "no":
+            return queryset.exclude(additional_contacts__is_claimant=True).distinct()
+        return queryset
+
```
vecchp, line 890:
```python
    def queryset(self, request: HttpRequest, queryset: QuerySet) -> QuerySet:
        value = self.value()
        claimant_exists = models.Exists(ContactInfo.objects.filter(shelter=OuterRef("pk"), is_claimant=True))
        if value == "yes":
            return queryset.filter(claimant_exists)
        if value == "no":
            return queryset.exclude(claimant_exists)
        return queryset
```
this is likely better as it doesn't use the distinct.

apps/betterangels-backend/shelters/models/shelter.py
```diff
@@ -258,6 +258,7 @@ class ContactInfo(models.Model):
     contact_number = PhoneNumberField(verbose_name="Contact Number")
     contact_email = models.EmailField(blank=True, null=True)
     contact_title = models.CharField(max_length=255, blank=True, null=True)
+    is_claimant = models.BooleanField(default=False)
```
vecchp: If we are going to filter on this perhaps an index?

vecchp (changes requested):

vecchp (approved):

PR #2056, opened 2026-04-21
filters on separate screen (SDB-93)

Davit-BetterAngels:
SDB-93

## Summary by Sourcery

Move shelter property filters from a modal to a dedicated filters page and wire it into the routing and search flow.

New Features:
- Add a standalone filters page and route for configuring shelter property filters.

Enhancements:
- Refactor ShelterFilters to be a controlled component driven by external filter state.
- Update shelter search to navigate to the filters page and always query using the global filter state.
- Expose a null/default shelter property filters constant for reusing initial filter values.
- Simplify filter action handling to delegate reset behavior to callers.

### DIFF
--- a/libs/react/shelter/src/lib/pages/filters/FiltersPage.tsx
+++ b/libs/react/shelter/src/lib/pages/filters/FiltersPage.tsx
@@ -6,9 +6,11 @@ import {
   nullShelterPropertyFilters,
   shelterPropertyFiltersAtom,
 } from '../../atoms';
+import {
+  FiltersActions,
+  ShelterFilters,
+} from '../../components/ShelterFilters';
 import { shelterHomePath } from '../../constants';
-import { FiltersActions } from './FiltersActions';
-import { ShelterFilters } from './ShelterFilters';
 
 export function FiltersPage() {
   const navigate = useNavigate();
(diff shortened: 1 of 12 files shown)


libs/react/shelter/src/lib/pages/filters/FiltersPage.tsx
```diff
@@ -0,0 +1,60 @@
+import { CloseIcon } from '@monorepo/react/icons';
```
tglaz, line 1: this is now a page that we navigate to using routing, so may be more consistent to put it under `pages/filters/FiltersPage.tsx` or `ShelterFiltersPage` (if we end up with other filters later)

less urgent:
we can still render it so the presentation is more like a modal, but we can separate that from the core page itself a bit. We could  replace this css `absolute ... animate-slideInUp ...` + close by wrapping the entire page with <Modal>, kind of like this.
Unless that breaks some requirement. 

```
  return (
    <Modal
      type="fullscreen"
      onClose={handleClose}
      footer={
        <FiltersActions
          className="pb-8"
          onDone={handleDone}
          onReset={handleReset}
        />
      }
    >
      <ShelterFilters
        className="w-full"
        filters={draftFilters}
        onFiltersChange={setDraftFilters}
      />
    </Modal>
  );
```

tglaz (changes requested):

shiozakiyuko (approved):
It looks good to me.

tglaz (approved):

PR #2061, opened 2026-04-23
revert minio user pass to default

mikefeldberg:
seems minio build fails with `admin`/`password` but succeeds with `minioadmin`/`minioadmin`

## Summary by Sourcery

Revert local MinIO service credentials in docker-compose to use the default minioadmin/minioadmin configuration for compatibility with the current image.

Build:
- Update docker-compose MinIO root user and password to the default minioadmin/minioadmin values used by the MinIO image.

Chores:
- Adjust mc alias initialization in docker-compose to authenticate with the reverted MinIO default credentials.


.compose/local.shared.env
```diff
@@ -10,5 +10,5 @@ IS_LOCAL_DEV=True
 AWS_S3_STORAGE_BUCKET_NAME=betterangels-local
 LOCAL_S3_INTERNAL_ENDPOINT_URL=http://minio:9000
 LOCAL_S3_PUBLIC_ENDPOINT_URL=http://localhost:9000
-AWS_ACCESS_KEY_ID=admin
-AWS_SECRET_ACCESS_KEY=password
+AWS_ACCESS_KEY_ID=minioadmin
+AWS_SECRET_ACCESS_KEY=minioadmin
```
tglaz, line 14: can we confirm this is the actual issue?
not saying it's not, but figured i'd be able to repro

if we do change then let's update readme as well

mikefeldberg: It's not. You and Paul didn't run into it. And it's not breaking my builds now. But it _was_ a switch I could flip to break/fix the minio container build 🤷‍♂️

mikefeldberg:
https://github.com/BetterAngelsLA/monorepo/pull/2061#discussion_r3162811345

PR #2062, opened 2026-04-27
vaccinations (SDB-145, SDB-175)

Davit-BetterAngels:
SDB-145
SDB-175

## Summary by Sourcery

Add support for recording and displaying shelter vaccination requirements across backend models, GraphQL schema, and frontend views.

New Features:
- Introduce a Vaccination model and enum to represent TB, flu, and COVID-19 vaccination options for shelters.
- Allow shelters to be associated with multiple vaccinations via a new many-to-many relationship and include this in shelter creation inputs and admin interfaces.
- Expose shelter vaccinations through the GraphQL API and render them on the shelter page UI.

Enhancements:
- Track vaccination add/remove events for shelters using pghistory triggers for auditing.


apps/betterangels-backend/shelters/models/shelter.py
```diff
@@ -116,6 +117,7 @@ class Shelter(BaseModel):
     entry_info = CKEditor5Field(null=True, blank=True)
     entry_requirements = models.ManyToManyField(EntryRequirement)
     referral_requirement = models.ManyToManyField(ReferralRequirement)
+    vaccinations = models.ManyToManyField(Vaccination)
```
mikefeldberg: this should probably be `vaccination_requirement`

apps/betterangels-backend/shelters/models/lookups.py
```diff
@@ -94,6 +95,13 @@ def __str__(self) -> str:
         return str(self.name)
 
 
+class Vaccination(models.Model):
+    name = TextChoicesField(choices_enum=VaccinationChoices, unique=True, blank=True, null=True)
```
mikefeldberg: this should be required
```suggestion
    name = TextChoicesField(choices_enum=VaccinationChoices)
```

not sure why all the existing choices are `blank=True, null=True`. Maybe it was meant to be temporary during a data migration? Generally don't want to combine `unique=True` and `blank=True` since two empty strings would violate the unique constraint

mikefeldberg (commented):
couple comments. looks like typecheck failing, too. The description says this is for 147 but that's a closed ticket. can you please remove or update if it's a typo?

apps/betterangels-backend/shelters/models/lookups.py
```diff
@@ -94,6 +95,13 @@ def __str__(self) -> str:
         return str(self.name)
 
 
+class Vaccination(models.Model):
```
mikefeldberg: since we updated the field from `vaccinations` to `vaccination_requirement`, all related items should be updated, too: models, types, inputs, fe fields, etc. Same pattern as the other shelter fields

apps/betterangels-backend/shelters/enums.py
```diff
@@ -111,6 +111,13 @@ class EntryRequirementChoices(models.TextChoices):
     IN_SPA_ONLY = "in_spa_only", _("In-SPA Only")
 
 
+@strawberry.enum
+class VaccinationRequirementChoices(models.TextChoices):
```
tglaz, line 115: Not blocking, but a comment:
But we might want `Vaccinations` enum as shared enums, and `VaccinationRequirementChoices` for shelters would use those enums?

Only bring it up cause migrating enums can be a pain (could be wrong), and we might need to surface them in Outreach App or elsewhere in the future.
@mikefeldberg thoughts?

tglaz (approved):

apps/betterangels-backend/shelters/enums.py
mikefeldberg: I think it's fine to leave here. Not a big deal if we have to move later. A lot of things may end up getting shared by Outreach and Shelter at some point. We've kept `common` limited to actively shared resources

PR #2063, opened 2026-04-29
Private shelter data model changes (DEV-2419, DEV-2420)

vecchp:
DEV-2419, DEV-2420
## Summary by Sourcery

Introduce private shelter visibility controls and expose the new privacy flag across the data model and GraphQL API.

New Features:
- Add an is_private flag to shelters and shelter events to support private shelter records.
- Expose the isPrivate field on shelter GraphQL types, filters, and creation input so clients can set and query shelter privacy.
- Introduce a view_private_shelter permission and grant it to the Caseworker permission group template to control access to private shelters.

Enhancements:
- Update shelter listing selector and GraphQL queryset resolution to respect the private shelter permission when returning shelters.
- Extend shelter admin configuration to surface and manage the new is_private field.
- Adjust shelter database indexes and audit triggers to include the is_private attribute for efficient querying and event logging.


apps/betterangels-backend/shelters/migrations/0046_add_view_private_shelter_to_caseworker.py
```diff
@@ -0,0 +1,42 @@
+from django.db import migrations
+
+
+def add_view_private_shelter_to_caseworker(apps, schema_editor):
```
mikefeldberg, line 4: we need to update the permission group templates to separate Caseworkers and Volunteers. Doesn't need to be done here, but before this feature goes live. Since Caseworker is the default template, anyone with access to Outreach will have access to private shelters

apps/betterangels-backend/shelters/selectors.py
```diff
@@ -15,12 +15,24 @@
...
     from accounts.models import User
+    from django.contrib.auth.base_user import AbstractBaseUser
+    from django.contrib.auth.models import AnonymousUser
     from shelters.models import Shelter
 
 
-def shelter_list(queryset: "QuerySet[Shelter]") -> "QuerySet[Shelter]":
-    """Filter to shelters approved for public display."""
-    return queryset.filter(status=StatusChoices.APPROVED)
+def shelter_list(
```
mikefeldberg, line 23: please add permission test verifying that `shelter` and `shelters` queries only return private shelters to authorized users. Tho, until we update the permission group templates, most authenticated users are authorized

mikefeldberg (approved):

PR #2066, opened 2026-04-30
Add Sign In/Sign Out Flow to Shelter Directory (SDB-188, DEV-2422, DEV-2423)

vecchp:
SDB-188
DEV-2422
DEV-2423

### Summary

Adds authentication UI to the public shelter directory so case managers can log in to access private shelter information. Unauthenticated users see a dismissible banner prompting them to sign in, and the navbar provides persistent sign in/out controls.

### What Changed

**Sign-In Route**
- Added `/sign-in` route to the public shelter-web app, reusing the existing `<SignIn>` component with redirect back to `/` on success
- Wrapped email/password and OTP steps in `<form>` elements so users can submit with Enter key

**Login Banner (HomePage)**
- Dismissible notification overlaid on the map: "Sign in with your Better Angels account to access privately shared shelters."
- Only shown to unauthenticated users
- Dismiss state persisted in localStorage

**NavBar — Desktop**
- Authenticated users see a circular avatar button (initials or user icon) with a dropdown showing their name, email, and "Sign Out"
- Unauthenticated users see a "Sign In" link

**NavBar — Mobile**
- Sign In / Sign Out added as the last item in the hamburger flyout menu
- Flyout closes on both Sign In navigation and Sign Out action

**Sign Out Hook**
- Added `useSignOut` hook (`logout` GraphQL mutation + `setUser(undefined)`)
- Mirrors the existing pattern in betterangels-admin

**Layout Fixes**
- Changed `overflow-x-hidden` → `overflow-x-clip` on MainLayout (prevents dropdown clipping)
- Added `overflow-visible` to Header, `relative z-50` to header wrapper


libs/react/shelter/src/lib/components/Map/Map.tsx
```diff
@@ -117,7 +117,14 @@ export function Map(props: TMap) {
           zIndex={99}
           onClick={marker.onClick}
         >
-          <MapPinIcon className="h-10" type={marker.type} />
+          <div className="relative">
+            <MapPinIcon className="h-10" type={marker.type} />
+            {marker.isPrivate && (
+              <div className="absolute -top-1 -right-1 bg-white rounded-full p-0.5 shadow-sm">
+                <LockIcon className="w-3 h-3 text-primary-60" />
```
tglaz, line 124: probably ok for now, but at some scale (100+ pins) on a phone the rendering may get expensive.
not sure we'd need to optimize yet, but a couple options are:
- create separate MapPinIconLocked that bakes in the lock icon. MapPinIcon is already customized, so can just create separate svg
- just use color (no lock icon). using `light gray` could probably already signal that it's "disabled"

tglaz (approved):

PR #2069, opened 2026-05-05
fix: hide empty shelter fields and restore missing Restrictions fields (SDB-76)

vecchp:
SDB-76

**Bug Fixes:**

1. **"Other Rules" showing when empty** — `WysiwygSection` used `!content?.length` to guard rendering, which only checks raw string length. Empty WYSIWYG HTML like `<p>&nbsp;</p>` or `<p></p>` passed this check, causing the section title to render with no visible content. Fixed by using `hasWysiwygContent()` which strips tags and whitespace entities before checking.

2. **"On-site Security" and "Other Rules" missing from Restrictions** — Commit `8659cadbf` ("display missing shelter fields") accidentally removed both fields from the `Restrictions` component's render output while adding exit policy and other fields. The `hasRestrictions` guard in `ShelterPage` still referenced them (so the card appeared), but nothing rendered them. Re-added both.

---

**Refactor (supporting testability):**

- Extracted section-level visibility logic from `ShelterPage` and field-level visibility from `Restrictions` into pure functions (`getShelterVisibility`, `getRestrictionsFieldVisibility`) in `utils/`. This allows testing all show/hide rules as plain data-in → booleans-out with zero React/Apollo/mock dependencies.

---

**Test Coverage:**

- 36 new unit tests covering all section and field visibility rules (empty HTML, null, false-vs-null booleans, populated content).

## Summary by Sourcery

Fix visibility of shelter restrictions and WYSIWYG-based sections while centralizing show/hide logic for easier testing.

Bug Fixes:
- Hide WYSIWYG sections whose HTML content is effectively empty so headings do not display without visible text.
- Restore missing On-site Security and Other Rules fields in the Restrictions section of the shelter page.

Enhancements:
- Extract shelter-level and restrictions field-level visibility rules into shared utility functions to standardize section rendering logic.

Tests:
- Add unit tests covering shelter section visibility and individual restriction field visibility for various empty and populated data scenarios.


libs/react/shelter/src/lib/pages/shelter/ShelterPage.tsx
```diff
@@ -39,54 +39,13 @@ export function ShelterPage({ id }: { id: string }) {
...
-    !!shelter.cities?.length ||
-    !!shelter.spa?.length ||
-    !!shelter.cityCouncilDistrict ||
-    !!shelter.supervisorialDistrict ||
-    !!shelter.shelterPrograms?.length ||
-    !!shelter.funders?.length;
-  const hasPhotos =
-    !!shelter.interiorPhotos?.length || !!shelter.exteriorPhotos?.length;
-  const hasMedia = hasPhotos || !!shelter.mediaLinks?.length;
+  const visibility = getShelterVisibility(shelter);
```
mikefeldberg, line 42: neat

mikefeldberg (approved):
nice lil clean up

PR #2075, opened 2026-05-12
update seed_shelters names to reflect shelter db id (SDB-187)

tglaz:
### [DEV Env] Simplify shelter detail access via Name/ID
SDB-187

This outputs names like below.
Also added `ID` column.

image

## Summary by Sourcery

Generate human-readable seeded shelter names that include their database IDs and surface more varied name lengths for UI testing.

New Features:
- Generate realistic, length-varied shelter names during seeding and append each shelter's database ID to its name.
- Display the shelter ID column in the Django admin shelter list view.

Build:
- Add Faker as a development dependency for generating seeded shelter names.


mikefeldberg (approved):
a welcome upgrade, thanks!

PR #2078, opened 2026-05-14
bed availability BE (SDB-193)

Davit-BetterAngels:
SDB-193

## Summary by Sourcery

Introduce a dedicated model and history tracking for shelter bed availability, with admin and testing support.

New Features:
- Add a ShelterAvailability model to track restrictive and non-restrictive bed counts per shelter with optional restriction notes and updater information.
- Expose shelter availability records in the Django admin with appropriate fields, filters, and automatic tracking of the updating user.

Enhancements:
- Track shelter availability changes via pghistory events for insert, update, and delete operations.
- Extend test data baker recipes to generate valid ShelterAvailability instances for tests.

Tests:
- Add model tests covering ShelterAvailability defaults, validation, cascade behavior, event tracking, ordering, related name, and baker recipe validity.


apps/betterangels-backend/shelters/models/availability.py
```diff
@@ -0,0 +1,67 @@
...
+)
+class ShelterAvailability(BaseModel):
+    objects: models.Manager["ShelterAvailability"] = models.Manager()
+
+    shelter = models.ForeignKey(
+        Shelter,
+        on_delete=models.CASCADE,
+        related_name="availabilities",
+    )
+    non_restrictive_beds = models.PositiveIntegerField(
```
mikefeldberg: maybe change `restrictive` to `restricted` everywhere

apps/betterangels-backend/shelters/models/availability.py
```diff
@@ -0,0 +1,67 @@
...
+        verbose_name = "Shelter Availability"
+        verbose_name_plural = "Shelter Availabilities"
+        ordering = ["-updated_at"]
+
+    def __str__(self) -> str:
+        return (
+            f"{self.shelter.name} — {self.non_restrictive_beds} non-restrictive / {self.restrictive_beds} restrictive"
+        )
+
+    def clean(self) -> None:
```
mikefeldberg: these restrictions are set at the db level so there's no need to check manually. see: [robot's comment](https://github.com/BetterAngelsLA/monorepo/pull/2078/changes#r3242823092)

apps/betterangels-backend/shelters/models/availability.py
```diff
@@ -0,0 +1,67 @@
...
+
+@pghistory.track(
+    pghistory.InsertEvent("shelter.availability.add"),
+    pghistory.UpdateEvent("shelter.availability.update"),
+    pghistory.DeleteEvent("shelter.availability.remove"),
+)
+class ShelterAvailability(BaseModel):
+    objects: models.Manager["ShelterAvailability"] = models.Manager()
+
+    shelter = models.ForeignKey(
```
mikefeldberg: this should be a `OneToOneField`. `ShelterAvailability` functions more as extension of shelter fields than as a true many to one object. We would have added it directly to the `Shelter` model if it weren't intended as a temporary measure, while `Bed`, `Room`, `Reservation`, etc. functionality is built out

apps/betterangels-backend/shelters/models/availability.py
```diff
@@ -0,0 +1,67 @@
...
+        default=0,
+        validators=[MinValueValidator(0)],
+        help_text="Number of restrictive beds available.",
+    )
+    restriction_notes = models.TextField(
+        blank=True,
+        default="",
+        help_text="Global restriction note for this availability record.",
+    )
+    updated_by = models.ForeignKey(
```
mikefeldberg: this is tracked by pghistory. Rather than storing here and using a custom `save_model` in [admin](https://github.com/BetterAngelsLA/monorepo/pull/2078/changes#diff-a30bc6ae309e92fffcf5d6e45ec9d21fb9e34344ba7be6f0dd8b81a3f0cfc6d2R1608), you can get it from history, like this: [resolver](https://github.com/BetterAngelsLA/monorepo/blob/165be9c1384cec3707d6bae017b7b959a9828176/apps/betterangels-backend/shelters/admin.py#L1335), [qs annotation](https://github.com/BetterAngelsLA/monorepo/blob/165be9c1384cec3707d6bae017b7b959a9828176/apps/betterangels-backend/shelters/admin.py#L1194)

apps/betterangels-backend/shelters/tests/test_availability.py
```diff
@@ -0,0 +1,130 @@
...
+from django.test import TestCase
+from model_bakery import baker
+from pghistory.models import Events
+from shelters.models import Shelter, ShelterAvailability
+from shelters.tests.baker_recipes import shelter_availability_recipe
+
+User = get_user_model()
+
+
+class ShelterAvailabilityModelTestCase(TestCase):
```
mikefeldberg: we don't need these tests. Most of these are testing base django (or pghistory) functionality, not our code. The two `test_clean` tests can come out because the custom `clean` is getting removed

apps/betterangels-backend/shelters/admin.py
```diff
@@ -1571,3 +1572,39 @@ class ReservationAdmin(admin.ModelAdmin):
             {"fields": ("notes",)},
         ),
     )
+
+
+@admin.register(ShelterAvailability)
+class ShelterAvailabilityAdmin(admin.ModelAdmin):
```
mikefeldberg, line 1594: please add search by shelter name

mikefeldberg (changes requested):

apps/betterangels-backend/shelters/admin.py
Davit-BetterAngels: it is already there

mikefeldberg: oh duh. sorry

apps/betterangels-backend/shelters/tests/test_availability.py
```diff
@@ -0,0 +1,23 @@
+from django.test import TestCase
+from shelters.models import Shelter, ShelterAvailability
+from shelters.tests.baker_recipes import shelter_availability_recipe
+
+
+class ShelterAvailabilityModelTestCase(TestCase):
```
mikefeldberg: you can delete this whole file. Availability is only exposed in django admin, and there's no custom logic, just basic crud. So there's nothing to test there. str representation is core django and recipe is core baker. Maybe if we had some very elaborate logic, it'd be worth testing.

apps/betterangels-backend/shelters/admin.py
```diff
@@ -1571,3 +1572,67 @@ class ReservationAdmin(admin.ModelAdmin):
...
+    list_display = ("shelter", "non_restricted_beds", "restricted_beds", "updated_by", "updated_at", "created_at")
+    list_filter = ("updated_at",)
+    search_fields = ("shelter__name",)
+    autocomplete_fields = ["shelter"]
+    readonly_fields = ("created_at", "updated_at", "updated_by")
```
mikefeldberg: I don't think `updated_by` needs to be in readonly since it's a computed value; there's no field to write to

apps/betterangels-backend/shelters/models/availability.py
```diff
@@ -0,0 +1,46 @@
...
+
+from .shelter import Shelter
+
+
+@pghistory.track(
+    pghistory.InsertEvent("shelter.availability.add"),
+    pghistory.UpdateEvent("shelter.availability.update"),
+    pghistory.DeleteEvent("shelter.availability.remove"),
+)
+class ShelterAvailability(BaseModel):
```
mikefeldberg, line 16: I think we should do two more things in this PR:
1. add a data migration to create a `ShelterAvailability` for each existing shelter
2. add a `post_save` signal to `Shelter` that will create a `ShelterAvailability` whenever a new shelter is created (see: apps/betterangels-backend/accounts/signals.py)

that will avoid issues like this
Image

mikefeldberg (commented):
a few more things. And CI is failing because of a migration conflict

apps/betterangels-backend/shelters/tests/baker_recipes.py
```diff
@@ -195,6 +196,15 @@ def make_services() -> list[Service]:
 )
 
 
+shelter_availability_recipe = Recipe(
+    ShelterAvailability,
+    shelter=foreign_key(shelter_recipe),
+    non_restricted_beds=lambda: random.randint(0, 50),
+    restricted_beds=lambda: random.randint(0, 20),
+    restriction_notes=seq("restriction note "),  # type: ignore
+)
```
mikefeldberg: don't need this now that the signal creates automatically

apps/betterangels-backend/shelters/migrations/0059_populate_shelter_availability.py
```diff
@@ -0,0 +1,24 @@
+# Generated by Django 6.0.4 on 2026-05-20 15:18
+
+from django.db import migrations
+
+
+def create_availability_for_existing_shelters(apps, schema_editor):
+    Shelter = apps.get_model("shelters", "Shelter")
+    ShelterAvailability = apps.get_model("shelters", "ShelterAvailability")
+    shelters_without = Shelter.objects.exclude(pk__in=ShelterAvailability.objects.values_list("shelter_id", flat=True))
+    ShelterAvailability.objects.bulk_create([ShelterAvailability(shelter=shelter) for shelter in shelters_without])
```
mikefeldberg: this is a little cleaner
```suggestion
    missing_shelter_ids = Shelter.objects.filter(availability__isnull=True).values_list('id', flat=True)
    ShelterAvailability.objects.bulk_create([ShelterAvailability(shelter_id=id) for id in missing_shelter_ids])
```

apps/betterangels-backend/shelters/models/availability.py
```diff
@@ -0,0 +1,46 @@
...
+    objects: models.Manager["ShelterAvailability"] = models.Manager()
+
+    shelter = models.OneToOneField(
```
mikefeldberg, line 19: [even better solution from Paul's PR](https://github.com/BetterAngelsLA/monorepo/pull/2089/changes/BASE..4fbd8b4589165b0471684d5af19bffe44258a3a4#diff-efde293d14c83a68bc187093ac48f3a6061d5fce7ffd6d42c42aa446164fb80aR199). Can use this instead of a signal

apps/betterangels-backend/shelters/tests/baker_recipes.py
```diff
@@ -42,6 +42,7 @@
     Schedule,
     Service,
     Shelter,
+    ShelterAvailability,
```
mikefeldberg: unused import breaking lint

apps/betterangels-backend/shelters/models/availability.py
Davit-BetterAngels: @mikefeldberg can't understand this one

mikefeldberg: ah, never mind actually. Paul added a lib with an `AutoOneToOneField` but I misunderstood how it works. I _thought_ that it created a related instance when an object is created (e.g., anytime a `Shelter` is created, a `ShelterAvailability` is automatically created for it). But it _actually_ creates a related instance when that field is accessed (e.g., when someone accesses `shelter.availability`, a `ShelterAvailability` is created)

mikefeldberg (approved):

PR #2089, opened 2026-05-21
feat: multi-org permission refactor (SDB-178)

vecchp:
SDB-178
- Add OrgType + OrganizationProfile models (M2M org types)
- Add ORG_TYPE_PRESETS config for org-type-driven permission templates
- Refactor permission assignment: org-type-aware member roles
- Add get_permission_group_for_org / get_member_permission_group utils
- Deprecate get_user_permission_group (legacy fallback)
- Plumb organization_id through notes/tasks/clients/hmis mutations
- Move caseworkerOrganizations query to notes schema
- Remove isOutreachAuthorized (replaced by userPermissions check)
- Extract CASEWORKER constant to notes/groups.py
- Add SHELTER_OPERATOR template + invite email
- Role-based invitation email template selection
- Add mobile app org preference UI + state
- Remove dead shelters/permissions.py + deprecated_permissions.py
- Simplify permission group creation signals

## Summary by Sourcery

Refactor organization permissions to be driven by org-type presets, introduce multi-org-aware member role resolution, and add mobile support for selecting a default organization used in backend mutations.

New Features:
- Introduce OrgType and OrganizationProfile models with org-type presets to drive organization-specific permission templates and member roles.
- Add role-based invitation email templates, including a Shelter Operator invite flow, and org-type-aware organization creation and membership services.
- Expose organization_id across notes, HMIS, tasks, client documents, and report flows to select the correct permission group per organization, and add GraphQL/mobile wiring accordingly.
- Add mobile user preference and UI for selecting a default organization used when creating interactions, tasks, and documents.

Bug Fixes:
- Harden permission group resolution by validating that users belong to the resolved member group and by avoiding admin/superuser groups for member actions.

Enhancements:
- Replace legacy caseworker-specific and outreach-authorization checks with org-type-driven member roles and explicit userPermissions-based access control.
- Simplify permission synchronization and group creation logic, making group-template permission updates template-agnostic and centralizing organization membership operations into service functions.
- Update reporting permissions to use a dedicated ReportOrgPermissions enum instead of piggybacking on UserOrganizationPermissions.

Build:
- Add django-annoying dependency for AutoOneToOneField support in OrganizationProfile.
- Register the annoying app and configure org-type presets and invite templates in Django settings for permission and email behavior.

Deployment:
- Add migrations to create organization profiles, org types, migrate existing org type data, add Shelter Operator permission templates, and adjust related schema.

Documentation:
- Add shelter-operator-specific invite email templates for HTML and text flows.

Tests:
- Update and extend tests to cover org-type-based permission groups, organization member add/remove services, and updated GraphQL fields and query counts.

Chores:
- Remove deprecated shelters permissions modules, caseworker-specific group constants from accounts, and outreach-authorization fields and tests from the user model and schema.


apps/betterangels-backend/accounts/models.py
```diff
@@ -1,8 +1,8 @@
 from typing import Any, Dict, Iterable, Tuple
 
 import pghistory
-from accounts.groups import GroupTemplateNames
 from accounts.managers import UserManager
+from annoying.fields import AutoOneToOneField
```
mikefeldberg, line 5: nice

apps/betterangels-backend/clients/schema.py
```diff
@@ -599,7 +599,13 @@ def create_client_document(self, info: Info, data: CreateClientDocumentInput) ->
                 [ClientProfilePermissions.CHANGE],
             ).get(id=data.client_profile)
 
-            permission_group = get_user_permission_group(user)
+            if data.organization_id:
+                from organizations.models import Organization
```
mikefeldberg: intentional or just claude being claude?

apps/betterangels-backend/accounts/migrations/0046_add_shelter_operator_template.py
```diff
@@ -0,0 +1,154 @@
...
+        "view_schedule",
+        "add_referralrequirement",
+        "change_referralrequirement",
+        "view_referralrequirement",
+        "add_service",
+        "change_service",
+        "view_service",
+        "add_careerservice",
+        "change_careerservice",
+        "view_careerservice",
```
mikefeldberg: there's no `CareerService` model

apps/betterangels-backend/accounts/migrations/0046_add_shelter_operator_template.py
```diff
@@ -0,0 +1,154 @@
...
+        "manage_shelter",
+        "manage_shelter_beds",
+    ]
+
+    org_permissions = Permission.objects.filter(codename__in=org_perm_codenames, content_type=ct)
+    template.permissions.add(*org_permissions)
+
+    # Model-level permissions (on shelter app content types) so that
+    # strawberry_django HasPerm checks pass for non-superusers.
+    shelter_model_codenames = [
```
mikefeldberg, line 50: are all of these perms necessary for operators? I get shelter, bed, room, and contactinfo. But why do they need add + change perms for shared resources? With the exception of `city` and `spa`, all of these are effectively locked, since they're unique and restricted to enums. I'm don't think they even need `view` unless we add direct queries

apps/betterangels-backend/accounts/migrations/0050_populate_org_types.py
```diff
@@ -0,0 +1,44 @@
+from django.db import migrations
+
+
+def migrate_org_types(apps, schema_editor):
```
mikefeldberg: any reason you structured the migrations this way? why not create the `OrgType` and `OrganizationProfile` models in `0047`, and start with an m2m `org_types`?

apps/betterangels-backend/accounts/org_types.py
```diff
@@ -0,0 +1,23 @@
...
+
+def get_invite_templates_for_role(role_name: str) -> dict[str, str]:
+    """Return invite email template paths for a given role name.
+
+    Falls back to the 'default' templates if the role has no specific entry.
+    """
+    return settings.INVITE_TEMPLATES.get(role_name, settings.INVITE_TEMPLATES["default"])
+
+
+def is_member_role(template_name: str) -> bool:
```
mikefeldberg: how is this going to be used? It's checking if the selected template matches an org type role but to what end? The docstring is also a bit confusing because it's not clear what _primary_ means in the context of a role

apps/betterangels-backend/betterangels_backend/settings.py
```diff
@@ -518,3 +519,36 @@
...
+from shelters.groups import SHELTER_OPERATOR  # noqa: E402
+
+# Org-type presets — define what permission group templates and invite behavior
+# are applied when creating an organization with a given preset.
+# Accounts is type-agnostic; it only reads keys from this dict.
+ORG_TYPE_PRESETS = {
+    "outreach": {
+        "label": "Outreach",
+        "templates": [CASEWORKER, GroupTemplateNames.ORG_ADMIN, GroupTemplateNames.ORG_SUPERUSER],
+        "member_role": CASEWORKER,
```
mikefeldberg: does `member_role` effectively mean "default role"? i.e., a basic, non-admin member

mikefeldberg (commented):
just initial comments. still going through it

apps/betterangels-backend/accounts/migrations/0048_seed_org_types_and_profiles.py
```diff
@@ -0,0 +1,28 @@
+from django.db import migrations
+
+
+def seed_org_types_and_backfill_profiles(apps, schema_editor):
```
mikefeldberg, line 4: you can add this runpython in `0047`. unless you want to keep the seed migration separate for tidiness purposes

apps/betterangels-backend/accounts/groups.py
```diff
@@ -2,6 +2,5 @@
 
 
 class GroupTemplateNames(StrEnum):
-    CASEWORKER = "Caseworker"
```
mikefeldberg, line 5: why keep the enum at all? I feel like it adds cognitive load when looking at something like 

```py
register_org_type(
    "outreach",
    label="Outreach",
--> templates=[CASEWORKER, GroupTemplateNames.ORG_ADMIN, GroupTemplateNames.ORG_SUPERUSER],
    default_role=CASEWORKER,
    is_default=True,
)
```

I think we should either get rid of the enum entirely or keep it as is and move it to `common/groups`

apps/betterangels-backend/notes/schema.py
```diff
@@ -85,8 +85,12 @@ class Query:
         permission_classes=[IsAuthenticated],
         extensions=[HasPerm(NotePermissions.ADD)],
     )
-    def caseworker_organizations(self, ordering: Optional[list[OrganizationOrder]] = None) -> "QuerySet[Organization]":
-        return Organization.objects.filter(permission_groups__template__name=CASEWORKER).all()  # type: ignore[no-any-return]
+    def caseworker_organizations(
+        self, info: Info, ordering: Optional[list[OrganizationOrder]] = None
+    ) -> "QuerySet[Organization]":
+        return Organization.objects.filter(  # type: ignore[no-any-return]
```
mikefeldberg, line 91: can remove the ignore with
```
        qs: QuerySet[Organization] = Organization.objects.filter(
            permission_groups__template__name=CASEWORKER,
        )
        return qs
```

or

```
        return cast(QuerySet[Organization], Organization.objects.filter(
            permission_groups__template__name=CASEWORKER,
        ))
```

apps/betterangels-backend/reports/models.py
```diff
@@ -118,6 +118,7 @@ class Meta:
         verbose_name = "Scheduled Report"
         verbose_name_plural = "Scheduled Reports"
         ordering = ["-created_at"]
+        permissions = [("view_reports", "Can view reports")]
```
mikefeldberg, line 121: seems like a good usecase for this util
```suggestion
        permissions = permission_enums_to_django_meta_permissions([ReportOrgPermissions])
```

apps/betterangels-backend/reports/permissions.py
mikefeldberg, line 22: not part of this work but, while we're here, I think `VIEW_REPORTS_PERMISSION` is a better name. This is the only place we use a perm constant like this and `VIEW_REPORTS_PERMISSION` sets a better precedent for when add others

apps/betterangels-backend/reports/migrations/0003_move_view_reports_to_reports_namespace.py
mikefeldberg, line 1: this can be rolled into `0002`

apps/betterangels-backend/accounts/utils.py
```diff
@@ -21,25 +21,26 @@ def remove_organization_permission_group(organization: Organization) -> None:
 
 
 def _get_member_role_for_org(organization: Organization) -> str:
```
mikefeldberg, line 23:
```suggestion
def _get_default_member_role_for_org(organization: Organization) -> str:
```

apps/betterangels-backend/accounts/utils.py
```diff
@@ -198,10 +219,12 @@ def set_role(self, user: User, role: OrgRoleEnum) -> None:
 
         if role == OrgRoleEnum.ADMIN:
             user.groups.add(self._org_admin_group.group)
-
-        if role == OrgRoleEnum.SUPERUSER:
+        elif role == OrgRoleEnum.SUPERUSER:
             user.groups.add(self._org_superuser_group.group)
 
+        # Always ensure the user remains in the default member group
+        add_default_org_permissions_to_user(user, self.organization)
```
mikefeldberg, line 226: this seems overly defensive. Not blocking but we should add a follow-up ticket to update `OrgPermissionManager` and add tests verifying that `clear_permissions` doesn't remove users from the default member group

apps/betterangels-backend/accounts/utils.py
```diff
@@ -1,42 +1,81 @@
...
+        org_type_keys = list(profile.org_types.values_list("key", flat=True))
+    except OrganizationProfile.DoesNotExist:
+        org_type_keys = []
+
+    presets = get_all_presets()
+
+    # Find the first preset that matches one of the org's types
+    for key in org_type_keys:
+        if key in presets:
+            return presets[key].default_role
```
mikefeldberg, line 40: this is non-deterministic for multi-type orgs. Unless I missed something elsewhere, this would result in a lot of unexpected behavior, starting with the invitation email. Seems that (1) in the invite email context, the role should be set by the inviter; and (2) in the perm setting context, this should return a list

PR #2092, opened 2026-05-26
[BP] Add pghistory tracking to bed and reservation models

StockerMC:
Adds pghistory event tracking to Bed and Reservation models so we can query historical status changes for the upcoming shelter reporting dashboard.

Tracks insert, update, delete, and status_change events on both models. The status_change event uses AnyChange("status") so it only fires on status transitions.

## Summary by Sourcery

Add pghistory event tracking for bed and reservation data to support querying historical changes.

Enhancements:
- Track insert, update, delete, and status-change events on Bed and Reservation models using pghistory.

Build:
- Add Django migration creating BedEvent and ReservationEvent models and associated PostgreSQL triggers for history capture.


StockerMC:
> Hey - I've left some high level feedback:
> 
> * For both Bed and Reservation you are tracking a generic `UpdateEvent` plus a `status_change` `UpdateEvent`; consider whether you really need both, since a status change will generate two history rows and may complicate downstream querying if you mostly care about status transitions.
> * In the generated `BedEvent` and `ReservationEvent` models you’ve mirrored `created_at`/`updated_at` from the base models; if those timestamps are meant to represent the original object’s lifecycle rather than the event row’s lifecycle, it might be clearer to drop `auto_now_add`/`auto_now` semantics in favor of plain `DateTimeField` populated solely by the triggers.

Removed the generic update events, since  we only need status_change for the reporting queries.

frank731 (approved):

vecchp (approved):

PR #2094, opened 2026-05-28
search box cleanup (SDB-172)

Davit-BetterAngels:
SDB-172

## Summary by Sourcery

Rework the shelter search experience to use a fullscreen modal and expose active search criteria as removable pills.

New Features:
- Introduce a fullscreen search modal that combines location and name search inputs with a dedicated footer action.
- Display active name and location search criteria as removable pills below the search controls.

Enhancements:
- Simplify the main shelter search bar to a single entry point that opens the new search modal.
- Ensure clearing the location search recenters the map to the user or default location and resets the autocomplete input.


libs/react/shelter/src/lib/components/ShelterSearch/ShelterSearch.tsx
```diff
@@ -120,3 +155,65 @@ export function ShelterSearch(props: TProps) {
...
+      onClear: onClearLocation,
+    });
+  }
+
+  if (!pills.length) return null;
+
+  return (
+    <div className="flex flex-wrap mt-2">
+      {pills.map((pill) => (
+        <div className={mergeCss(pillCss)} key={pill.id}>
```
tglaz: should we use the existing `Pill.tsx` from the react lib?
- it'd need to be adjusted to accept an `onClick` prop and render the Close btn conditionally
- as for styling can just check with Yuko (not sure if the existing Pill style is same or can be used or whether we add a variant or just override the styles here

shiozakiyuko (commented):
Hi @Davit-BetterAngels 

3 things I noticed.

1. I noticed that when I entered keywords in the search box, that shows up in the blue tag like filters. We don't need to show them in the tags so can we remove that? (Screen shot 1)
2. Is it possible to remove the gap between 2 search boxes like shown in the design? (Screen shot 2)
3. I noticed when I searched for example "Wilshire Blvd", and "Shelter 19", I expected to see "Shelter 19" would come up on top of the result. But the result showed 5 unrelated results on top of the Shelter 19. What's happening?

Thanks!

**For 1**
Image

**For 2**
Image

libs/react/shelter/src/lib/components/ShelterSearch/SearchModalContent.tsx
```diff
@@ -0,0 +1,83 @@
...
+        leftIcon={<LocationIcon className="text-neutral-70 w-4 h-4" />}
+      />
+      <div className="mt-2">
+        <Input
+          value={nameValue}
+          placeholder="Search by name"
+          className="w-full"
+          onChange={handleNameChange}
+          onKeyDown={(e) => {
+            if (e.key === 'Enter') {
```
tglaz: Not blocking - but should it do "autocomplete" in the same way as the location search (with debounce and show suggestions)?
* though perhaps ok to defer on this
* if we do add this, may be worth creating a Suggestions or SuggestionList component (or whatever is the standard name) to render the suggestions list. And can use it inside AddressAutocomplete as well.

libs/react/shelter/src/lib/components/ShelterSearch/SearchModalContent.tsx
```diff
@@ -0,0 +1,83 @@
...
+export function SearchModalContent(props: TProps) {
+  const {
+    locationSearchInputKey,
+    initialNameSearchValue,
+    onPlaceSelect,
+    onNameSearchChange,
+    onSearchClick,
+  } = props;
+
+  const [nameValue, setNameValue] = useState(initialNameSearchValue);
```
tglaz, line 25: we're storing `nameValue` in state in 2 places. here and in `ShelterSearch` - can we avoid this?

tglaz (changes requested):

libs/react/shelter/src/lib/components/ShelterSearch/ShelterSearch.tsx
```diff
@@ -66,48 +83,66 @@ export function ShelterSearch(props: TProps) {
...
-          <AddressAutocomplete
-            key={locationSearchInputKey}
-            className="w-full"
-            placeholder="Search by location"
-            onPlaceSelect={onPlaceSelect}
-            leftIcon={<LocationIcon className="text-neutral-70 w-4 h-4" />}
-          />
-          <button onClick={onFilterClick} className="self-start ml-4 mt-4">
-            <FilterIcon className="w-6 text-primary-20" />
+          <button
```
tglaz, line 92: perhaps may may want to set `cursor-pointer` on these - for browsers
- seems like this used to be the default, but not anymore it seems

tglaz (approved):

PR #2097, opened 2026-06-01
shelter combined name + location search (SDB-235)

mikefeldberg:
SDB-235

## Summary by Sourcery

Enhance shelter search to combine name and location queries using Places API viewports and map-bound filters, improving how search results are focused on the map.

New Features:
- Support passing and using Places API recommended viewports for locations throughout the shared places types and React clients.
- Add a map helper component that automatically fits the Google Map to a provided Places viewport when a location search completes.
- Enable the shelter search modal to capture both a selected location and its viewport, applying them when finalizing a combined search.

Enhancements:
- Refine shelter home page search logic to derive map-bound filters from selected locations and optionally preserve existing map bounds when running name searches.
- Extend address autocomplete to request and propagate place viewports into map bounds for better visualization of areas like cities or ZIP codes.
- Introduce reusable utilities for converting between app-level map bounds and Google Maps LatLngBounds and for deriving map bounds from a center point.


shiozakiyuko (commented):
Hey @mikefeldberg 

A couple of things we should make sure of when keywords appear in the map view search box:

- Clearing keywords via the X button should refresh the map and return it to the default view.
- Clearing keywords via the X button should also reset the search criteria in the Location and Name fields.

I also noticed a few related issues. I believe these will resolve once your fix is in, but flagging them for visibility:

- **Map View:** Clearing the keyword in the search box and returning to map view does not reset it to the default/refreshed state.
- **Location Search:** After searching, navigating back to the Search page clears the keyword.
- **Name Search:** After searching, navigating back to the Search page leaves the keyword still populated in the search box.

Let me know what you think.

mikefeldberg:
replaced by #2102

PR #2098, opened 2026-06-01
Refactor permissions to domain-based capabilities (SDB-178)

vecchp:
## Summary

Replaces the flat `userPermissions` array with a structured `permissions` object on organizations, supporting multiple permission domains (accounts, reports, shelters). Moves the `view_reports` permission from the `Organization` to the `ScheduledReport` content type. Introduces a reusable `GrantedPermissions` factory pattern for resolving enum-based permissions via DB annotations.

### Backend

- __Permissions cleanup:__ Removed `TypeAlias` wrappers; `@strawberry.enum` decorator applied directly to `UserOrganizationPermissions`, `ReportPermissions`, and `ShelterPermissions` `TextChoices` classes
- __`GrantedPermissions` factory:__ New reusable pattern in `common/permissions/granted.py` that creates strawberry types with annotation-keyed resolution for any permission domain — eliminates boilerplate for each new domain
- __`OrgPermissions` type:__ New GraphQL type with `accounts`, `reports`, `shelters` fields replacing the old flat `userPermissions` list
- __ReportPermissions migration:__ Moved `view_reports` permission from `Organization` content type to `ScheduledReport` content type (migrations 0002 + 0003), with data migration that swaps the permission in all existing `PermissionGroupTemplate` and `Group` records
- __Meta permissions helper:__ Added `permission_enums_to_django_meta_permissions()` utility; `ScheduledReport` model now declares its permissions via the enum

### Frontend

- __`hasPermission` utility:__ New helper in `activeOrg/hasPermission.ts` with O(1) cached `Set` lookup across all permission domains
- __Bound hook method:__ `useActiveOrg()` returns `hasPermission(perm)` — no need to pass `activeOrg` explicitly
- __Simplified consumers:__ `PermissionGuard`, `AppSidebar`, `Home`, and `Users` page all use the bound `hasPermission` from the hook
- __Type exports:__ `PermissionEnum` union type (`UserOrganizationPermissions | ReportPermissions | ShelterPermissions`) exported for call sites
- __Updated GraphQL queries:__ `UserProvider.graphql` now queries `permissions { accounts reports shelters }` instead of `userPermissions`

### Schema changes

```graphql
type OrgPermissions {
  accounts: [UserOrganizationPermissions!]!
  reports: [ReportPermissions!]!
  shelters: [ShelterPermissions!]!
}
```

`organizationsOrganization` now exposes `permissions` instead of `userPermissions`.

### Test coverage

- __`test_capabilities.py`:__ 11 tests covering annotation key uniqueness, cross-enum collisions, missing attributes, and integration with real project enums
- __`test_views.py`:__ 20 tests covering DRF export (unauthenticated, no-org, no-permission, authorized, CSV with legacy & modern date params, date validation, multi-org authorization) and GraphQL `reportSummary` (auth, empty ranges, org filtering, default dates)

### Verification

- ✅ `ynx-generate` — schema + 5 GraphQL type generation targets pass
- ✅ `ynx-typecheck` — 7 projects, zero errors
- ✅ `yarn nx run betterangels-backend:test` — 818 passed, 31 skipped, 0 failures


apps/betterangels-backend/accounts/tests/test_queries.py
```diff
@@ -224,7 +220,9 @@ def test_logged_in_user_org_permissions_query(
                     firstName
                     organizations: organizationsOrganization {
                         name
-                        userPermissions
+                        capabilities {
```
mikefeldberg: and we're sure we don't want to call this `permissions`? It seems to ma that the only reason to call it something else is to avoid conflict/confusion between actual group permissions. But, since permissions and Capabilities have a 1:1 relationship, I wonder if this will cause more confusion than it prevents. Am I just missing something? Is this a defensive measure against some likely future collision?

apps/betterangels-backend/accounts/capabilities.py
```diff
@@ -0,0 +1,33 @@
...
+class AccountsCapabilities:
+    granted: List[UserOrganizationPermissions]
+
+    @classmethod
+    def get_annotations(cls, user: User) -> dict:
+        """Return DB-level Exists annotations for accounts capabilities."""
+        annotations = {}
+        for perm in UserOrganizationPermissions:
+            app_label, codename = perm.value.split(".")
+            annotations[f"_perm_{perm.name}"] = Exists(
```
mikefeldberg: were these keys supposed to include `app_label`?

apps/betterangels-backend/accounts/types.py
```diff
@@ -98,34 +101,39 @@ def get_queryset(
...
-                    output_field=CharField(),
-                ),
-                filter=Q(
-                    permission_groups__group__user=user,
-                ),
-                distinct=True,
-            )
+            **AccountsCapabilities.get_annotations(user),
+            **ReportsCapabilities.get_annotations(user),
+            **SheltersCapabilities.get_annotations(user),
```
mikefeldberg: per above question, if perms keys don't include app label, so if two apps have the same name perm, whichever comes next overrides. 

Image

apps/betterangels-backend/accounts/capabilities.py
```diff
@@ -0,0 +1,33 @@
...
+from .models import PermissionGroup, User
+from .permissions import UserOrganizationPermissions
+
+
+@strawberry.type
+class AccountsCapabilities:
+    granted: List[UserOrganizationPermissions]
+
+    @classmethod
+    def get_annotations(cls, user: User) -> dict:
```
mikefeldberg: I think this can be abstracted. (This includes the app label in the key) e.g.

```py
# common/permissions/capabilities.py

def _annotation_key(perm: TextChoices) -> str:
    app_label, _ = perm.value.split(".", 1)
    return f"_perm_{app_label}_{perm.name}"


def permission_annotations(user: User, permissions: Type[TextChoices]) -> dict:
    """Return DB-level Exists annotations for organization permission checks."""
    annotations = {}
    for perm in permissions:
        app_label, codename = perm.value.split(".", 1)
        annotations[_annotation_key(perm)] = Exists(
            PermissionGroup.objects.filter(
                organization=OuterRef("pk"),
                group__user=user,
                group__permissions__codename=codename,
                group__permissions__content_type__app_label=app_label,
            )
        )
    return annotations


def granted_permissions(instance: object, permissions: Type[TextChoices]) -> list:
    return [perm for perm in permissions if getattr(instance, _annotation_key(perm), False)]
```

```py
# apps/betterangels-backend/accounts/capabilities.py

@strawberry.type
class AccountsCapabilities:
    granted: List[UserOrganizationPermissions]

    @classmethod
    def get_annotations(cls, user: User) -> dict:
        return permission_annotations(user, UserOrganizationPermissions)

    @classmethod
    def from_instance(cls, instance: object) -> "AccountsCapabilities":
        return cls(granted=granted_permissions(instance, UserOrganizationPermissions))

```

vecchp:
@sourcery-ai review

apps/betterangels-backend/accounts/permissions.py
```diff
@@ -1,66 +1,50 @@
...
+    ``"app_label.codename"`` (e.g. ``ReportPermissions.VIEW_REPORTS``).
     """
-    qs = get_user_permitted_orgs(user, permissions, any_perm=any_perm)
-    if org_id:
-        return qs.filter(pk=org_id).first()
+    qs = get_user_permitted_orgs(user).filter(pk=org_id)
+    if permission:
+        app_label, codename = permission.value.split(".")
+        qs = qs.filter(
+            permission_groups__group__user=user,
```
tglaz: may be able to remove this query? `qs` should already include it?

apps/betterangels-backend/accounts/permissions.py
```diff
@@ -1,66 +1,50 @@
...
-    Otherwise falls back to the first permitted org.
+    *permission* should be a TextChoices enum member whose value is
+    ``"app_label.codename"`` (e.g. ``ReportPermissions.VIEW_REPORTS``).
     """
-    qs = get_user_permitted_orgs(user, permissions, any_perm=any_perm)
-    if org_id:
-        return qs.filter(pk=org_id).first()
+    qs = get_user_permitted_orgs(user).filter(pk=org_id)
+    if permission:
+        app_label, codename = permission.value.split(".")
```
tglaz: this assumes as single `.`, which is probably ok
but could do `split(".", 1)` - think you do elsewhere

I'd possibly suggest making a testable util function such as `_get_perm_keys` where you can abstract the `split` and possibly handle any malformed values.
then `_annotation_key` and other callers would all use it

apps/betterangels-backend/accounts/types.py
```diff
@@ -102,30 +107,29 @@ def get_queryset(
         if not user or not getattr(user, "pk", None):
             return queryset
```
tglaz: I know this has been there for a while, but is it safe to return `queryset` here?
Seems like it'd return everything for an unauthenticated user - just not sure that's intended.

Also, should we just check `user.is_authenticated` here?

apps/betterangels-backend/accounts/permissions.py
```diff
@@ -0,0 +1,81 @@
+from __future__ import annotations
+
+from typing import Any, List, Protocol, Type
+
+import strawberry
+from accounts.models import PermissionGroup, User
```
tglaz, line 6: should `granted.py` live in `common` or `accounts`?
seem like it pulls specific app models and "common" probably wants to avoid that

apps/betterangels-backend/common/permissions/granted.py
```diff
@@ -0,0 +1,81 @@
...
+        )
+    return annotations
+
+
+def granted_permissions(instance: object, permissions: Type[TextChoices]) -> list[TextChoices]:
+    """Return the list of permissions granted on the annotated instance."""
+    return [perm for perm in permissions if getattr(instance, _annotation_key(perm), False)]
+
+
+def GrantedPermissions(permissions_enum: Type[TextChoices]) -> type[GrantedPermissionsType]:
```
tglaz: minor, but would it be more clear to name it`make_granted_permissions` or something that hints at the fact that it's a factory/generator?
Otherwise name looks like a class

apps/betterangels-backend/common/tests/test_capabilities.py
```diff
@@ -0,0 +1,133 @@
...
+        keys = [_annotation_key(p) for p in FakeAppPerms]
+        assert len(keys) == len(set(keys))
+
+    def test_keys_across_multiple_enums_are_unique(self) -> None:
+        """When merging annotations from multiple domains, no collisions occur."""
+        all_keys = [_annotation_key(p) for p in FakeAppPerms] + [_annotation_key(p) for p in AnotherAppPerms]
+        assert len(all_keys) == len(set(all_keys))
+
+
+class TestGrantedPermissions:
```
tglaz, line 68: i'd suggest testing some other truthy/falsy values outside of True/False just to be safe/explicit

apps/betterangels-backend/common/tests/test_capabilities.py
```diff
@@ -0,0 +1,133 @@
...
+
+    def test_from_instance_returns_granted(self) -> None:
+        CapType = GrantedPermissions(FakeAppPerms)
+
+        class FakeInstance:
+            _perm_myapp_view_thing = True
+            _perm_myapp_add_thing = False
+
+        caps = CapType.from_instance(FakeInstance())
+        assert caps.granted == [FakeAppPerms.VIEW]
```
tglaz, line 155: I'd add some unhappy path scenarios

tglaz (changes requested):

tglaz (approved):

PR #2100, opened 2026-06-02
detail page image  latency (SDB-141)

Davit-BetterAngels:
SDB-141

## Summary by Sourcery

Enhancements:
- Use the SHELTER_HERO image preset for shelter hero images in the detail page GraphQL query to improve image delivery.


libs/react/shelter/src/lib/pages/shelter/shelter.graphql
mikefeldberg, line 1: it's not just the hero image. We need to fetch images configured for the components in which they are displayed. I believe those are: carousel, gallery thumb, and gallery full. I'm not sure if it makes sense to pull all three, but we should at least be pulling whatever we want to display in "full size" view. You can but don't have to use presets for this. `url` also takes a `processingOptions` param which parses imgproxy [processing options](https://docs.imgproxy.net/usage/processing)

mikefeldberg (changes requested):

libs/react/shelter/src/lib/pages/shelter/shelter.graphql
shiozakiyuko: Hey @mikefeldberg and @Davit-BetterAngels , should I wait for my side of testing until the fix gets applied to those areas Mike mentioned?

mikefeldberg (approved):

PR #2102, opened 2026-06-02
various search & filter fixes (SDB-235)

mikefeldberg:
this PR addresses a number of bugs and usability issues with shelter search

- nests `/filters` under HomePage to prevent reset on nav (introduced in [#2056](https://github.com/BetterAngelsLA/monorepo/pull/2056))
- moves name + location search from modal to own page under `/search` route (also nested)
  - this wasn't a requirement but [#2094](https://github.com/BetterAngelsLA/monorepo/pull/2094) moved search to a full-screen modal which would likely result in the same user behavior as filters (i.e., hitting "back" to get back to the home page)
- moves search params to atoms for simpler state management and query timing
- adds combined name + location search (used to be one or the other)
- fixes `shelters` query firing on Places autocomplete selection
- updates location search to use Places viewport payload to readjust the Map such that it includes the entirety of the selected location (e.g., an entire zipcode)
- updates sequencing such that shelters within the map bounds are displayed even if they aren't within the bounds of the Place. When user selects a Place and hits "Done":
  - Map repositioned using viewport
  - mapBounds set based on repositioned Map
  - shelters query hit with visible bounds, not Place bounds
- fixes "results based on" string

a few things that were considered but not implemented:
1. updating home page "search" box to display name and location params & function as a "clear params" button 
   1. I initially thought this was just a pressable input but it's actually a button styled to look like an input so this will be done later.
2. persisting location on the search page
   1. when the user enters a search location -> navigates away -> navigates back to search, the location input is cleared
   2. will do in a follow-up
3. resetting the HomePage map to initial state when filters are cleared. 
   1. this behavior seems counterintuitive and detrimental to the user experience. e.g., if I'm looking at a map area and have 3 active filters, when I remove one of them, the results are updated to reflect the removal of that filter. When I remove the second, the same. As a user, I would expect removing the third filter to follow that pattern, _not_ to snap the map back to its initial state. This could be especially frustrating if I'm specifically searching for something within a given area.


apps/shelter-web/src/app/router/buildShelterRoutes.tsx
```diff
@@ -23,12 +23,11 @@ export function buildShelterRoutes() {
...
         element={<HomeRoute />}
-      />
-      <Route
-        key={shelterFiltersPath}
-        path={shelterFiltersPath}
-        element={<FiltersRoute />}
-      />
+      >
+        {/* Nested so HomePage stays mounted when opening overlays, preserving mapBoundsFilter state */}
+        <Route path="filters" element={<FiltersRoute />} />
```
tglaz: would prefer to avoid the path magic strings

libs/react/shelter/src/lib/atoms/searchTriggerAtom.ts
```diff
@@ -0,0 +1,8 @@
+import { atom } from 'jotai';
+
+/**
+ * Incrementing this atom fires a new shelter search.
+ * Consumers freeze their query variables at the moment this value changes
+ * so that intermediate filter-state updates don't trigger spurious queries.
+ */
+export const searchTriggerAtom = atom(0);
```
tglaz: perhaps be specific and call it shelterSearchTriggerAtom?
could also possibly move it to `shelterSearchStateAtom.ts` file as seems related (but ok either way)

libs/react/shelter/src/lib/atoms/shelterSearchStateAtom.ts
```diff
@@ -0,0 +1,24 @@
...
+ * Stored as an atom (not component state) so it updates atomically with
+ * searchTriggerAtom, avoiding a render where the trigger has incremented
+ * but the nameFilter prop hasn't propagated yet.
+ */
+export const shelterNameFilterAtom = atom<string | undefined>(undefined);
+
+export type TSearchSubmission = {
+  nameValue: string;
+  pendingLocation: { location: TLatLng; mapBounds?: TMapBounds } | null;
+};
```
tglaz, line 26: the terminology like "pending" and "submission" seem a bit confusing to me as it seems tied to implementation.

i'd suggest something more like this, but up to you:
- shelterNameSearchInputAtom
- shelterNameFilterAtom
- TShelterSearchRequest (name + location)
- shelterSearchRequestAtom

libs/react/shelter/src/lib/atoms/shelterSearchStateAtom.ts
```diff
@@ -0,0 +1,24 @@
...
+/** The name text currently in the search input (kept in sync so SearchPage pre-populates correctly). */
+export const shelterNameSearchValueAtom = atom('');
+
+/**
+ * The committed name filter for the active query.
+ * Stored as an atom (not component state) so it updates atomically with
+ * searchTriggerAtom, avoiding a render where the trigger has incremented
+ * but the nameFilter prop hasn't propagated yet.
+ */
+export const shelterNameFilterAtom = atom<string | undefined>(undefined);
```
tglaz: I'd suggest using the word `search` instead of `filter` -> `shelterNameSearchAtom`
- it's just a query term and filter vs search is more of an implementation thing

libs/react/shelter/src/lib/components/Map/Map.tsx
```diff
@@ -105,6 +117,10 @@ export function Map(props: TMap) {
       onIdle={() => setShowSearchButton(true)}
       {...cameraProps}
     >
+      <MapPlaceViewportFitter
```
tglaz: this would preferably be a hook, like `useFitMapToPlaceViewport` etc...
using a component that returns null may be ok, but hooks make the intention more clear (otherwise I expect it to render something)

tglaz (changes requested):

apps/shelter-web/src/app/router/buildShelterRoutes.tsx
mikefeldberg: good catch 👍

tglaz (approved):

PR #2104, opened 2026-06-03
add update shelter endpoint (SDB-210)

tglaz:
### Enable edit endpoint for shelter
SDB-210

- move `sheter.services.py` into a `shelter/services` directory
  - shetler, bed, room split into separate files (same for tests)
- add `update_shelter` mutation
- make only `shelter.name` required 
  -  all other fields are optional
  - we may add validations per different `status`, such as `APPROVED` later


apps/betterangels-backend/shelters/services/room.py
```diff
@@ -0,0 +1,50 @@
+from typing import TYPE_CHECKING, Any, Dict
```
Davit-BetterAngels, line 1: _validate_subset_attributes() does not validate occupants; it only checks M2M fields shared with Shelter. Since room.occupants.set(raw_occupants) directly assigns input-provided occupants, this appears to allow associating occupants from another shelter. Please validate that every occupant belongs to the current shelter, and ideally is within the user's authorized organization, before setting the relationship.

apps/betterangels-backend/shelters/services/shelter.py
```diff
@@ -0,0 +1,207 @@
+from typing import TYPE_CHECKING, Any, Dict, List
```
Davit-BetterAngels, line 1: shelter_update() authorizes the user against the shelter’s existing organization, but the update input can also change organization_id without checking access to the target organization. A user with access to Org A could move a shelter into Org B. Please either disallow changing organization in this endpoint or validate membership in the new organization before applying the update.

Davit-BetterAngels (changes requested):

apps/betterangels-backend/shelters/services/room.py
tglaz: I didn't really touch the `bed` or `room` code as these were not in scope (I just moved existing code to a separate `room/bed svc` file).

Having said that, `occupants` are just a `clients.ClientProfile` relation and I don't believe we have any clients that belong strictly to any shelter. So it doesn't seem like it's an actual `subset attribute`.
We may add some sort of validation for this, but I'm not sure if that's been defined yet.

@mikefeldberg does this sound correct?

apps/betterangels-backend/shelters/services/shelter.py
tglaz: i popped the `organization` key off of data in shelter_update (shelter svc) just to be safe.
but `org` is not part of `UpdateShelterInput` so this should already not be permitted.

Davit-BetterAngels (approved):

apps/betterangels-backend/shelters/services/room.py
mikefeldberg: short answer: safe to ignore because I'm in the process of removing occupants from `Bed` and `Room`

long answer: this validation fn was put in place to ensure that a Bed with demo x couldn't be added to a Shelter with demo y. It's a temporary solution until we resolve the larger question of: does it even make sense to have properties on `Bed`, `Room`, and `Shelter`, or if they should only live on, say, `Bed`, and bubble up to `Room` and `Shelter`. Strictly speaking, this validation _could_ be applied to occupants, but it shouldn't cus occupants will definitely bubble up.

PR #2105, opened 2026-06-03
persist location input (SDB-235)

mikefeldberg:
SDB-235

persist location input on search page. clear on map area search

### DIFF
--- a/libs/react/shelter/src/lib/atoms/index.ts
+++ b/libs/react/shelter/src/lib/atoms/index.ts
@@ -1,9 +1,14 @@
 export { nullShelterPropertyFilters, shelterPropertyFiltersAtom } from './shelterPropertyFiltersAtom';
 export {
+  shelterLocationSearchInputAtom,
   shelterNameSearchAtom,
   shelterNameSearchInputAtom,
+  shelterSearchPendingLocationAtom,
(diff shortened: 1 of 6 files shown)


tglaz (approved):

shiozakiyuko (commented):
Hey @mikefeldberg 

I noticed a behavior while testing and wanted to check if this is expected or if I should open a separate ticket for it.

**Steps to reproduce**:
1. Started on the map with 84 pins on the Westside (note: "current location" didn't seem to be working).
2. Applied the Single Men and Open Now filters $\rightarrow$ results correctly narrowed down to 21 pins.
3. Tapped on a shelter to view its details page, then navigated back to the map.

**Expected behavior**:  Based on the understanding from our convo yesterday, I expected to see the filtered list of 21 pins again.

**Actual behavior:**  The map automatically re-focused on the specific shelter I just viewed (and its immediate neighborhood), reducing the visible pins to just 2. This center/focus shifts depending on which shelter was last tapped.

Is this the intended navigation behavior, or is the map state resetting incorrectly on back-navigation?

Here is the screen recoring:
https://github.com/user-attachments/assets/3368c04c-745c-4d88-9bd9-93554d62c165



Otherwise, it looks good to me. :)

PR #2106, opened 2026-06-03
[BP] Add bed status reporting selector

Angus-Sun:
Added a bed status counts selector for the shelter reporting dashboard. Takes a shelter, start date, and end date and returns daily counts broken down by status (available, occupied, reserved, out_of_service) by replaying bed event history.

Left demographic filtering as a TODO, as pghistory only tracks scalar fields, so historical M2M membership can't be reconstructed at the moment.

Includes 6 unit tests covering single day counts, status changes, removed beds, cross-shelter isolation, date ranges, and beds created before the query window.


frank731 (approved):
LGTM, main point to note for BA reviewers if whether https://github.com/BetterAngelsLA/monorepo/pull/2106/changes#diff-a75dd1d8f92ec6e87a1730987bdb6e1e7477a1e6418dbbe22998f896e08d61b1R33 is the correct place to be importing BedEvent given that it needs the `# type: ignore[attr-defined]` tag

vecchp (approved):

PR #2112, opened 2026-06-05
[BP] add delete reservation mutation

emilyqqian:
Added a deleteReservation GraphQL mutation so reservations can be removed from the system, only for authenticated users with the delete permission. 

Added a ReservationPermissions class, a ReservationType output type, and a reservation_delete service function

Includes unit tests confirming the record is removed from the database after the mutation is called. 

GraphQL schema and frontend types were also regenerated.

## Summary by Sourcery

Add a GraphQL mutation and supporting types and permissions to allow deleting reservations with appropriate authorization checks.

New Features:
- Introduce a deleteReservation GraphQL mutation for removing reservations.
- Add a ReservationType GraphQL output type to represent reservation deletions.
- Define ReservationPermissions to manage reservation-related authorization.

Tests:
- Add GraphQL tests verifying reservations are deleted when permitted and preserved when permissions are missing.


apps/betterangels-backend/shelters/types/outputs.py
```diff
@@ -292,6 +292,11 @@ def get_queryset(cls, queryset: QuerySet, info: Info) -> QuerySet[models.Bed]:
     type: Optional[BedTypeChoices]
 
 
+@strawberry_django.type(models.Reservation)
+class ReservationType:
+    id: ID
```
frank731, line 299: could either write a note saying only delete implemented so far or implement the whole type like how is done here https://github.com/BetterAngelsLA/monorepo/pull/1979/changes#diff-341744ea935de5bb91fc17bdccfdde0221d6cd18e19e05b960a34ab215c5a58bR218

frank731 (approved):

frank731:
LGTM, only note is whether @strawberry_django.mutation(permission_classes=[IsAuthenticated]) should be used [here](https://github.com/BetterAngelsLA/monorepo/pull/2112/changes#diff-3f0be08684f844401301929b85ec824df39a2aa8934363413763dd8cb5803406R101) to align with the other mutations. I see both methods used in the codebase for deletion mutations, so either should be fine, can leave it up to BA to decide which format is better.

vecchp (changes requested):
Thanks for the work on this! The permission approach and tests need a bit of discussion before merging.

### How the current permission model works

The mutation uses `HasRetvalPerm`. Under the hood, Django guardian's `filter_for_user` works like this: if a user has the Django model-level permission (e.g. granted via a group), they get **all** objects back, so `HasRetvalPerm` passes for any reservation regardless of org. Guardian object-level rows are only checked when the user *doesn't* have the model-level perm.

This means the tests pass (they grant the model perm), but the actual behavior is: **any user with `delete_reservation` can delete any reservation across all organizations**. There's no org scoping anywhere.

### We should land `createReservation` first

This is ultimately a design question about the permission model, and the create mutation is what will lock us into a path:

- If create assigns guardian permissions via `assign_object_permissions()`, then `HasRetvalPerm` on delete makes sense. But the tests will need to be updated to use guardian object-level permissions (via `assign_perm` or `assign_object_permissions`), not model-level `user_permissions.add()`.
- If create relies on queryset-level org scoping (like `createShelter`/`createBed`), then delete should too: switch to `HasPerm` and add a `get_queryset` that filters by org.

Landing create first ensures delete's permission model matches the system's actual design rather than guessing. Either way, `createReservation` tells us which path we're on.

PR #2113, opened 2026-06-05
[BP] Daily occupancy reporting selector

StockerMC:
## Description

Adds `daily_occupancy` to `shelters/selectors/reports.py`: per-day occupied-bed counts and occupancy percentage for a shelter over a date range, returned as `list[DailyOccupancyMetricsType]`.

## What changed

- Occupancy is reconstructed from the `Reservation` event stream. A bed is occupied on an LA-local day when a reconstructed `CHECKED_IN` interval overlaps it (`_reservation_occupancy_intervals`). This is correct across cancellations, overdue, bed reassignment, and since-deleted beds, and doesn't depend on the removed `Bed.status`.
- `total_beds` is the number of beds that existed that day, from `bed.add` / `bed.remove`. `occupancy_pct` is `occupied / total * 100` rounded to two places.
- Uses the shared `report_date_range_to_utc` (from #2227) for LA-local day boundaries. Adds `report_day_bounds_utc` and `iter_report_days` alongside it.
- Takes `shelter` + `start_date` + `end_date` only. No filters: demographic and bed filtering are out of scope per design.

Selector + tests only. No GraphQL wiring, no migrations.

## Notes

- Builds on the shared helper merged in #2227.
- `_reservation_occupancy_intervals` is shared with the days-to-occupancy selector (#2114), which stacks on this branch.


apps/betterangels-backend/shelters/selectors.py
```diff
@@ -183,3 +187,322 @@ def shelters_open_at(
...
+
+
+def _as_list(value: Any) -> list:
+    """Normalize a scalar-or-iterable filter value into a list."""
+    if isinstance(value, (list, tuple, set)):
+        return list(value)
+    return [value]
+
+
+def _report_date_range_to_utc(
```
frank731: [Might just have selectors assume utc](https://github.com/BetterAngelsLA/monorepo/pull/2114#discussion_r3399559264)

apps/betterangels-backend/shelters/selectors.py
```diff
@@ -183,3 +187,322 @@ def shelters_open_at(
...
+    results = []
+    for day in _iter_days(start_date, end_date):
+        # End-of-day boundary (exclusive) in UTC — also used to read client state.
+        day_bound = datetime.datetime.combine(
+            day + datetime.timedelta(days=1), datetime.time.min, tzinfo=LA_TZ
+        ).astimezone(datetime.timezone.utc)
+        beds = snapshots.get(day, {})
+        total_beds = 0
+        occupied_count = 0
+        for status, occupant_id in beds.values():
```
frank731: Should we batch query on the occupant ids by prefetching all of them and then running a single query instead of calling client_matcher individually on each making                 client_event_model.objects.filter(pgh_obj_id=occupant_id) run for each client? Could cut down query count to constant

apps/betterangels-backend/shelters/selectors.py
```diff
@@ -183,3 +187,354 @@ def shelters_open_at(
...
+                client_matcher is None or client_matcher(occupant_id, day_bound)
+            ):
+                occupied_count += 1
+        occupancy_pct = round(occupied_count / total_beds * 100, 2) if total_beds else 0.0
+        results.append(
+            {
+                "date": day,
+                "occupied_count": occupied_count,
+                "total_beds": total_beds,
+                "occupancy_pct": occupancy_pct,
```
frank731: should we be returning this or should we assume this should just be calculated using occupied count and total beds afterwards just to return less stuff

apps/betterangels-backend/shelters/selectors.py
StockerMC: Good call, the matcher now collects all occupant ids up front and bulk-loads their history in one query instead of one per occupant, so it's constant regardless of occupant count. Also added a query-count test to assert this

apps/betterangels-backend/shelters/selectors.py
StockerMC: Good point, the percentage is fully derivable from occupied_count and total_beds, so I'm fine not returning it from the selector and computing it at the type layer instead.

The one thing I want to avoid is the rounding and the divide-by-zero rule (return 0 when total_beds is 0) getting reimplemented in a few different places and drifting. So my plan would be to pull that into a small shared helper and have the GraphQL type resolve occupancyPct through it, which keeps the return clean like you're suggesting and keeps the calculation in one spot.

If daily_occupancy ends up being used only through GraphQL, we might not even need the helper, just a resolved field on the type. Let me know which way you want the types to go

apps/betterangels-backend/shelters/selectors.py
StockerMC: That works for the selectors that just filter a datetime range (those can stay UTC). This one is different because it groups results into LA-local days, so it needs a timezone to know where each day starts and ends, including DST days that aren't a clean 24 hours. So I don't think the timezone can't really move out, it can only be a param (default LA) if you want it consistent with the others. Happy to do that if you prefer

apps/betterangels-backend/shelters/selectors.py
frank731: honestly for simplicity we could just keep it here to handle those rounding/zero div issues, having one extra field isn't horrible there are a couple places (exports and charts) where it would be used, could just keep it tbh

apps/betterangels-backend/shelters/selectors.py
StockerMC: Keeping this LA-local on purpose. The reporting UI (#2177) hands us a date range, not timestamps. It's a date picker with presets like "Last 30 days" and dates shown as YYYY-MM-DD. Since daily occupancy and bed status are per-day numbers, we have to pick a timezone for where the day starts and ends. If we go by UTC the day flips over around 4-5pm LA time, so an evening check-in gets counted on the next day, which isn't what we want on these charts. Every BA shelter is in LA so that's the right anchor, and it's noted in the module. If we ever get a shelter outside LA we'd add a timezone field to the shelter rather than switch to UTC.

For keeping things consistent, I'd rather move report_bed_status_counts (which is UTC right now) over to this same LA-local boundary in a separate PR so both per-day charts line up, instead of making occupancy use UTC. reservation_status_change_counts just takes datetimes and counts over a range so it's fine as is. Can file that follow-up if we want it.

vecchp (approved):

PR #2114, opened 2026-06-05
[BP] Average days-to-occupancy reporting selector

StockerMC:
## Description

Adds `avg_days_to_occupancy` to `shelters/selectors/reports.py`: the mean number of days a bed sat empty before becoming occupied over a date range, or `None` when there are no qualifying pairs.

## What changed

- Reuses the shared `_reservation_occupancy_intervals` reconstruction. For each in-range check-in that is not a bed's first occupancy, measures the gap in days from the previous interval's freed instant to this check-in.
- A bed becomes free on any exit from `CHECKED_IN` (checkout or cancellation), so cancelled stays are measured correctly, not skipped. Beds occupied from creation (no prior interval) and reverts on the same reservation are excluded. The prior free may pre-date the range.
- Uses `report_date_range_to_utc` (from #2227) for LA-local boundaries. `shelter` + dates only, no filters.

Selector + tests only. No GraphQL wiring, no migrations.

## Stacking

Based on #2113 (daily occupancy) so it can share `_reservation_occupancy_intervals`. Review/merge #2113 first; this retargets to main once it lands.

### DIFF
--- a/apps/betterangels-backend/shelters/selectors/__init__.py
+++ b/apps/betterangels-backend/shelters/selectors/__init__.py
@@ -13,6 +13,7 @@ from shelters.selectors.operator import (
     shelter_queryset,
     user_shelter_list,
 )
+from shelters.selectors.reports import avg_days_to_occupancy  # noqa: F401
 from shelters.selectors.reports import daily_occupancy  # noqa: F401
 from shelters.selectors.reports import report_bed_status_counts  # noqa: F401
 from shelters.selectors.reports import reservation_status_change_counts
--- a/apps/betterangels-backend/shelters/selectors/reports.py
+++ b/apps/betterangels-backend/shelters/selectors/reports.py
@@ -7,8 +7,8 @@ from typing import TYPE_CHECKING, Any, Iterable, TypedDict
 from zoneinfo import ZoneInfo
 
 import pghistory
-from django.db.models import Count, OuterRef, Q, Subquery, TextField
-from django.db.models.functions import Cast
+from django.db.models import Avg, Count, F, OuterRef, Q, Subquery, TextField
+from django.db.models.functions import Cast, Extract
 from shelters.enums import BedStatusChoices
 
 if TYPE_CHECKING:
@@ -370,13 +370,13 @@ def avg_days_to_occupancy(
 ) -> float | None:
     """Average number of days a bed sat unoccupied before becoming occupied.
 
-    For every occupancy interval whose check-in falls in the LA-local range and
-    that is not the bed's first interval, the vacancy is the gap in days from the
-    previous interval's freed instant to this check-in. A bed becomes free on any
-    exit from CHECKED_IN (checkout or cancellation), so cancelled stays are
-    measured correctly. Beds occupied from creation (no prior interval) and
-    reverts on the same reservation are excluded. Returns the mean rounded to two
-    places, or None when there are no qualifying pairs.
+    For every check-in event (status changed TO checked_in) whose timestamp
+    falls in the LA-local range, the vacancy is the gap in days from the most
+    recent free event (any exit from CHECKED_IN, including cancellations) on the
+    same bed from a *different* reservation.  Beds occupied from creation (no
+    preceding free event) and reverts on the same reservation are excluded.
+    Returns the mean rounded to two places, or None when there are no qualifying
+    pairs.
 
     Both endpoints are inclusive and interpreted in America/Los_Angeles.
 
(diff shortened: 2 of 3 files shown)


apps/betterangels-backend/shelters/selectors.py
```diff
@@ -113,3 +118,262 @@ def shelters_open_at(
...
+    Raises:
+        ``ValueError`` if ``start_date`` is after ``end_date`` or the range
+        exceeds :data:`MAX_REPORT_RANGE_DAYS`.
+    """
+    if start_date > end_date:
+        raise ValueError("start_date must be on or before end_date")
+    if (end_date - start_date).days + 1 > MAX_REPORT_RANGE_DAYS:
+        raise ValueError(f"Date range may not exceed {MAX_REPORT_RANGE_DAYS} days")
+
+    start_local = datetime.datetime.combine(start_date, datetime.time.min, tzinfo=LA_TZ)
```
frank731: Should we be converting timezones here or in some preprocessing outside? I feel like this logic will likely need to be reused outside so prob can just assume everything is UTC or smth like that

apps/betterangels-backend/shelters/selectors.py
```diff
@@ -113,3 +118,262 @@ def shelters_open_at(
...
+            only qualifies if the occupant matched (as of the occupancy moment)
+            (see :func:`_build_client_demographic_matcher`).
+
+    Returns:
+        The mean number of days (rounded to two places) or ``None`` when there
+        are no qualifying pairs.
+    """
+    start_utc, end_utc = _report_date_range_to_utc(start_date, end_date)
+
+    bed_event_model = apps.get_model("shelters", "BedEvent")
```
frank731: Can this be imported like `from shelters.models import BedEvent`

apps/betterangels-backend/shelters/selectors.py
```diff
@@ -113,3 +118,262 @@ def shelters_open_at(
...
+    event_qs = bed_event_model.objects.filter(shelter_id=shelter_id, pgh_created_at__lt=end_utc)
```
frank731: Should we be filtering on the pgh_created_at__gte=start_utc as well

frank731 (commented):
Looks good, couple of suggestions, main thing is how we should handle timezones

apps/betterangels-backend/shelters/selectors.py
StockerMC: `BedEvent` is pghistory-generated, so a direct import needs `# type: ignore[attr-defined]`. `apps.get_model` avoids that (same as `ClientProfileEvent` below). Can switch if you'd prefer.

apps/betterangels-backend/shelters/selectors.py
StockerMC: The preceding non-occupied event we measure back to usually pre-dates `start_utc`, so a lower bound would drop it. Occupancy event is still range-bound in the loop. Added a comment to clarify.

apps/betterangels-backend/shelters/selectors.py
StockerMC: Converting it here on purpose: the dates are LA-local but we filter on `pgh_created_at` (UTC), so the range has to be converted to a UTC window. It's in the shared `_report_date_range_to_utc` helper, so reuse goes through that instead of each caller redoing it, and taking dates means no one can query in the wrong tz.

frank731: I think some of the other selectors that were made just assume UTC, could have it be like it takes in a UTC timezone and when selectors are run, the time chosen in the frontend that gets passed to the selectors can be converted in one place there and then passed to all the selectors

StockerMC: Keeping this LA-local on purpose. The reporting UI (#2177) hands us a date range, not timestamps. It's a date picker with presets like "Last 30 days" and dates shown as YYYY-MM-DD. Since daily occupancy and bed status are per-day numbers, we have to pick a timezone for where the day starts and ends. If we go by UTC the day flips over around 4-5pm LA time, so an evening check-in gets counted on the next day, which isn't what we want on these charts. Every BA shelter is in LA so that's the right anchor, and it's noted in the module. If we ever get a shelter outside LA we'd add a timezone field to the shelter rather than switch to UTC.

For keeping things consistent, I'd rather move report_bed_status_counts (which is UTC right now) over to this same LA-local boundary in a separate PR so both per-day charts line up, instead of making occupancy use UTC. reservation_status_change_counts just takes datetimes and counts over a range so it's fine as is. Can file that follow-up if we want it.

vecchp (approved):

vecchp (approved):

PR #2115, opened 2026-06-05
[BP] - Reservation status selector

c4bae:
Creates a selector `reservation_status_change_counts` that returns the number of reservation status changes within a given time period, specifically:

- Total reservations that changed status to `CHECK_IN_OVERDUE`
- Total reservations that changed status to `CANCELLED`
- Total reservations that changed status to `CHECKED_IN`
- Total reservations that changed status from `CHECK_IN_OVERDUE` to `CHECKED_IN`

Additionally implemented unit tests to verify correct functionality.

## Summary by Sourcery

Add a selector to report reservation status change counts over a date range and cover it with unit tests.

New Features:
- Introduce a reservation_status_change_counts selector that aggregates reservation status transitions for a shelter within a given date range.

Tests:
- Add comprehensive unit tests covering per-status counting, overdue-to-checked-in transitions, date boundaries, shelter scoping, lookback limits, and empty-result behavior for the new selector.


apps/betterangels-backend/shelters/selectors.py
```diff
@@ -113,3 +116,49 @@ def shelters_open_at(
...
+    Count how many reservations changed to given statuses for a shelter in a date range (end_date inclusive).
+
+    Each reservation is counted once per status.
+    """
+    now = timezone.now()
+    furthest_date = now - datetime.timedelta(days=365)
+    if start_date < furthest_date.date() or end_date < furthest_date.date():
+        return {}
+
+    from shelters.models import Reservation
```
frank731, line 224: Should this be using the ReservationEvent instead (from shelters.models import ReservationEvent)? Can filter on that and should be cleaner (smth like events = ReservationEvent.objects.filter(...) )

apps/betterangels-backend/shelters/selectors.py
```diff
@@ -113,3 +116,49 @@ def shelters_open_at(
...
+    )
+
+    events = pghistory.models.Events.objects.filter(
+        pgh_obj_id__in=reservation_ids,
+        pgh_label="reservation.status_change",
+        pgh_created_at__gte=start_date,
+        pgh_created_at__lt=end_date + datetime.timedelta(days=1),
+    )
+
+    return {
```
frank731: Probably would be more useful to return a list of changes per day so that we can show as a table rather than a single row lol

frank731 (changes requested):

apps/betterangels-backend/shelters/selectors.py
c4bae: I originally used `ReservationEvent`, but I changed to using the general pghistory event model because it had `pgh_diff` which allowed me to easily compare status changes needed for `OVERDUE` to `CHECKED_IN`

apps/betterangels-backend/shelters/selectors.py
```diff
@@ -113,3 +116,74 @@ def shelters_open_at(
...
+    shelter_id: int,
+    start_date: datetime.date,
+    end_date: datetime.date,
+) -> dict[str, int]:
+    """
+    Count how many reservations changed to given statuses for a shelter in a date range (end_date inclusive).
+
+    Each reservation is counted once per status.
+    """
+    now = timezone.now()
```
frank731: I think we should assume UTC for selectors, prob will do something like timezone conversion before running selector so we can have a shared thing to do the timezone conversion instead of implementing it for each, not sure if that changes anything here

apps/betterangels-backend/shelters/selectors.py
```diff
@@ -113,3 +116,74 @@ def shelters_open_at(
...
+) -> dict[str, dict[str, int]]:
+    """
+    Count daily reservation status changes for a shelter in a date range.
+
+    Each reservation is counted once per status.
+    """
+    results = {}
+
+    current_date = start_date
+    while current_date <= end_date:
```
frank731: Instead of doing loop which might need a lot of db queries prob should fetch using a single thing using smth like TruncDate to aggregate all at once

from django.db.models.functions import TruncDate

apps/betterangels-backend/shelters/selectors.py
```diff
@@ -113,3 +116,74 @@ def shelters_open_at(
...
+    events = pghistory.models.Events.objects.filter(
+        pgh_obj_id__in=reservation_ids,
+        pgh_label="reservation.status_change",
+        pgh_created_at__gte=start_date,
+        pgh_created_at__lt=end_date + datetime.timedelta(days=1),
+    )
+
+    return {
+        "STATUS_TO_CHECK_IN_OVERDUE": events.filter(pgh_data__status="check_in_overdue")
+        .values("pgh_obj_id")
```
frank731: This could be optimized into a single query with smth like 

STATUS_TO_CHECK_IN_OVERDUE=Count(
  │           "pgh_obj_id",
  │           distinct=True,
  │           filter=Q(pgh_data__status="check_in_overdue"),
  │       ),

frank731 (changes requested):

frank731 (approved):

PR #2116, opened 2026-06-07
[BP] Add update reservation mutations

Angus-Sun:
## Summary

Adds two new mutations to the shelters GraphQL schema:

- **`updateReservation`** — updates scalar fields on a reservation (notes, duration, dates, room/bed assignment) using patch semantics (omitted fields are left unchanged)
- **`updateReservationStatus`** — updates only the status of a reservation

## Details

Both mutations:
- Require authentication and the `change_reservation` permission
- Restrict access to reservations belonging to shelters in the user's organization
- Return `OperationInfo` on error (nonexistent ID, wrong org)

## Tests

Coverage includes:
- Field updates
- Patch semantics
- Status updates
- Nonexistent IDs
- Nullable field clearing
- Cross-org access rejection


apps/betterangels-backend/shelters/services/reservation.py
julianchen24, line 1: Right now, we're only updating the status of reservation. However, as the ticket notes:

"- updateReservationStatus
    - When this runs, we need to update the status of the clients to show that they are reserved (should be a `status` field attached to them). We also need to list the beds/rooms as reserved as well, should have a status attr as well."

We want to update both the beds and rooms as required. I looked into it and BedStatusChoices.RESERVED and RoomStatusChoices.RESERVED statuses should already exist so you can just use those.

Looking at the ticket, when the status changes we also need to sync the linked bed and room statuses

frank731: NGL we probably can make a smaller client field update to just add the status field, we can add the rest as needed, should be less of a worry for BA then

apps/betterangels-backend/shelters/types/inputs.py
```diff
@@ -258,3 +259,21 @@ class UpdateRoomInput:
     storage: Maybe[bool]
     type: Maybe[RoomStyleChoices | None]
     type_other: Maybe[str | None]
+
+
+@strawberry.input
+class UpdateReservationInput:
+    reservation_id: ID
```
julianchen24, line 266: Referring to other parts of the codebase, we don't normally include the reservation_id in the UpdateReservationInput class. we normally include it in the resolver. I got claude to give you an example of what I mean:


Example:
How updateBed does it (the existing pattern)
The mutation resolver — id is a separate parameter, next to data:


def update_bed(self, info: Info, id: ID, data: UpdateBedInput) -> BedType:
    clean = strawberry.asdict(data)
    return cast(BedType, bed_update(user=user, bed_id=id, data=clean))
The input — notice UpdateBedInput has no id / bed_id field. It's only the fields you're editing:


class UpdateBedInput:
    room_id: Maybe[ID | None]
    accessibility: ...
    fees: ...
    # no bed_id here
So the ID rides on the outside (update_bed(..., id, data)), and data is purely the stuff being changed.

How this PR does it (the deviation)
The ID is inside the input:


class UpdateReservationInput:
    reservation_id: ID          # ← lives inside the data object
    notes: Maybe[Optional[str]] = UNSET
    duration: ...

def update_reservation(self, info, data: UpdateReservationInput) -> ReservationType:
    clean = strawberry.asdict(data)
    return reservation_update(user=user, data=clean)   # id is buried in `clean`
…and then the service has to dig it back out:


reservation_id = data.pop("reservation_id")   # extra step the others don't need




You can refer to this: apps/betterangels-backend/shelters/types/inputs.py:208 (class UpdateBedInput)

apps/betterangels-backend/shelters/services/reservation.py
julianchen24: Adding to this:

"When this runs, we need to update the status of the clients to show that they are reserved (should be a status field attached to them). "

The ticket asks us to update the client status field, but this hasn't been implemented yet. However, I don't think that the client model even has a status field to begin with. I was working on updating the client model at the start of the term but then BA devs asked us to pause on making any changes to the client model so we we need to confirm with Frank on how to continue. We can ask at the end of the meeting

apps/betterangels-backend/shelters/services/reservation.py
```diff
@@ -0,0 +1,44 @@
...
+if TYPE_CHECKING:
+    from accounts.models import User
+
+
+@transaction.atomic
+def reservation_update(*, user: "User", data: Dict[str, Any]) -> Reservation:
+    data = dict(data)
+    reservation_id = data.pop("reservation_id")
+
+    try:
```
julianchen24, line 18: Nit:
This fetches with Reservation.objects.get(...) and then does a separate shelter_get(...). We already have a selector pattern for this: bed_get in selectors.py and bed_update (services/bed.py) which just calls it.

Could we add a reservation_get to selectors.py mirroring bed_get, and use it here?

selectors.py for reference:

def admin_reservation_list(qs, *, user):
    return qs.filter(shelter__in=admin_shelter_list(Shelter.objects.all(), user=user))

def reservation_get(*, user, reservation_id):
    return admin_reservation_list(Reservation.objects.select_related("shelter"), user=user).get(pk=reservation_id)

Then both functions here collapse to one line:
reservation = reservation_get(user=user, reservation_id=reservation_id)

The reason is because objects.get() just grabs the row, it doesn't check who's allowed to see it, so the code has to add shelter_get() right after. reservation_get only searches within the user's own org, so a wrong-org id just comes back as "not found", fetch and permission check in one call. Keeps the service simple and matches how beds and rooms already work.

Reference: bed_get + admin_bed_list in apps/betterangels-backend/shelters/selectors.py, and bed_update in apps/betterangels-backend/shelters/services/bed.py.

apps/betterangels-backend/shelters/services/reservation.py
```diff
@@ -0,0 +1,44 @@
...
+        if value is not UNSET:
+            setattr(reservation, field, value)
+
+    reservation.full_clean()
+    reservation.save()
+    return reservation
+
+
+@transaction.atomic
+def reservation_update_status(*, user: "User", reservation_id: Any, status: Any) -> Reservation:
```
julianchen24, line 34: Minor nit: could reservation_id/status be typed (int | str / ReservationStatusChoices) instead of Any, to match bed_update/room_update?

apps/betterangels-backend/shelters/tests/test_mutations.py
julianchen24, line 1: Missing tests for updateReservationStatus, might want to add some

apps/betterangels-backend/shelters/services/reservation.py
```diff
@@ -0,0 +1,44 @@
...
+    try:
```
julianchen24, line 35: reminder to update this as well when updating reservation_update

PR #2125, opened 2026-06-09
unknown for filters (SDB-156)

Davit-BetterAngels:
SDB-156


apps/betterangels-backend/shelters/types/filters.py
```diff
@@ -126,10 +134,33 @@ def properties(
...
+            "shelter_types",
+            "room_styles",
+            "parking",
+        ]
+
         value_dict = asdict(value)
-        filters = {f"{k}__name__in": v for k, v in value_dict.items() if v is not None}
+        combined_q = Q()
+
+        for field in property_fields:
```
mikefeldberg, line 155: please add a test for include_null

libs/react/shelter/src/lib/components/ShelterFilters/config.ts
```diff
@@ -28,13 +30,19 @@ export type TFilterOptionType =
...
-  | SpecialSituationRestrictionChoices;
+  | SpecialSituationRestrictionChoices
+  | typeof UNKNOWN_FILTER_VALUE;
 
 export type TShelterFilterOption = {
   label: string;
   value: TFilterOptionType;
 };
 
+const unknownOption: TShelterFilterOption = {
```
mikefeldberg, line 41: this is more of a product comment but, since "Unknown" isn't a property, I don't think we should display it the same way we display properties. I thought we wanted it to look more like Max Stay. 
Image

mikefeldberg (changes requested):

libs/react/shelter/src/lib/components/ShelterFilters/config.ts
shiozakiyuko: Hey @Davit-BetterAngels,

Agreed, Mike makes a valid point. 'Unknown' represents missing data or items that haven't been researched yet, so conceptually it shouldn't sit in the same group as the successfully researched options. Let's move it outside of the checkbox group, similar to how 'Max Stay' is positioned. Thanks.

image

mikefeldberg: @shiozakiyuko @Davit-BetterAngels lets do `"Include unknown"` rather than `"Include: Unknown"`. Also think we should make the checkboxes line up vertically

mikefeldberg (approved):

PR #2134, opened 2026-06-10
shelter operator room and bed backend

mikefeldberg:
SDB-226, SDB-227, SDB-228, SDB-229, SDB-230, SDB-231


vecchp (commented):
## Summary

This PR adds the missing CRUD operations for Bed and Room — update, clone, bulk-delete, and single-entity queries. The architecture follows the established project pattern (thin schema → services → selectors) and has thorough test coverage.

## Overall Verdict

✅ Approve with minor changes requested. Two bugs should be fixed before merge (items 1 and 3 below). The remaining items are performance improvements and consistency fixes.

| # | File | Severity | Issue |
|---|---|---|---|
| 1 | bed.py | 🔴 Bug | bed_update double-queries bed_get |
| 2 | bed.py, room.py | 🔴 Perf | Bulk delete uses N individual .delete() instead of QuerySet.delete() |
| 3 | test utils.py | 🟡 Bug | Duplicate add_room in SHELTER_OPERATOR_PERMISSIONS |
| 4 | outputs.py | 🟡 Arch | BedType/RoomType.get_queryset inline instead of delegating to selectors |
| 5 | bed.py, room.py | 🟡 Perf | Clone doesn't prefetch_related M2M on source |
| 6 | bed.py, room.py | 🟡 Perf | Create doesn't prefetch shelter M2M before _validate_subset_attributes |
| 7 | bed.py, room.py | 🟢 Style | Duplicate _clone_label across services |
| 8 | bed.py | 🟢 Doc | bed_delete docstring says rooms |

## What's Done Well

- Full CRUD coverage with proper org-scoping at every layer
- @transaction.atomic on all writes, full_clean() before save()
- Patch semantics (None values skipped in updates)
- Clone correctly copies M2M; room clone intentionally doesn't copy beds
- Reusable BulkDeleteInput/BulkDeleteResult in common types
- Thorough tests with query-count assertions and org-isolation checks
- Follows Django-Styleguide service/selector separation
- Permission gating via @hasPerm on all operations

## Notes

For full inline comments, see the individual review threads. The must-fix items are the duplicate bed_get() in bed_update (line 64/67) and the duplicate add_room in SHELTER_OPERATOR_PERMISSIONS (test utils line 18-19).

apps/betterangels-backend/shelters/services/bed.py
```diff
@@ -45,4 +41,104 @@ def bed_create(*, user: "User", data: Dict[str, Any]) -> Bed:
...
+        ``django.core.exceptions.ValidationError`` on invalid data.
+    """
+    data = dict(data)
+    try:
+        bed = bed_get(user=user, bed_id=bed_id)
+    except Bed.DoesNotExist:
+        raise ObjectDoesNotExist(f"Bed matching ID {bed_id} could not be found.")
+    bed = bed_get(user=user, bed_id=bed_id)
+
+    m2m_data: Dict[str, Any] = {
```
vecchp, line 66: 🐛 **Bug: Redundant  query**

The first  result on line 64 is discarded — it's only used for the  guard, then the function calls  again on line 67. This duplicates 2-4 DB queries per update.

bed

Note:  in  line 67 gets this right — the result is captured in the  and used directly.

apps/betterangels-backend/shelters/services/bed.py
```diff
@@ -45,4 +41,104 @@ def bed_create(*, user: "User", data: Dict[str, Any]) -> Bed:
...
+@transaction.atomic
+def bed_delete(*, user: "User", ids: list[int]) -> list[int]:
+    """Delete rooms by their IDs and return the deleted instances.
+
+    Raises:
+        ``ObjectDoesNotExist`` when any of the given IDs does not match a room.
+    """
+    beds = admin_bed_list(Bed.objects.all(), user=user).filter(pk__in=ids)
+    deleted_ids = []
+
```
vecchp: ⚡ **Performance: Use `QuerySet.delete()` for bulk deletion**

The current implementation iterates and deletes each bed individually — that's N separate DELETEs. Django's `QuerySet.delete()` issues a single `DELETE FROM ... WHERE id IN (...)`:

    # Current (N DELETEs)
    beds = admin_bed_list(Bed.objects.all(), user=user).filter(pk__in=ids)
    deleted_ids = []
    for bed in beds:
        id = bed.pk
        bed.delete()
        deleted_ids.append(id)
    return deleted_ids

    # Fix (1 DELETE)
    beds = admin_bed_list(Bed.objects.all(), user=user).filter(pk__in=ids)
    deleted_ids = list(beds.values_list("pk", flat=True))
    beds.delete()
    return deleted_ids

Same issue applies to `room_delete` in `room.py` line 134-137.

apps/betterangels-backend/shelters/services/bed.py
```diff
@@ -45,4 +41,104 @@ def bed_create(*, user: "User", data: Dict[str, Any]) -> Bed:
...
+
+def _clone_label(label: str | None) -> str | None:
+    if not label:
+        return None
+    return f"{label} (Copy)"
+
+
+@transaction.atomic
+def bed_delete(*, user: "User", ids: list[int]) -> list[int]:
```
vecchp, line 86: 📝 **Typo in docstring**

The docstring says `"""Delete rooms by their IDs..."""` — should be `"""Delete beds by their IDs..."""`.

apps/betterangels-backend/shelters/services/bed.py
```diff
@@ -45,4 +41,104 @@ def bed_create(*, user: "User", data: Dict[str, Any]) -> Bed:
...
+        maintenance_flag=source.maintenance_flag,
+        name=_clone_label(source.name),
+        room=source.room,
+        shelter=source.shelter,
+        status=source.status,
+        status_notes=source.status_notes,
+        storage=source.storage,
+        type=source.type,
+    )
+    clone.full_clean()
```
vecchp: ⚡ Enrichment: Clone doesn't handle name collisions on repeat clones

Room clone has _unique_clone_name that generates "Room-101 (Copy)" then "Room-101 (Copy 2)" on repeat clones. Bed clone uses plain _clone_label, always producing "Bed 1 (Copy)". While beds don't have a unique constraint on name, duplicate names could confuse operators in the UI. Consider adding the same increment logic for consistency, or documenting the intentional asymmetry.

apps/betterangels-backend/shelters/services/bed.py
```diff
@@ -45,4 +41,104 @@ def bed_create(*, user: "User", data: Dict[str, Any]) -> Bed:
...
+
+    bed.full_clean()
+    bed.save()
+
+    if m2m_data:
+        _set_m2m_from_enums(bed, m2m_data)
+
+    return bed
+
+
```
vecchp, line 84: 🔧 Style: _clone_label is duplicated in bed.py and room.py

Both bed.py (line 85) and room.py (line 87) define their own _clone_label with different signatures:

- bed.py: _clone_label(label) returns str | None (returns None when label is falsy)
- room.py: _clone_label(label) returns str (returns "(Copy)" when label is falsy)

Since services/utils.py already houses shared helpers like _BED_M2M_FIELDS and _set_m2m_from_enums, this belongs there too. Choose one version and import from both services.

apps/betterangels-backend/shelters/services/bed.py
```diff
@@ -45,4 +41,104 @@ def bed_create(*, user: "User", data: Dict[str, Any]) -> Bed:
...
+@transaction.atomic
+def bed_clone(*, user: "User", bed_id: str) -> Bed:
+    """Clone an existing bed, including all M2M relationships.
+
+    Validates org access via ``bed_get``.
+
+    Raises:
+        ``ObjectDoesNotExist`` when the bed is not found.
+        ``django.core.exceptions.ValidationError`` on invalid data.
+    """
```
vecchp, line 107: ⚡ Performance: Add prefetch_related to source fetch in clone

The clone loop calls getattr(source, field_name).all() for each M2M field - that's 5 separate SELECT queries. Prefetching reduces this to 1 query.

Same applies to room_clone in room.py (4 M2M fields).

apps/betterangels-backend/shelters/services/room.py
```diff
@@ -47,4 +44,132 @@ def room_create(*, user: "User", data: Dict[str, Any]) -> Room:
...
+
+    Raises:
+        ``ObjectDoesNotExist`` when any of the given IDs does not match a room.
+    """
+    rooms = admin_room_list(Room.objects.all(), user=user).filter(pk__in=ids)
+    deleted_ids = []
+
+    for room in rooms:
+        id = room.pk
+        room.delete()
```
vecchp: ⚡ Performance: Use QuerySet.delete() for bulk deletion

Same issue as bed_delete - instance-by-instance deletion instead of bulk. Fix by using rooms.delete() which also handles SET_NULL cascade on Bed.room properly.

apps/betterangels-backend/shelters/tests/utils.py
```diff
@@ -0,0 +1,87 @@
...
+SHELTER_OPERATOR_PERMISSIONS = (
+    (Bed, "add_bed"),
+    (Bed, "change_bed"),
+    (Bed, "delete_bed"),
+    (Bed, "view_bed"),
+    (Room, "add_room"),
+    (Room, "add_room"),
+    (Room, "change_room"),
+    (Room, "delete_room"),
+    (Room, "view_room"),
```
vecchp, line 18: Bug: Duplicate add_room permission entry

(Room, "add_room") appears twice. Drop one of the two entries to avoid granting the permission redundantly in grant_operator_permissions.

apps/betterangels-backend/shelters/services/bed.py
vecchp, line 107: 💡 Enhancement: Use model meta introspection instead of hand-coding every field

Currently both `bed_clone` and `room_clone` hardcode every scalar field (11 fields each). This is fragile — if a new field is added to the model, the clone silently drops it. Django's _meta API can introspect fields at runtime:

```python
# Shared util in services/utils.py
def _clone_model_instance(source, *, overrides=None, exclude_fields=None):
    exclude = {"pk", "id", "created_at", "updated_at"}
    if exclude_fields:
        exclude |= exclude_fields
    clone = source.__class__(
        **{f.name: getattr(source, f.name)
           for f in source._meta.concrete_fields
           if f.name not in exclude}
    )
    if overrides:
        for key, value in overrides.items():
            setattr(clone, key, value)
    clone.full_clean()
    clone.save()
    return clone
```

Then `bed_clone` reduces to:
```python
clone = _clone_model_instance(source, overrides={"name": _clone_label(source.name)})
for field_name in _BED_M2M_FIELDS:
    getattr(clone, field_name).set(getattr(source, field_name).all())
return clone
```

This is consistent with the project's existing use of `_meta` (e.g. `_get_m2m_field_names` already calls `model._meta.many_to_many`). The public `_meta` API is stable and documented. As a bonus, you can add `prefetch_related` before the M2M loop to save N queries:
```python
source = source.__class__.objects.prefetch_related(
    *[f.name for f in source._meta.many_to_many]
).get(pk=source.pk)
```

apps/betterangels-backend/shelters/services/bed.py
vecchp: 💡 **Enhancement: Consider `django-clone` library to eliminate manual cloning boilerplate**

[django-clone](https://github.com/tj-django/django-clone) (v5.5.0, released June 2025, actively maintained) is a Django cloning library that supports:
- Scalar field copying via `_meta.concrete_fields` introspection
- **M2M copy** via `_clone_m2m_fields` / `_clone_excluded_m2m_fields` Model Meta
- **Unique name suffix increment** automatic ("Bed 1 (Copy)" → "Bed 1 (Copy 2)")
- **Exclusion control** (e.g. don't copy beds when cloning rooms: `_clone_excluded_m2o_or_o2m_fields = ["beds"]`)

Usage would be:
```python
class Bed(CloneMixin, BaseModel):
    class Meta:
        _clone_m2m_fields = ["accessibility", "demographics", "funders", "medical_needs", "pets"]

class Room(CloneMixin, BaseModel):
    class Meta:
        _clone_m2m_fields = ["accessibility", "demographics", "funders", "pets"]
        _clone_excluded_m2o_or_o2m_fields = ["beds"]
```

Then `bed_clone` reduces to:
```python
source = bed_get(user=user, bed_id=bed_id)
clone = source.make_clone(attrs={"name": _clone_label(source.name)})
return Bed.objects.get(pk=clone.pk)  # re-fetch for org-scoped GraphQL resolution
```

This would eliminate ~40 lines of boilerplate (`bed_clone()`, `room_clone()`, `_clone_label()`, `_copy_number_pattern()`, `_unique_clone_name()`). Tradeoff: adds `django-clone` + `conditional` as dependencies. This is a follow-up consideration — the current manual approach works correctly and is mergeable as-is.

vecchp (approved):

PR #2136, opened 2026-06-10
shelter operator beds frontend

mikefeldberg:
SDB-226, SDB-227, SDB-228, SDB-229, SDB-230, SDB-231


vecchp (commented):
## Overall Assessment

Solid work adding the bed create/edit pages and plumbing through mutations. The form structure with `react-hook-form` + `zod`, the `key={bed.id}` pattern on EditBedPage, and the loading/error/not-found states in EditBedPage are all good. A few areas worth addressing before merging:

---

## 🟡 Design & Architectural Concerns

### 1. Delete confirmation should not live inside `BedTable`

The PR adds a `ConfirmationModal` directly inside `BedTable` along with `useState` for managing its open/close state. This couples the generic table component to one specific interaction pattern and makes it harder to reuse the table without confirmation, or with a different confirmation UX.

**Recommendation:** Move the confirmation state + modal into `BedsView` (the parent). `BedTable` should keep a simple `onDeleteBeds` callback (which it does), and the parent can decide what confirmation UI to render before calling the mutation.

```tsx
// BedsView should own the confirmation modal:
<>
  {/* ...error banner... */}
  <BedTable rooms={rooms} onDeleteBeds={handleDeleteBedsRequest} ... />
  <ConfirmationModal
    isOpen={isConfirmOpen}
    title={confirmTitle}
    onClose={closeConfirm}
    primaryAction={{ label: "Delete", onClick: () => { handleDeleteBeds(confirmBedIds); closeConfirm(); } }}
    secondaryAction={{ label: "Cancel", onClick: closeConfirm }}
    variant="danger"
    description="This action cannot be undone."
  />
</>
```

### 2. Inconsistent error handling between `handleClone` and `handleDeleteBeds`

`handleClone` uses `errorPolicy: 'all'` and checks for `OperationInfo` union types in the response. `handleDeleteBeds` does neither — it only has a bare `try/catch`. GraphQL mutations that return union types (`BulkDeleteResult | OperationInfo`) can succeed at the HTTP level but return `OperationInfo` with error messages. Without checking this, application-level errors from `deleteBeds` will be silently swallowed.

**Recommendation:** Use `errorPolicy: 'all'` for `deleteBeds` too and check `result.deleteBeds.__typename === 'OperationInfo'` before treating it as success, matching the `handleClone` pattern.

### 3. Inconsistent prop naming — `onClone` in `BedTable`

The PR renames `onDelete` → `onDeleteBeds` but leaves `onClone` as-is. If `onDeleteBeds` was renamed to signal it takes `bedIds: string[]` (plural/bulk semantics), that is reasonable — but then having `onClone` alongside it in the same component interface is inconsistent. If the distinction matters, consider `onDuplicateBed` to match the pattern; if not, keep both simple (`onEdit`, `onDelete`, `onClone`).

---

## 🟡 Code Organization

### 4. `buildUpdateBedInput` lives in `createBedMutation.ts`

The file `createBedMutation.ts` exports both `buildCreateBedInput` and `buildUpdateBedInput`. `updateBedMutation.ts` then re-exports it. This is confusing — a file named `createBedMutation` should not be the source of truth for update logic.

**Recommendation:** Move the shared `buildBedFieldInput` into a separate file (e.g., `bedFormInput.ts`) and import it from both mutation files, or keep each builder in its respective named file.

### 5. Unnecessary re-export

`createBedMutation.ts` exports `CreateBedDocument` (generated import) as `CREATE_BED_MUTATION`. The generated type `CreateBedDocument` is already the canonical name. The all-caps re-export adds an alias without semantic benefit and diverges from how other mutations in this codebase are referenced (e.g., `GetBedsDocument`, `DeleteBedsDocument`).

**Recommendation:** Drop the `CREATE_BED_MUTATION` / `UPDATE_BED_MUTATION` aliases and use the generated constants directly.

---

## 🟡 Minor Issues

### 6. Duplicate error banner styles

Both `BedsView` and `BedForm` define identical error banner markup:
```tsx
<div className="... rounded-md border border-red-200 bg-red-50 px-4 py-3 text-sm text-red-800" role="alert">
```

**Recommendation:** Extract to a shared `<ErrorBanner>` component in `base-ui/`.

### 7. Magic z-index on FAB button

The "Create Bed" FAB uses `z-20`, which is an arbitrary value. If other UI elements (modals, dropdowns, etc.) use higher z-indexes this will conflict. Using a design token or consistent z-index scale across the app would be safer.

### 8. `handleCancel` resets to defaults in edit mode

In `BedForm`, `handleCancel` calls `reset()` which resets to `defaultValues` set at form initialization. In create mode this is fine (back to empty). In edit mode with `initialData` provided, `reset()` without arguments should reset to the last-submitted values or the defaultValues. This _should_ be correct since you pass `defaultValues: initialData`, but it is worth verifying this behaves correctly on a dirty edit form.

### 9. Stories appear out of sync

The `BedTable.stories.tsx` diff shows it was updated from `onDelete` → `onDeleteBeds`. This is good, but verify these stories still render correctly with the confirmation modal now embedded in `BedTable` (the stories pass `onDeleteBeds={() => undefined}` which means the modal primary action calls `undefined?.(...)` — fine, but the modal still opens/stays open with no-op actions, which is not ideal for Storybook).

---

## ✅ What Looks Good

- `EditBedPage` properly handles loading, error, and not-found states
- Using `key={bed.id}` on `BedForm` to force re-mount when switching between beds
- `formSchema` uses `z.array(z.enum(...))` for type-safe enum validation
- `mapBedToFormData` now uses the query result type directly instead of a manual `Pick<>`
- `CreateBedPage`/`EditBedPage` are thin page components that delegate to `BedForm`
- The `useBed` hook is clean and properly exported
- Error banners use `role="alert"` for accessibility

---

**Reviewed by Cline**

vecchp (approved):
approved but take a look at the ai comments some of them seem pretty real.

PR #2137, opened 2026-06-11
shelter operating hours (SDB-243)

tglaz:
### Operating Hours form
SDB-243

- create `ScheduleForm` (screenshots below)
  - used for Shelter schedules
- add lib components
  - `ButtonDropdown`
  - `Tabs`
- update `Dropdown` component
  - can un-select option
  - abstract dropdown items into reusable components: `MenuPanel`, `MenuOption`
  - add `date` type support to `Input`
  

### screenshot: regular schedule
image

### screenshot: schedule exception
image


vecchp (commented):
## Review: Shelter Operating Hours (SDB-243)

Great work on this — the component decomposition is clean, the zod discriminated union for `closedAllDay` is well done, and the shared `MenuPanel`/`MenuOption` abstraction is a nice improvement.

One 🔴 issue (stale closure in `Dropdown`) that should be fixed before merge, plus a few 🟡 suggestions below.

---

### 🔴 `Dropdown.tsx` — stale closure: `onChange(null)` bypasses `emitChange`

On `main`, `handleOptionClick` only uses `emitChange`. The PR adds a deselection path for single-select that calls `onChange(null)` directly. But `onChange` is **not** in `handleOptionClick`'s dependency array — if the parent re-renders with a new `onChange` while the dropdown is open, clicking to deselect fires the stale closure.

`emitChange` already handles this correctly: passing `[]` with `isMulti=false` produces `onChange(null)`. And `emitChange` *is* in the deps array.

**Suggested fix:**
```diff
-          onChange(null);
+          emitChange([]);
```

---

### 🟡 `DateOrRangePicker.tsx` — range end date silently lost on collapse

When collapsing from range back to single day, the user's end date is discarded:

```ts
} else {
  // Collapse back to single day
  onChange('endDate', startDate);  // ← overwrites user-selected endDate
}
```

Scenario: user sets startDate = "2026-06-10", endDate = "2026-06-15", then clicks "Remove end date". If they toggle back to range, it recomputes `startDate + 1` = "2026-06-11", not "2026-06-15".

**Suggested fix** — preserve the previous end date in a ref:
```ts
const prevEndDateRef = useRef<string>(endDate);

const handleSingleToggle = () => {
  if (isSingleDay) {
    const restored = prevEndDateRef.current && prevEndDateRef.current !== startDate
      ? prevEndDateRef.current
      : (() => {
          const next = new Date(startDate);
          next.setDate(next.getDate() + 1);
          return next.toISOString().slice(0, 10);
        })();
    onChange('endDate', restored);
  } else {
    prevEndDateRef.current = endDate;
    onChange('endDate', startDate);
  }
};
```

---

### 🟡 `ScheduleExceptionsForm/utils.ts` + `ScheduleForm/utils.ts` — `localId` collision

Two separate ID counters can produce duplicates. `buildDefaultExceptionEntry` uses a module-level `_nextId`, while `hydrateExceptions` uses a function-scoped `_id`. If a user hydrates 2 exceptions (IDs "1", "2") and then adds a new one via `buildDefaultExceptionEntry()` (which also starts at "1"), you get duplicate `localId: "1"`.

This breaks the `htmlFor`/`id` pairing on the "Closed all day" checkbox in `ExceptionRow`:

```tsx
<input id={`closed-${entry.localId}`} ... />
<label htmlFor={`closed-${entry.localId}`} ...>Closed all day</label>
```

**Suggested fix** — use `crypto.randomUUID()` in both places:

```ts
// buildDefaultExceptionEntry
localId: crypto.randomUUID(),

// hydrateExceptions
localId: s.id ?? crypto.randomUUID(),
```

`useFieldArray`'s `keyName: '_rhfId'` already gives stable React list keys, so `localId` doesn't need to be sequential.

---

### 🟡 `ShelterSchedules.tsx` — toast mislabeling when form yields zero inputs

```ts
const isDelete = scheduleInputs.length === 0;
```

`buildScheduleInputs` returns an empty array when all weekly days have no time ranges. If a user clears all hours and clicks "Save Schedule", they see a "Schedule deleted." toast — but they never intended to delete.

The explicit delete path already exists (confirmation modal → `onSave(currentTab, [])`), so don't derive delete vs save intent from the input length:

```ts
// After successful save:
showToast({
  status: 'success',
  title: 'Schedule updated.',
});
```

The confirmation modal handles the delete toast.

---

### 🟡 `ScheduleForm.tsx` — no re-validation trigger for exception changes

The weekly editor clears field errors on edit after a failed submit:

```ts
function onSetWeekly(next: WeeklyFormState, validate = true) {
  setValue('weekly', next);
  if (isSubmitted && validate) {
    trigger('weekly');
  }
}
```

But the `ExceptionRow` `onChange` doesn't trigger re-validation:

```tsx
onChange={(patch) => {
  onChange({ ...value, ...patch });
  // ← missing: if (isSubmitted) trigger(`exceptions.${index}`);
}}
```

This is inconsistent UX — weekly errors clear on edit, but exception errors persist until next submit.

**Suggested fix:**
```tsx
onChange={(patch) => {
  onChange({ ...value, ...patch });
  if (isSubmitted) trigger(`exceptions.${index}`);
}}
```

`isSubmitted` and `trigger` are already destructured at the top of `ScheduleForm`.

---

### Verdict

Approve with the 🔴 fix required. The rest are quality improvements.

libs/react/shelter-operator/src/lib/components/ShelterProfile/segments/Schedules/ScheduleForm/ScheduleExceptionsForm/DateOrRangePicker.tsx
```diff
@@ -0,0 +1,71 @@
+import { Input } from '../../../../../base-ui/input';
```
vecchp, line 2: 🟡 **Range end date silently lost on collapse**

When collapsing from range back to single day, the users end date is overwritten:

```ts
} else {
  onChange('endDate', startDate);  // ← overwrites user-selected endDate
}
```

Scenario: user sets start = "2026-06-10", end = "2026-06-15", clicks "Remove end date". If they toggle back to range, it computes `startDate + 1` = "2026-06-11", not "2026-06-15". Original end date is lost.

**Suggested fix — preserve in a ref:**
```ts
const prevEndDateRef = useRef<string>(endDate);

const handleSingleToggle = () => {
  if (isSingleDay) {
    const restored = prevEndDateRef.current && prevEndDateRef.current !== startDate
      ? prevEndDateRef.current
      : (() => {
          const next = new Date(startDate);
          next.setDate(next.getDate() + 1);
          return next.toISOString().slice(0, 10);
        })();
    onChange('endDate', restored);
  } else {
    prevEndDateRef.current = endDate;
    onChange('endDate', startDate);
  }
};
```

libs/react/shelter-operator/src/lib/components/ShelterProfile/segments/Schedules/ScheduleForm/ScheduleExceptionsForm/utils.ts
```diff
@@ -0,0 +1,15 @@
+import type { ExceptionEntry } from '../types';
```
vecchp, line 1: 🟡 **`localId` collision between `buildDefaultExceptionEntry` and `hydrateExceptions`**

Two independent ID counters can produce duplicates. Module-level `_nextId` in this file and function-scoped `_id` in `ScheduleForm/utils.ts` both start at 0/1. If user hydrates 2 exceptions (IDs "1", "2") and adds a new one via `buildDefaultExceptionEntry()` (also "1"), you get duplicate `localId: "1"`.

This breaks the `htmlFor`/`id` pairing on the checkbox label in `ExceptionRow`:
```tsx
<input id={`closed-${entry.localId}`} />
<label htmlFor={`closed-${entry.localId}`}>Closed all day</label>
```

**Fix:** Use `crypto.randomUUID()` here, and do the same in `hydrateExceptions`:
```suggestion
localId: crypto.randomUUID(),
```

`useFieldArray`s `keyName: '_rhfId'` already gives stable React list keys, so `localId` doesn't need to be sequential.

libs/react/shelter-operator/src/lib/components/ShelterProfile/segments/Schedules/ShelterSchedules.tsx
```diff
@@ -0,0 +1,157 @@
+import {
```
vecchp, line 1: 🟡 **Toast mislabeling: empty `scheduleInputs` triggers "Schedule deleted."**

```ts
const isDelete = scheduleInputs.length === 0;
```

`buildScheduleInputs` returns an empty array when all weekly days have no time ranges. If a user clears all hours and clicks "Save Schedule", they see a "Schedule deleted." toast — but they never intended to delete.

The explicit delete path already exists (confirmation modal → `onSave(currentTab, [])`). The intent shouldn't be derived from the input length. Just always use "Schedule updated." and let the confirmation modal handle the delete toast.

libs/react/shelter-operator/src/lib/components/ShelterProfile/segments/Schedules/ScheduleForm/ScheduleForm.tsx
```diff
@@ -0,0 +1,147 @@
+import { zodResolver } from '@hookform/resolvers/zod';
```
vecchp, line 1: 🟡 **No re-validation trigger for exception changes after failed submit**

The weekly editor clears field errors on edit after a failed submit:

```ts
function onSetWeekly(next, validate = true) {
  setValue('weekly', next);
  if (isSubmitted && validate) trigger('weekly');
}
```

But the `ExceptionRow` `onChange` doesn't trigger re-validation:

```tsx
onChange={(patch) => {
  onChange({ ...value, ...patch });
  // missing: if (isSubmitted) trigger(`exceptions.${index}`);
}}
```

This means exception errors persist until next submit, while weekly errors clear as you type — inconsistent UX.

**Suggested fix:**
```tsx
onChange={(patch) => {
  onChange({ ...value, ...patch });
  if (isSubmitted) trigger(`exceptions.${index}`);
}}
```

`isSubmitted` and `trigger` are already destructured at the top of `ScheduleForm`.

vecchp (commented):
🔴 **Stale closure: `onChange(null)` bypasses `emitChange`**

On `main`, `handleOptionClick` only uses `emitChange`. This adds a deselection path that calls `onChange(null)` directly. But `onChange` is **not** in `handleOptionClick`s dependency array. If the parent re-renders with a new `onChange` while open, clicking to deselect fires the stale closure.

`emitChange` already handles this correctly: passing `[]` with `isMulti=false` produces `onChange(null)`. And `emitChange` *is* in the deps array.

**Suggested fix (one line):**
```suggestion
emitChange([]);
```

vecchp (commented):
Two additional React best practice items from a second pass.

libs/react/shelter-operator/src/lib/components/ShelterProfile/segments/Schedules/ScheduleForm/WeeklyScheduleEditor/DayRow.tsx
```diff
@@ -0,0 +1,135 @@
+import { mergeCss } from '@monorepo/react/shared';
```
vecchp, line 1: 🟡 **`key={index}` on a mutable list — React anti-pattern**

```tsx
{schedule.ranges.map((range, index) => (
  <div key={index} className="flex flex-col gap-1">
```

When users add/remove ranges, React reconciles siblings by index. Removing a mid-list range causes all subsequent DOM nodes to shift — which can cause transient state bleeding or focus loss on adjacent inputs. Use stable ids instead:

```ts
// Add a localId to each TimeRange when creating ranges
const EMPTY_RANGE = { startTime: '', endTime: '', localId: crypto.randomUUID() };

// Then use it as the key
<div key={range.localId} ...>
```

```ts
// Update the TimeRange type in types.ts to include localId:
export type TimeRange = {
  localId: string;
  startTime: string;
  endTime: string;
};
```

Low severity today since the items are stateless time strings, but a well-known React pitfall worth fixing.

libs/react/shelter-operator/src/lib/components/ShelterProfile/segments/Schedules/ScheduleForm/WeeklyScheduleEditor/DayRow.tsx
```diff
@@ -0,0 +1,135 @@
...
+import { DayOfWeekChoices } from '@monorepo/react/shelter';
```
vecchp, line 2: ⚪ **Unused `day` prop**

```tsx
const {
  day: _day,  // ← never referenced in the component body
  label,
  schedule,
  ...
} = props;
```

`day` is destructured as `_day` (the convention for "intentionally unused") but the component never reads it. If it was intended for a `data-day` attribute or `aria-describedby` in a follow-up, please add a comment. Otherwise, remove it from `TProps` and the destructure to keep the interface clean.

vecchp (approved):

libs/react/shelter-operator/src/lib/components/ShelterProfile/segments/Schedules/ShelterSchedules.tsx
tglaz: i can live with that

PR #2143, opened 2026-06-16
SDB-178 PR #4: Self-signup flow + org-scoped permissions (HasOrgPerm)

vecchp:
## Overview

PR #4 of 4 for SDB-178 (shelter operator self-onboarding). Enables the complete self-signup flow: new users can sign in via OTP (allauth auto-creates their account), then create a shelter organization from the operator dashboard.

**Base:** _main_ (builds on PR #1-3 already merged)

### Org-Scoped Permission Architecture (NEW)
- **common/permissions/utils.py** — `permissioned_queryset()` is the single source of truth for org-scoped permission checks, shared across `HasOrgPerm`, selectors, and type hooks. `_perm_q()` accepts optional `prefix` parameter for querying different models (Organization vs PermissionGroup).
- **accounts/extensions.py** — `HasOrgPerm` Strawberry extension validates ALL requested permissions against the active organization via `permissioned_queryset()`. Reads org ID from `X-Organization-ID` header. Properly handles `organization_field="pk"` for Organization model.
- **accounts/permissions.py** — `permission_annotations()` uses `_perm_q()` with `prefix="group__permissions"` for direct PermissionGroup queries.
- **Deleted dead code**: `AdminShelterQuerySet`/`AdminShelterManager`/`admin_objects` manager (replaced by `permissioned_queryset()` + selectors + `HasOrgPerm`).

### Admin → Operator Renaming
- `adminShelters` → `operatorShelters`, `adminShelter` → `operatorShelter`
- `AdminShelterType` → `OperatorShelterType`
- `AdminShelterQueryTestCase` → `OperatorShelterQueryTestCase`
- `test_admin_queries.py` → `test_operator_queries.py`
- Frontend hooks: `useAdminShelterProfile` → `useShelterOperatorProfile`

### Selector Architecture
- **shelters/selectors/** — package replacing monolithic `selectors.py`:
  - `operator.py`: list filters, queryset wrappers (`shelter_queryset`, `room_queryset`, `bed_queryset`), entity lookups (`shelter_get`, `room_get`, `bed_get`)
  - `reports.py`: report aggregation functions
  - `__init__.py`: re-exports
- **shelters/open_at.py** — extracted `shelters_open_at` helper (breaks circular import between models/managers/selectors)

### Schema + Services
- **shelters/schema.py** — thin GraphQL layer: mutations extract user/org_id, delegate to services. All queries/mutations use `HasOrgPerm`.
- **shelters/services/** — services take `(user, organization_id, ...)` and scope internally via selectors (no queryset parameters, no duplicate org checks).

### Testing Fixes
- Query count expectations updated for `HasOrgPerm` (+1 query per permission check)
- `ObjectDoesNotExist` handled via `OperationInfo` (caught by `MUTATIONS_DEFAULT_HANDLE_ERRORS`)
- `execute_graphql()` accepts `**extra` kwargs (e.g., `HTTP_X_ORGANIZATION_ID`) for header-based testing
- `owner_roles=(SHELTER_OPERATOR,)`, `preset_names=["shelter"]` in service tests

### Documentation
- **docs/permissions.md** — rewritten with org-scoped permissions architecture, key files reference table, `permissioned_queryset()` API docs, testing guide

## Safety

- New accounts have no org memberships or permission groups → cannot access any shelter data
- All org-scoped mutations/queries use `@HasOrgPerm` (header-based org validation)
- `HasOrgPerm` validates ALL requested permissions (single query, respects `anyPerm` flag)
- `fail_silently=False` by default — permission denials raise rather than silently returning empty results
- No fallback to global permission checks when org header is missing
- Missing org header raises `DjangoNoPermission` immediately (before any query)

## Generated files
- `schema.graphql` — regenerated
- `**/__generated__/types.ts` — regenerated for all libs
- Build verified


apps/betterangels-backend/accounts/emails.py
```diff
@@ -0,0 +1,37 @@
...
+
+from accounts.models import Organization, User
+from django.conf import settings
+from django.core.mail import EmailMultiAlternatives
+from django.template.loader import render_to_string
+
+logger = logging.getLogger(__name__)
+
+
+def send_welcome_email(user: User, organization: Organization) -> None:
```
vecchp: should be generic function or in the shelter app. makes no sense to put it here.

apps/betterangels-backend/accounts/schema.py
```diff
@@ -243,3 +253,70 @@ def remove_organization_member(
         )
 
         return DeletedObjectType(id=removed_id)
+
+    # ── Organization Creation ──────────────────────────────────────
+
+    @strawberry.mutation(permission_classes=[IsAuthenticated])
+    def create_organization(self, info: Info, data: CreateOrganizationInput) -> CreateOrganizationResponse:
```
vecchp, line 250: This should also be generic if possible.

apps/betterangels-backend/shelters/selectors.py
```diff
@@ -109,8 +109,20 @@ def shelter_list(
...
+    *,
+    user: "User",
+    organization_id: str | None = None,
+) -> "QuerySet[Shelter]":
+    """Filter to shelters whose organization the *user* belongs to.
+
+    If *organization_id* is provided, scopes to that org.  Otherwise
+    returns all orgs the user belongs to.
+    """
+    if organization_id:
```
vecchp: This is a special action so i think we need to prevent anyone from doing this.  I thin kwe need to require orgid here. don't fall back to listeng them all.

apps/betterangels-backend/accounts/schema.py
```diff
@@ -243,3 +253,70 @@ def remove_organization_member(
...
+    ) -> OrganizationMemberType:
+        """Promote or demote a member within an organization.
+
+        Replaces all of the member's current org-scoped roles with the
+        single requested template.  Requires the ``CHANGE_ORG_MEMBER_ROLE``
+        permission.
+        """
+        current_user = cast(User, get_current_user(info))
+
+        organization = get_user_permitted_org(
```
vecchp: do we want to make a new decorator / extension taht checks thes perm outright instead of checking the global per on the outside and then again on the inside. Probably an opportunity to create something reusable

apps/betterangels-backend/betterangels_backend/settings.py
```diff
@@ -170,8 +170,8 @@
 
 ACCOUNT_ADAPTER = "accounts.adapters.AccountAdapter"
 ACCOUNT_EMAIL_UNKNOWN_ACCOUNTS = False
-ACCOUNT_EMAIL_VERIFICATION = "mandatory"
-ACCOUNT_EMAIL_VERIFICATION_BY_CODE_ENABLED = True
+ACCOUNT_EMAIL_VERIFICATION = "optional"
+ACCOUNT_EMAIL_VERIFICATION_BY_CODE_ENABLED = False
 ACCOUNT_EMAIL_VERIFICATION_BY_CODE_TIMEOUT = 300
```
vecchp: Given we are reverting these settings i think we can completely remove them now?

apps/betterangels-backend/accounts/emails.py
```diff
@@ -0,0 +1,37 @@
...
+
+    if not html_template or not txt_template:
+        logger.warning("No welcome email templates configured for Shelter Operator — skipping.")
+        return
+
+    context = {
+        "user_email": user.email,
+        "user_first_name": user.first_name,
+        "organization_name": organization.name,
+        "dashboard_url": getattr(settings, "SHELTER_OPERATOR_DASHBOARD_URL", "http://localhost:4200/dashboard"),
```
vecchp: the defaults should be in the actual settings.

apps/betterangels-backend/schema.graphql
```diff
@@ -1951,8 +1972,8 @@ type Query {
   reportSummary(organizationId: ID!, startDate: Date = null, endDate: Date = null): ReportSummaryType!
   task(pk: ID!): TaskType! @hasRetvalPerm(permissions: [{app: "tasks", permission: "view_task"}], any: true)
   tasks(ordering: [TaskOrder!] = null, filters: TaskFilter, pagination: OffsetPaginationInput): TaskTypeOffsetPaginated! @hasRetvalPerm(permissions: [{app: "tasks", permission: "view_task"}], any: true)
-  adminShelter(pk: ID!): AdminShelterType! @hasPerm(permissions: [{app: "shelters", permission: "view_shelter"}], any: true)
-  adminShelters(pagination: OffsetPaginationInput, filters: ShelterFilter, ordering: [ShelterOrder!]! = []): AdminShelterTypeOffsetPaginated! @hasPerm(permissions: [{app: "shelters", permission: "view_shelter"}], any: true)
+  adminShelter(pk: ID!): AdminShelterType! @hasOrgPerm
```
vecchp: We sitll want the additional metadata here which basically tells the user they need the org permission to view shetlers.

apps/betterangels-backend/accounts/extensions.py
```diff
@@ -0,0 +1,116 @@
...
+            obj_perm_checker if obj_perm_checker is not None else _default_obj_perm_checker
+        )
+        self.with_anonymous = with_anonymous
+        self.with_superuser = with_superuser
+
+    def resolve_for_user(self, user: Any, info: Info) -> None:
+        request = info.context.request
+        org_id: str | None = getattr(request, "organization_id", None)
+
+        if not org_id:
```
vecchp: Do we really want this? Right now all orgs are part of groups that have stuff like view shelter etc.  If we fall back to group permissions won't they bypass the org requirement?

vecchp: We also want to revisit of django guardian is the appropriate permission approach.

vecchp:
### PR comment resolution summary

1. **"emails.py should be generic"** ✅ Fixed in `31f541075` — `send_welcome_email` now accepts `TemplateConfig` from registry, no shelter import.

2. **"create_organization should be generic"** ✅ Fixed in `31f541075` — mutation loops over `REGISTRY.templates_for(org)`, sends welcome per template.

3. **"admin_shelter_list should require org_id"** ✅ Membership check added in `b08e47d8f` — when `organization_id` is provided, selector verifies user belongs to that org.

4. **"create a reusable extension"** ✅ `HasOrgPerm` created in `a2a6c2170` — extends `HasPerm`, validates org membership + permission.

5. **"remove reverted BY_CODE settings"** ✅ Fixed in `767baf278` — removed `ACCOUNT_EMAIL_VERIFICATION_BY_CODE_ENABLED` and `_TIMEOUT`.

6. **"dashboard URL default in settings"** ✅ Fixed in `767baf278` — added `SHELTER_OPERATOR_DASHBOARD_URL` to settings.

7. **"schema directive metadata"** ✅ Fixed in `767baf278` — `HasOrgPerm` now extends `HasPerm` so permissions show in schema.

8. **"don't fall back to group permissions"** ✅ Fixed in `767baf278` — removed the `super().resolve_for_user()` fallback. Missing header always raises `PermissionDenied`.

9. **"revisit django-guardian"** 📋 Deferred — separate discussion, not scope of this PR.

apps/betterangels-backend/pyproject.toml
```diff
@@ -65,3 +65,11 @@ django-clone = "^5.5.0"
...
 requires = ["poetry-core"]
 build-backend = "poetry.core.masonry.api"
+
+[dependency-groups]
+dev = [
+    "vcrpy (>=8.2.1,<9.0.0)",
+    "pytest (>=9.1.0,<10.0.0)",
+    "pytest-django (>=4.12.0,<5.0.0)",
+    "model-bakery (>=1.23.5,<2.0.0)"
+]
```
vecchp: This shouldn't be here remove it and make sure we only do this at the global level.

apps/betterangels-backend/shelters/groups.py
```diff
@@ -15,4 +15,7 @@
     ],
     invite_html="account/email/shelter_operator_invite.html",
     invite_txt="account/messages/shelter_operator_invite.txt",
+    welcome_html="shelters/email/shelter_operator_welcome.html",
+    welcome_txt="shelters/email/shelter_operator_welcome.txt",
+    welcome_dashboard_url="/operator",
```
vecchp: shouldn't this be a settings?

apps/betterangels-backend/betterangels_backend/settings.py
```diff
@@ -411,6 +410,9 @@
 EMAIL_FILE_PATH = str(BASE_DIR / "tmp" / "app-emails")
 INVITATION_BACKEND = "accounts.backends.CustomInvitations"
 
+# Welcome email dashboard URL (default for local dev — overridden per-org-type via TemplateConfig)
+SHELTER_OPERATOR_DASHBOARD_URL = "http://localhost:4200/dashboard"
```
vecchp: follow the .env patter here wehere we register the config, and have a sane default.

apps/betterangels-backend/accounts/services.py
```diff
@@ -207,3 +207,39 @@ def create_default_org_permission_groups(organization: Organization) -> None:
...
+    Links *user* as the organization owner and assigns the Shelter
+    Operator role.
+
+    Returns ``(user, organization)``.
+
+    Does **not** send a welcome email — callers (mutations) are
+    responsible for triggering email delivery after the transaction
+    commits successfully.
+    """
+    from shelters.groups import SHELTER_OPERATOR
```
vecchp: can we remove the inline import?

apps/betterangels-backend/accounts/schema.py
```diff
@@ -243,3 +253,72 @@ def remove_organization_member(
...
```
vecchp: Perhaps we can use the hasorgperm here?

apps/betterangels-backend/accounts/services.py
vecchp: Would prefer the enum / type safe value here vs string version of shelter.  Is there a way to do this without circular imports?

apps/betterangels-backend/accounts/services.py
```diff
@@ -207,3 +207,41 @@ def create_default_org_permission_groups(organization: Organization) -> None:
...
+    responsible for triggering email delivery after the transaction
+    commits successfully.
+    """
+    org_config = REGISTRY.org_type("shelter")
+    if not org_config or not org_config.allow_public_signup:
+        raise ValidationError("This org type does not support self-signup.")
+
+    # The first template of the configured org type is the member-level
+    # role (Shelter Operator for shelters, Caseworker for outreach).
+    member_template = org_config.templates[0]
```
vecchp: Not a fan AT all of this templates[0] approach please consider a new way to do this.  It is very hacky and gross.

apps/betterangels-backend/shelters/templates/shelters/email/shelter_operator_welcome.txt
```diff
@@ -0,0 +1,12 @@
+Welcome to BetterAngels, {{ user_first_name }}!
+
+You're all set! Your organization "{{ organization_name }}" has been created and you've been assigned as the Shelter Operator.
+
+As a Shelter Operator you can:
+- Create and manage shelters, rooms, and beds
+- Invite other staff members to join your organization
+- View shelter availability and occupancy
+
+Go to your Dashboard: {{ base_url }}/operator
```
vecchp, line 10: Is there a better way to have the paths, maybe setup the context in the org template? and pass the full value? What are your thoughts here?

apps/betterangels-backend/shelters/groups.py
vecchp: any way to actually reference the settings instead of doing it by string value?

apps/betterangels-backend/shelters/selectors.py
```diff
@@ -127,31 +138,37 @@ def shelter_get(*, user: "User", shelter_id: int | str) -> "Shelter":
...
-    return queryset.filter(shelter__in=admin_shelter_list(Shelter.objects.all(), user=user))
+def operator_room_list(queryset: "QuerySet[Room]", *, user: "User", organization_id: str) -> "QuerySet[Room]":
+    return queryset.select_related("shelter").filter(
+        Exists(Organization.objects.filter(pk=OuterRef("shelter__organization_id"), users=user)),
+        shelter__organization_id=organization_id,
+    )
 
 
-def room_get(*, user: "User", room_id: int | str) -> "Room":
+def room_get(*, user: "User", room_id: int | str, organization_id: str) -> "Room":
```
vecchp: Fix me

apps/betterangels-backend/common/permissions/utils.py
```diff
@@ -181,6 +183,130 @@ def has_permission(self, source: Any, info: strawberry.Info, **kwargs: Any) -> b
...
+    implicit — ``permission_groups__group__user`` proves both.
+
+    Parameters
+    ----------
+    queryset : QuerySet
+        The base queryset to filter (e.g. ``Shelter.objects.all()``).
+    user : User
+        The authenticated user.
+    organization_id : str
+        The active organization UUID.
```
mikefeldberg: not super important but we don't use uuids

apps/betterangels-backend/accounts/schema.py
```diff
@@ -243,3 +249,65 @@ def remove_organization_member(
...
+        current_user = cast(User, get_current_user(info))
+        user, organization = create_organization_service(
+            user=current_user,
+            organization_name=data.organization_name,
+            org_type_name=data.org_type,
+        )
+
+        templates = [t for t in REGISTRY.templates_for(organization) if t.welcome_html]
+        for template in templates:
+            send_welcome_email(user, organization, template)
```
mikefeldberg: not blocking, but selecting templates and sending emails seems like it belongs in a service rather than a mutation

apps/betterangels-backend/accounts/schema.py
```diff
@@ -243,3 +249,65 @@ def remove_organization_member(
...
+        org_id = get_current_organization(info)
+        organization = Organization.objects.get(pk=org_id)
+
+        template = REGISTRY.template(data.permission_template.value)  # type: ignore[attr-defined, union-attr]
+        if template is None:
+            valid = REGISTRY.invitable_template_names_for(organization)
+            raise ValidationError(
+                f"Invalid permission template '{data.permission_template.value}'. "  # type: ignore[attr-defined, union-attr]
+                f"Available: {', '.join(valid)}"
+            )
```
mikefeldberg: `add_organization_member` has an identical chunk. Maybe worth pulling both to a util or private fn

apps/betterangels-backend/common/org_types.py
```diff
@@ -33,6 +33,15 @@ class OrgTypeConfig:
     name: str  # "outreach"
     label: str  # "Outreach"
     templates: tuple[TemplateConfig, ...]  # (CASEWORKER, ORG_ADMIN, ORG_SUPERUSER)
+    member_template: TemplateConfig
+    """The member-level role template (e.g. Caseworker, Shelter Operator).
+
+    This is the default role assigned to new members and the role used
```
mikefeldberg, line 39:
```suggestion
    This specifies the default role assigned to new members and the role used
```

apps/betterangels-backend/shelters/selectors/operator.py
```diff
@@ -0,0 +1,191 @@
...
+    return queryset.filter(Exists(Organization.objects.filter(pk=OuterRef("organization_id"), users=user)))
+
+
+def operator_shelter_list(
+    queryset: "QuerySet[Shelter]",
+    *,
+    user: "User",
+    organization_id: str,
+) -> "QuerySet[Shelter]":
+    """Filter to shelters in *organization_id* that *user* belongs to."""
```
mikefeldberg: to keep consistent with `user_shelter_list` docstring
```suggestion
    """Filter to shelters belonging to *organization_id* that *user* is a member of."""
```

apps/betterangels-backend/shelters/selectors/operator.py
```diff
@@ -0,0 +1,191 @@
...
+    """Filter to shelters approved for public display.
+
+    If the user has ``view_private_shelter``, private shelters are included.
+    """
+    from shelters.models import Shelter
+
+    queryset = queryset.filter(status=StatusChoices.APPROVED)
+    if user and user.is_authenticated and hasattr(user, "has_perm") and user.has_perm(Shelter.perms.VIEW_PRIVATE):
+        return queryset
+    return queryset.filter(is_private=False)
```
mikefeldberg, line 38: technically not an `operator` selector. Makes sense to keep here for now, since it's the only public list we have atm. I can see us eventually having public facing `bed` and `room` lists, which would also live in a separate selectors file.

apps/betterangels-backend/shelters/services/bed.py
```diff
@@ -3,29 +3,34 @@
...
 
-    Validates that *user* belongs to the shelter's organization via
-    ``shelter_get`` before creating the bed.
+    Resolves *shelter* via :func:`~shelters.selectors.shelter_get` with
+    ``view_shelter`` permission.
 
     Raises:
-        ``ObjectDoesNotExist`` when the shelter is not found or the user
-        does not belong to its organization.
+        ``Shelter.DoesNotExist`` when the shelter is not found.
```
mikefeldberg: this raises `ObjectDoesNotExist`. Same for `bed_update`. `bed_delete` is technically correct but uses the full `django.core.exceptions.ObjectDoesNotExist`

apps/betterangels-backend/shelters/services/room.py
```diff
@@ -4,29 +4,34 @@
...
 
-    Validates that *user* belongs to the shelter's organization via
-    ``shelter_get`` before creating the room.
+    Resolves *shelter* via :func:`~shelters.selectors.shelter_get` with
+    ``view_shelter`` permission.
 
     Raises:
-        ``ObjectDoesNotExist`` when the shelter is not found or the user
-        does not belong to its organization.
+        ``Shelter.DoesNotExist`` when the shelter is not found.
```
mikefeldberg: raises `ObjectDoesNotExist`. same for `room_update`

apps/betterangels-backend/shelters/services/shelter.py
```diff
@@ -168,27 +152,37 @@ def shelter_create(*, user: "User", data: Dict[str, Any]) -> Shelter:
...
+    ``change_shelter`` permission, so the caller does not need to
+    pre-lookup the entity.
+
     Only fields present in *data* (i.e. not ``UNSET``) are modified.
     Schedules and services use full-replacement semantics when provided.
 
     Raises:
-        ``Shelter.DoesNotExist`` when no shelter matches the given ID.
-        ``PermissionError`` when the user is not a member of the shelter's organization.
+        ``Shelter.DoesNotExist`` when no shelter matches the given ID
```
mikefeldberg: `ObjectDoesNotExist`

apps/betterangels-backend/accounts/emails.py
```diff
@@ -0,0 +1,45 @@
...
+    }
+
+    msg = EmailMultiAlternatives(
+        subject=f"Welcome to BetterAngels, {user.first_name}!",
+        body=render_to_string(template_config.welcome_txt, context),
+        to=[user.email or ""],
+    )
+    html_body = render_to_string(template_config.welcome_html, context)
+    msg.attach_alternative(html_body, "text/html")
+    msg.send(fail_silently=True)
```
mikefeldberg: why `fail_silently=True`?

apps/betterangels-backend/shelters/services/shelter.py
```diff
@@ -136,21 +129,12 @@ def resolve_pending_service_entries(entries: list[tuple[int, str]]) -> list[Serv
...
-    creating anything.
-
     Accepts a plain dict (e.g. from ``strawberry.asdict(data)`` with
     ``UNSET`` keys already removed).
 
     Raises:
-        ``PermissionError`` when the user is not a member of the org.
         ``django.core.exceptions.ValidationError`` on invalid data.
     """
-    org_id = data.get("organization")
```
mikefeldberg, line 149: is there a reason `shelter_create` gets org_id from mutation vars while `shelter_update` gets it from the header? This is the only service we don't explicitly pass `organization_id` to. afaict, this effectively allows a user with "create shelter" perm to create a shelter for any org. e.g., x-org-id header has the ID of the user's actual org; data.org has another org's ID

apps/betterangels-backend/shelters/tests/test_mutations.py
```diff
@@ -45,7 +31,7 @@ def test_create_shelter_minimal_fields(self) -> None:
             "data": {
                 "name": "Test Shelter",
                 "description": "A test shelter for unit testing",
-                "organization": str(self.org_1.pk),
+                "organization": str(self.org.pk),
```
mikefeldberg: related to the [create_shelter comment](https://github.com/BetterAngelsLA/monorepo/pull/2143/changes#r3455076021), I think `organization` needs to be removed from all create tests and `CreateShelterInput`

apps/betterangels-backend/shelters/services/shelter.py
mikefeldberg: ya, verified

```python
def test_create_shelter_other_org(self) -> None:
    """Test creating a shelter with only required fields"""
    other_org = organization_recipe.make(name="not my org")

    mutation = """
        mutation ($data: CreateShelterInput!) {
            createShelter(data: $data) {
                ... on ShelterType {
                    id
                    name
                    description
                    status
                }
            }
        }
    """

    variables: dict[str, Any] = {
        "data": {
            "name": "Test Shelter",
            "description": "A test shelter for unit testing",
            "organization": str(other_org.pk),
        }
    }

    initial_shelter_count = Shelter.objects.count()

    response = self.execute_graphql(mutation, variables)
    shelter = response["data"]["createShelter"]

    updated_shelter_count = Shelter.objects.count()
    self.assertEqual(updated_shelter_count, initial_shelter_count + 1)

    bad_shelter = Shelter.objects.get(pk=shelter["id"])
    self.assertEqual(bad_shelter.organization_id, other_org.pk)
```
passes

mikefeldberg (commented):
about halfway through. first batch of comments

apps/betterangels-backend/shelters/services/shelter.py
```diff
@@ -126,15 +126,23 @@ def resolve_pending_service_entries(entries: list[tuple[int, str]]) -> list[Serv
 
 
 @transaction.atomic
-def shelter_create(*, user: "User", data: Dict[str, Any]) -> Shelter:
+def shelter_create(*, user: "User", organization_id: str, data: Dict[str, Any]) -> Shelter:
     """Create a new Shelter with all M2M relationships and schedules.
 
+    *organization_id* (from the ``X-Organization-ID`` header, validated by
+    ``HasOrgPerm``) is used as the owning organization.  Any
+    ``organization`` key in *data* is overwritten so the caller cannot
```
mikefeldberg: why include organization on the input at all?

apps/betterangels-backend/shelters/services/shelter_photo.py
```diff
@@ -125,9 +120,15 @@ def resolve_uploads(
...
+def delete_shelter_photos(*, user: "User", organization_id: str, ids: list[int]) -> list[int]:
+    """Delete shelter photos scoped to *organization_id*.
+
+    Only photos belonging to shelters in the active organization are
+    eligible for deletion.
+    """
+    org_shelters = shelter_queryset(user=user, organization_id=organization_id)
     photos = ShelterPhoto.objects.filter(
-        shelter__in=user_shelter_list(Shelter.objects.all(), user=user),
+        shelter__in=org_shelters,
```
mikefeldberg: why the change to the broader list?

mikefeldberg (approved):

apps/betterangels-backend/shelters/services/shelter.py
mikefeldberg: if we're leaving it, maybe add a test that a user can't create a shelter for an org other than the header id org e.g.

```python
def test_create_shelter_fails_for_other_orgself) -> None:
    """Test creating a shelter fails for org other than X-Organization-ID header org"""
    other_org = organization_recipe.make(name="not my org")

    mutation = """
        mutation ($data: CreateShelterInput!) {
            createShelter(data: $data) {
                ... on ShelterType {
                    id
                }
            }
        }
    """

    variables: dict[str, Any] = {
        "data": {
            "name": "Test Shelter",
            "description": "A test shelter for unit testing",
            "organization": str(other_org.pk),
        }
    }

    initial_shelter_count = Shelter.objects.count()

    with self.assertRaises:
        response = self.execute_graphql(mutation, variables)
    
    self.assertEqual(Shelter.objects.count(), initial_shelter_count)
```

PR #1960, opened 2026-03-15
[BP] Update Client Models

nathanaela2002:
### What changed:                                                          
  - Updated the models to match [client attributes](https://www.notion.so/uwblueprintexecs/Client-Attributes-28b10f3fb1dc80b9b454f61183a5af4f)
  - Added new client profile fields: adl_capacity, criminal_history,     
  destination, funding_source, harm_reduction, homelessness_notes,       
  income_annual, income_source, justice_involvement_details,      
  medical_needs, medical_notes, pets, requires_transportation,           
  social_security_number, spa, status                         
  - Changed sexual_orientation from a single string field to a
  multi-select array                                          
  - Renamed GenderEnum values: MALE to CIS_MALE, FEMALE to CIS_FEMALE    
  - Added SexualOrientationEnum with values: Asexual, Bisexual, Gay, 
  Straight, Lesbian, Pansexual, Queer, Other, Prefer Not to Say          
  - Added data migration to update existing records from male/female to  
  cis_male/cis_female                                                    
  - Regenerated schema.graphql and all frontend/mobile GraphQL types     
                                                                    
 ### Why the graphql-inspector:approved-breaking-change label was added:    
  The GraphQL Inspector CI check failed because two breaking changes were
   detected. These are intentional:                                      
  1. GenderEnum -- MALE/FEMALE removed in favor of CIS_MALE/CIS_FEMALE.  
  The mobile app already uses the new values so no mobile changes are    
  needed.                                                            
  2. Three ServiceRequest mutations removed (createServiceRequest,       
  updateServiceRequest, removeNoteServiceRequest) -- these were part of a
   backend refactor. Note: the mobile app still calls these in 3 screens 
  (NoteCreateHmis, NoteEditHmis, ServicesModal), so a follow-up mobile PR
   is needed before deploying to production.    

## Summary by Sourcery

Expand client and HMIS profile data models and enums to capture additional demographic, medical, funding, status, and document metadata, while keeping database history triggers in sync and making schema changes backwards compatible.

New Features:
- Add new client profile attributes for status, legal/justice involvement, medical needs, income, pets, transportation, funding sources, SPA, social security number, ADL capacity, destination, and sexual orientation to GraphQL schema, Django models, HMIS profiles, and admin/export views.
- Introduce new enums for client status, ADL capacity, medical needs, client pets, funding sources, SPA regions, destinations, and sexual orientation, and extend existing enums such as language, living situation, race, social media platform, HMIS agency, and client document namespaces.
[...]

PR #2144, opened 2026-06-16
feat: add status field to ClientProfile

julianchen24:
## What

Adds a `status` field to `ClientProfile`, backed by a new `ClientStatusEnum` (`reserved` / `checked_in` / `checked_out`), and exposes it through the GraphQL `ClientProfileType` plus the create/update inputs.

## Why

The reservation work in #2116 (`updateReservationStatus`) needs to cascade the linked client's status to **RESERVED**, but `ClientProfile` had no status field. BA previously paused broad client-model changes; per discussion with @frank731 on #2116, this is the agreed **minimal** client-model addition to unblock that cascade.

The enum values and field mirror the larger client-model work in #1960, so this stays consistent with the planned design.

## Details

- **Default `CHECKED_OUT`** (non-nullable). A brand-new client isn't reserved or in a shelter, so `CHECKED_OUT` is the accurate baseline — the migration backfills existing rows to it. (Chosen over #1960's `default=RESERVED`, which would mislabel every existing client.)
- Field placed alphabetically in `AbstractClientProfile` (between `spoken_languages` and `unhoused_start_date`), matching the file convention.
- Migration also regenerates the three pghistory triggers to include the new column (standard pghistory behavior).
- Regenerated `schema.graphql` and the frontend `__generated__` GraphQL types.

## Tests

- `test_update_client_profile_status_mutation` — verifies create defaults to `CHECKED_OUT` and an update transitions it to `RESERVED`.
- `status` added to the shared client-profile query selection; create/update/query expected fixtures updated accordingly.

> ⚠️ The DB-backed Django test suite and `mypy` were not runnable in my local sandbox (no Postgres/PostGIS container; pinned mypy plugin stack unavailable) — relying on CI for those. Locally verified: `manage.py check`, schema export in sync, `makemigrations`, flake8/black/isort, frontend codegen, and `react-shelter-operator` tests all pass.

## Follow-up

Once merged, #2116 can implement the client cascade: `reservation.client_profile.status = ClientStatusEnum.RESERVED`.

## Summary by Sourcery

Add a status field to client profiles and expose it through the GraphQL API to support client status workflows.

New Features:
- Introduce ClientStatusEnum with RESERVED, CHECKED_IN, and CHECKED_OUT values for client profiles.
- Add a non-nullable status field to ClientProfile with a default of CHECKED_OUT, including historical event tracking.
- Expose client status on the ClientProfile GraphQL type and in create/update input types for backend and frontend consumers.

Build:
- Regenerate GraphQL schema and frontend generated types to include the new client status field and enum.

Tests:
- Extend client profile mutation and query tests to cover the new status field, including defaulting and status updates.


yanxue06:
hello a beautiful pr this is @julianchen24

PR #2147, opened 2026-06-17
shelter reservations backend (SDB-237)

mikefeldberg:
SDB-237


vecchp (commented):
## PR Review Summary

This is an architectural refactor replacing stored status fields on Bed/Room with computed status derived from Reservation data — a sound direction. Excellent test coverage (~1,200 lines).

**Key issues to address before merge:**
- RoomFilter.IN_TURNAROUND has two bugs (wrong bed filter + missing last_cleaned comparison)
- Reservation.save() does unnecessary DB lookups on every call
- ReservationType.shelter uses assert (stripped in optimized Python)
- Destructive migration drops data with no RunPython step
- ~400 lines of duplicated Bed/Room status logic
- Broken report_bed_status_counts left in codebase

**Compatibility note:** Upcoming PR #2143 renames AdminShelterType → OperatorShelterType and migrates @HasPerm → @HasOrgPerm. The new reservation schema code will need corresponding updates.

apps/betterangels-backend/shelters/types/filters.py
```diff
@@ -222,21 +238,246 @@ class BedFilter(CommonBedRoomFilterMixin):
...
+
+                if choice == BedStatusChoices.OCCUPIED:
+                    has_checked_in = Q(
+                        Exists(
+                            models.Reservation.objects.filter(
+                                bed_id=OuterRef("pk"),
+                                status=ReservationStatusChoices.CHECKED_IN,
+                            )
+                        )
+                    )
```
vecchp: 🔴 **Bug: RoomFilter.IN_TURNAROUND** — Two issues:

1. `bed__isnull=False` excludes room-only reservations (where `bed=None`). A room with only room-only completed reservations can be in turnaround but won't be found.

2. No comparison against `last_cleaned`. A room cleaned AFTER checkout still appears as IN_TURNAROUND. The model property `Room.computed_status` correctly checks `last_cleaned <= last_checkout`, and `_rooms_in_turnaround_subquery()` in outputs.py does this correctly.

**Fix:** Align this filter with `_rooms_in_turnaround_subquery()` — use a Subquery to get the latest checkout and compare against `last_cleaned`, and drop `bed__isnull=False`.

apps/betterangels-backend/shelters/migrations/0008_clean_up_models.py
```diff
@@ -0,0 +1,244 @@
+# Generated by Django 6.0.6 on 2026-06-16 22:16
```
vecchp: 🔴 **Data loss risk:** This migration drops `Bed.status`, `Bed.occupant`, `Room.status`, `Room.occupants`, and `Reservation.shelter` columns with **no `RunPython` data migration**. If any environment has existing data in these columns, it will be silently destroyed.\n\nIf this is intentional (clean-slate), please document it clearly in the migration docstring. Otherwise, add a `RunPython` step that converts existing `Bed.status` → Reservation-backed computed equivalents and backfills `Reservation.shelter` from `bed.shelter` / `room.shelter` before dropping columns.

apps/betterangels-backend/shelters/models/shelter.py
```diff
@@ -222,37 +229,81 @@ class Bed(CloneMixin, BaseModel):
         "medical_needs",
```
vecchp, line 231: 🟡 **Style (HackSoftware Django-Styleguide):** The `computed_status` property and `_get_last_completed_checkout()` are copy-pasted identically between `Bed` and `Room` with only variable renames (same logic, same prefetch-awareness, same DB queries through `self.reservations.filter()`).\n\nPer the style guide: "keep models lean — push logic to services, queries to selectors." These properties contain business rules, DB queries, and GraphQL prefetch-awareness (`hasattr(self, \"_active_reservations\")`) — tight coupling to the GraphQL layer.\n\n**Recommendation (follow-up):** Extract a shared `selectors/computed_status.py` with a single function for both Bed and Room.

apps/betterangels-backend/shelters/services/reservation.py
```diff
@@ -0,0 +1,114 @@
...
+    reservation.full_clean()
+    reservation.save()
+    _set_clients(reservation, clients_data)
+
+    return reservation
+
+
+@transaction.atomic
+def reservation_delete(*, user: "User", ids: list[int]) -> list[int]:
+    """Delete reservations by their IDs and return the deleted IDs.
```
vecchp: 🟡 **Docstring inaccuracy:** The docstring says "Raises ObjectDoesNotExist when any of the given IDs does not match", but the actual behavior silently skips unmatched IDs and returns only successfully deleted ones. This is consistent with `bed_delete`/`room_delete`, but the docstring should reflect reality.

apps/betterangels-backend/shelters/schema.py
```diff
@@ -68,6 +72,15 @@ class Query:
         extensions=[HasPerm(Bed.perms.VIEW)],
```
vecchp: 🟡 **Forward-looking (PR #2143 compatibility):** Upcoming PR #2143 by @vecchp migrates all shelter queries/mutations from `@hasPerm` (global) to `@hasOrgPerm` (org-scoped via `X-Organization-ID` header) and renames `AdminShelterType` → `OperatorShelterType`.\n\nThe new reservation queries and mutations here use `@HasPerm` — coordinate on merge order. If #2143 merges first, update to `@HasOrgPerm` + use `get_current_organization(info)` to read org from the header instead of deriving it from bed/room.

apps/betterangels-backend/shelters/types/filters.py
vecchp: 🔴 **Bug: RoomFilter.IN_TURNAROUND** — Two issues:

1. `bed__isnull=False` excludes room-only reservations (where `bed=None`). A room with only room-only completed reservations can be in turnaround but won't be found.

2. No comparison against `last_cleaned`. A room cleaned AFTER checkout still appears as IN_TURNAROUND. The model property `Room.computed_status` correctly checks `last_cleaned <= last_checkout`, and `_rooms_in_turnaround_subquery()` in outputs.py does this correctly.

**Suggested fix — align with `_rooms_in_turnaround_subquery()`:**

```python
elif choice == RoomStatusChoices.IN_TURNAROUND:
    latest_completed_checkout = Subquery(
        models.Reservation.objects.filter(
            room_id=OuterRef("pk"),
            status=ReservationStatusChoices.COMPLETED,
            checked_out_at__isnull=False,
        )
        .order_by("-checked_out_at")
        .values("checked_out_at")[:1],
        output_field=DateTimeField(),
    )
    q |= (
        Q(maintenance_flag=False)
        & Q(
            Exists(
                models.Reservation.objects.filter(
                    room_id=OuterRef("pk"),
                    status=ReservationStatusChoices.COMPLETED,
                    checked_out_at__isnull=False,
                )
            )
        )
        & (Q(last_cleaned__isnull=True) | Q(last_cleaned__lt=latest_completed_checkout))
    )
```

apps/betterangels-backend/shelters/models/reservation.py
```diff
@@ -35,22 +37,111 @@ class Reservation(BaseModel):
...
+        """Return True if the assigned bed is in turnaround (last checkout after last cleaned)."""
+        if not self.bed:
+            return False
+        last_checkout = self.bed._get_last_completed_checkout()
+        if not last_checkout:
+            return False
+        return self.bed.last_cleaned is None or self.bed.last_cleaned <= last_checkout
+
+    def _is_room_in_turnaround(self) -> bool:
+        """Return True if the assigned room is in turnaround (last checkout after last cleaned)."""
```
vecchp: 🔴 **Performance: Reservation.save() does unnecessary DB lookups.** Every `save()` calls `Reservation.objects.get(pk=self.pk)` to check previous status. On creates this raises `DoesNotExist` (caught silently — wasted query). On updates it's an extra round-trip.

**Suggested fix — use `__init__`-level tracking instead:**

```python
def __init__(self, *args: Any, **kwargs: Any) -> None:
    super().__init__(*args, **kwargs)
    self._initial_status = self.status

def save(self, *args: Any, **kwargs: Any) -> None:
    if self.pk is not None and self.status != self._initial_status:
        if (
            self.status == ReservationStatusChoices.COMPLETED
            and self._initial_status != ReservationStatusChoices.COMPLETED
        ):
            self.checked_out_at = timezone.now()
        elif (
            self.status == ReservationStatusChoices.CHECKED_IN
            and self._initial_status != ReservationStatusChoices.CHECKED_IN
        ):
            self.checked_in_at = timezone.now()
    super().save(*args, **kwargs)
    self._initial_status = self.status
```

This avoids the `DoesNotExist` catch on creates and eliminates the extra query on updates.

apps/betterangels-backend/shelters/types/outputs.py
```diff
@@ -320,18 +539,85 @@ def get_queryset(cls, queryset: QuerySet, info: Info) -> QuerySet[models.Room]:
...
+    last_cleaned: auto
     last_cleaned_inspected: auto
     maintenance_flag: bool
     medical_respite: auto
     name: auto
     notes: auto
     pets: List[PetType]
     shelter: "ShelterType"
-    status: Optional[RoomStatusChoices]
     storage: bool
```
vecchp, line 361: 🔴 **Production risk: `assert` used for control flow.** Python `assert` statements are stripped when running with `-O` (optimized mode). If both `root.bed` and `root.room` are `None`, this will crash with `AttributeError` in production.

**Suggested fix — replace `assert` with explicit guard:**

```python
@strawberry_django.field(select_related=["bed__shelter", "room__shelter"])
def shelter(self, root: models.Reservation) -> "AdminShelterType":
    if root.bed:
        return cast("AdminShelterType", root.bed.shelter)
    if root.room is None:
        raise ValueError(
            f"Reservation {root.pk} has neither bed nor room assigned."
        )
    return cast("AdminShelterType", root.room.shelter)
```

This gives a clear error message with the reservation PK for debugging, and works correctly under `-O` (unlike `assert`).

apps/betterangels-backend/shelters/models/reservation.py
vecchp: 🔴 **Performance: Reservation.save() does unnecessary DB lookups.**

Every `save()` calls `Reservation.objects.get(pk=self.pk)` to check the previous status. On creates this raises `DoesNotExist` (caught silently — wasted query). On updates it's an extra round-trip since Django already tracks initial values from `__init__`.

**Fix:** Use `__init__`-level tracking instead:

```python
def __init__(self, *args, **kwargs):
    super().__init__(*args, **kwargs)
    self._initial_status = self.status

def save(self, *args, **kwargs):
    if self.pk is not None and self.status != self._initial_status:
        if self.status == ReservationStatusChoices.COMPLETED:
            self.checked_out_at = timezone.now()
        elif self.status == ReservationStatusChoices.CHECKED_IN:
            self.checked_in_at = timezone.now()
    super().save(*args, **kwargs)
    self._initial_status = self.status
```

apps/betterangels-backend/shelters/types/outputs.py
vecchp, line 361: 🔴 **Production risk: `assert` used for control flow.**

Python `assert` statements are stripped when running with `-O` (optimized mode). If both `root.bed` and `root.room` are `None`, this will crash with `AttributeError: 'NoneType' object has no attribute 'shelter'` in production.

**Fix:** Replace with an explicit check:

```python
if root.room is None:
    raise ValueError("Reservation has neither bed nor room")
return cast("AdminShelterType", root.room.shelter)
```

vecchp:
## 🏗️ Refactor Opportunity: Collapse computed status logic into a single `Case`/`When` annotation

The computed status priority chain is a simple 5-condition waterfall:

```
1. maintenance_flag=True              → OUT_OF_SERVICE
2. last_checkout > last_cleaned       → IN_TURNAROUND
3. has CHECKED_IN reservation         → OCCUPIED
4. has CONFIRMED/OVERDUE reservation  → RESERVED
5. otherwise                          → AVAILABLE
```

This exact chain is currently implemented **10+ times** across the codebase — in `Bed.computed_status`, `Room.computed_status`, 10 subquery helpers in `outputs.py`, ~140 lines of filter conditions in `filters.py`, and `Reservation.clean()` validation.

All of these can collapse into a single shared annotation function:

```python
# shelters/selectors/computed_status.py (new)
from django.db.models import Case, DateTimeField, Exists, F, OuterRef, Q, Subquery, Value, When
from shelters.enums import BedStatusChoices, ReservationStatusChoices
from shelters.models.shelter import ACTIVE_RESERVATION_STATUSES

def _latest_completed_checkout_subquery(reservable_field: str) -> Subquery:
    """Most recent completed checkout for a reservable (bed or room)."""
    from shelters.models import Reservation
    return Subquery(
        Reservation.objects.filter(
            **{f"{reservable_field}_id": OuterRef("pk")},
            status=ReservationStatusChoices.COMPLETED,
            checked_out_at__isnull=False,
        )
        .order_by("-checked_out_at")
        .values("checked_out_at")[:1],
        output_field=DateTimeField(),
    )

def annotate_computed_status(queryset, *, reservable_type: str = "bed"):
    """Annotate a Bed or Room QuerySet with ``_computed_status``."""
    from shelters.models import Reservation
    field = f"{reservable_type}_id"
    StatusEnum = BedStatusChoices if reservable_type == "bed" else RoomStatusChoices

    latest_cleaned = _latest_completed_checkout_subquery(field)

    has_checked_in = Exists(
        Reservation.objects.filter(**{field: OuterRef("pk")}, status=ReservationStatusChoices.CHECKED_IN)
    )
    has_confirmed_or_overdue = Exists(
        Reservation.objects.filter(**{field: OuterRef("pk")}, status__in=[
            ReservationStatusChoices.CONFIRMED, ReservationStatusChoices.CHECK_IN_OVERDUE
        ])
    )
    has_active = Exists(
        Reservation.objects.filter(**{field: OuterRef("pk")}, status__in=ACTIVE_RESERVATION_STATUSES)
    )

    in_turnaround = Q(last_cleaned__isnull=True) | Q(last_cleaned__lt=F("_latest_checkout"))

    return queryset.annotate(
        _latest_checkout=latest_cleaned,
        _computed_status=Case(
            When(maintenance_flag=True, then=Value(StatusEnum.OUT_OF_SERVICE)),
            When(in_turnaround, then=Value(StatusEnum.IN_TURNAROUND)),
            When(has_checked_in, then=Value(StatusEnum.OCCUPIED)),
            When(has_confirmed_or_overdue, then=Value(StatusEnum.RESERVED)),
            default=Value(StatusEnum.AVAILABLE),
        ),
    )
```

**What this replaces:**

| Old | New |
|---|---|
| 10 subquery helpers (~250 lines) | `annotate_computed_status(qs).filter(_computed_status=...)` |
| `BedFilter.status` (~70 lines) | `.annotate(...).filter(_computed_status__in=value)` |
| `RoomFilter.status` (~70 lines) | `.annotate(...).filter(_computed_status__in=value)` |

**Result:** ~400 lines → ~50 lines, single source of truth for the status chain, and filters become trivial `_computed_status__in=[...]` lookups. The model properties (`Bed.computed_status`, `Room.computed_status`) would call this annotation for consistency.

This is a follow-up optimization — not a merge blocker. The current implementation is correct; this just consolidates the duplicated logic into one place.

vecchp (approved):

apps/betterangels-backend/shelters/models/reservation.py
```diff
@@ -35,22 +38,112 @@ class Reservation(BaseModel):
...
-        shelter_pk = self.shelter.pk
-        if self.room and shelter_pk is not None and self.room.shelter_id != shelter_pk:
-            errors["room"] = "The selected room must belong to the same shelter as the reservation."
-        if self.bed and shelter_pk is not None and self.bed.shelter_id != shelter_pk:
-            errors["bed"] = "The selected bed must belong to the same shelter as the reservation."
+
+        errors: dict[str, str] = {}
+
+        if not self.bed and not self.room:
+            errors[NON_FIELD_ERRORS] = "A reservation must have a bed or room assigned."
```
vecchp, line 66: [Must-fix] Missing bed/room shelter consistency validation

When `Reservation.shelter` FK was removed, the old consistency checks (`bed.shelter_id == shelter_pk`, `room.shelter_id == shelter_pk`) were dropped. Nothing now prevents a reservation with `bed` from shelter A and `room` from shelter B.

**Recommended fix** -- add after the requires-bed-or-room check on line 65:

```python
def clean(self) -> None:
    super().clean()
    errors: dict[str, str] = {}

    if not self.bed and not self.room:
        errors[NON_FIELD_ERRORS] = "A reservation must have a bed or room assigned."

    # ADD THIS BLOCK:
    if self.bed and self.room:
        if self.bed.shelter_id != self.room.shelter_id:
            errors[NON_FIELD_ERRORS] = "Bed and room must belong to the same shelter."

    # If bed is assigned, validate bed
    if self.bed and not errors:
        ...
```

**Also add a test**:
```python
def test_rejects_bed_and_room_from_different_shelters(self) -> None:
    other_shelter = baker.make(Shelter)
    other_bed = baker.make(Bed, shelter=other_shelter)
    with self.assertRaises(ValidationError) as ctx:
        Reservation(bed=other_bed, room=self.room_1).clean()
    self.assertIn("same shelter", str(ctx.exception))
```

vecchp, line 66: [Must-fix] Missing bed-room belonging check

Even if bed and room share a shelter, there is no validation that the bed is physically in that room. A reservation could have `bed=A` (in room C) and `room=B` -- a data integrity gap.

**Recommended fix** -- add after the shelter consistency check:

```python
    if self.bed and self.room:
        if self.bed.shelter_id != self.room.shelter_id:
            errors[NON_FIELD_ERRORS] = "Bed and room must belong to the same shelter."
        # ADD THIS CHECK:
        elif self.bed.room_id and self.bed.room_id != self.room.pk:
            errors["bed"] = "The selected bed does not belong to the selected room."
```

This also means the existing test `test_multiple_bed_reservations_allowed_for_same_room` (line 66 in test_reservation_models.py) needs fixing -- it puts `bed_2` (which is in `room_2`) into a reservation with `room=room_1`.

apps/betterangels-backend/shelters/models/reservation.py
```diff
@@ -35,22 +38,112 @@ class Reservation(BaseModel):
...
+        return is_in_turnaround(
+            last_cleaned=self.room.last_cleaned,
+            last_checkout=get_last_completed_checkout(self.room.reservations.all()),
+        )
+
+    def __init__(self, *args: Any, **kwargs: Any) -> None:
+        super().__init__(*args, **kwargs)
+        self._initial_status = self.status
+
+    def save(self, *args: Any, **kwargs: Any) -> None:
```
vecchp: [Must-fix] HackSoft style guide: business logic in save()

The HackSoft Django style guide explicitly says: "Business logic should not live in: Model save method."

The `checked_in_at`/`checked_out_at` auto-timestamping belongs in the service layer. This also lets you remove the `__init__` override (line 123) which only exists to track `_initial_status` for `save()`.

**Step 1** -- Move logic to `reservation_update` service:

```python
# Add import at top of services/reservation.py:
from django.utils import timezone

# In reservation_update, after setattr loop:
previous_status = reservation.status

for key, value in data.items():
    if value is not None:
        setattr(reservation, key, value)

# Auto-set timestamps on status transitions
new_status = data.get("status")
if new_status is not None and new_status != previous_status:
    if new_status == ReservationStatusChoices.COMPLETED:
        reservation.checked_out_at = timezone.now()
    elif new_status == ReservationStatusChoices.CHECKED_IN:
        reservation.checked_in_at = timezone.now()
```

**Step 2** -- Remove from model:
- `def __init__(self, *args, **kwargs): ...` (lines 123-125)
- `def save(self, *args, **kwargs): ...` (lines 127-141)
- imports: `from typing import Any`, `from django.utils import timezone`

apps/betterangels-backend/shelters/models/reservation.py
```diff
@@ -35,22 +38,112 @@ class Reservation(BaseModel):
...
+                name="unique_active_reservation_per_bed",
+            ),
+            models.UniqueConstraint(
+                fields=["room"],
+                condition=models.Q(status__in=ACTIVE_RESERVATION_STATUSES, bed__isnull=True),
+                name="unique_active_reservation_per_room",
+            ),
         ]
 
     def clean(self) -> None:
```
vecchp, line 60: [Must-fix] HackSoft style guide: complex cross-model validation in clean()

The guide says: *"Validation should be moved to the service layer if: spanning relations & fetching additional data is required."*

`clean()` queries `Reservation.objects`, calls `get_last_completed_checkout()`, accesses `bed.reservations.all()` -- all cross-model DB operations.

**Step 1** -- Thin out model `clean()` to simple field checks only:

```python
def clean(self) -> None:
    super().clean()
    errors: dict[str, str] = {}

    if not self.bed and not self.room:
        errors[NON_FIELD_ERRORS] = "A reservation must have a bed or room."

    if self.bed and self.room:
        if self.bed.shelter_id != self.room.shelter_id:
            errors[NON_FIELD_ERRORS] = "Bed and room must belong to the same shelter."
        elif self.bed.room_id and self.bed.room_id != self.room.pk:
            errors["bed"] = "The selected bed does not belong to the selected room."

    if errors:
        raise ValidationError(errors)
```

**Step 2** -- Add complex validation to the service:

```python
# In services/reservation.py -- add these imports:
from shelters.models.shelter import ACTIVE_RESERVATION_STATUSES
from shelters.status import get_last_completed_checkout, is_in_turnaround

def _validate_reservation(reservation):
    """Validate availability, turnaround, and conflicts."""
    errors = {}
    if reservation.bed:
        _validate_reservable(errors, reservation, reservation.bed, "bed")
    if reservation.room and not reservation.bed:
        _validate_reservable(errors, reservation, reservation.room, "room", room_only=True)
    if errors:
        raise ValidationError(errors)

def _validate_reservable(errors, reservation, reservable, field_name, *, room_only=False):
    if reservable.maintenance_flag:
        errors[field_name] = f"The selected {field_name} is out of service."
        return
    if is_in_turnaround(last_cleaned=reservable.last_cleaned,
                       last_checkout=get_last_completed_checkout(reservable.reservations.all())):
        errors[field_name] = f"The selected {field_name} is currently in turnaround."
        return
    conflicting_qs = Reservation.objects.filter(
        **{f"{field_name}_id": reservable.pk},
        status__in=ACTIVE_RESERVATION_STATUSES)
    if room_only:
        conflicting_qs = conflicting_qs.filter(bed__isnull=True)
    if reservation.pk:
        conflicting_qs = conflicting_qs.exclude(pk=reservation.pk)
    if conflicting_qs.exists():
        errors[field_name] = f"This {field_name} already has an active reservation."
```

**Step 3** -- Call from create and update:
```python
# In reservation_create:
reservation.full_clean()
_validate_reservation(reservation)  # ADD
reservation.save()

# In reservation_update:
reservation.full_clean()
_validate_reservation(reservation)  # ADD
reservation.save()
```

**Step 4** -- Remove from model: `_is_bed_in_turnaround` (lines 105-112) and `_is_room_in_turnaround` (lines 114-121)

apps/betterangels-backend/shelters/managers.py
```diff
@@ -37,3 +58,220 @@ def open_at(
...
+
+
+def in_turnaround_filter_q(fk_field: str) -> Q:
+    """``Q`` for in-turnaround without a prior ``annotate`` (uses correlated subqueries)."""
+    has_completed_checkout = Q(has_completed_checkout_exists(fk_field))
+    needs_cleaning = Q(last_cleaned__isnull=True) | Q(last_cleaned__lte=latest_completed_checkout_subquery(fk_field))
+    return has_completed_checkout & needs_cleaning
+
+
+def status_filter_q(
```
vecchp: [Must-fix] HackSoft style guide: domain logic in custom managers/querysets

The guide says: *"Business logic should not live in: Custom managers or querysets."*

These functions encode the status priority chain and should be in selectors:
- `status_filter_q` (line 136)
- `computed_status_case` (line 165)
- `in_turnaround_filter_q` (line 129)
- `latest_completed_checkout_subquery` (line 70)
- `has_checked_in_exists` (line 81)
- `has_confirmed_or_overdue_exists` (line 90)
- `has_completed_checkout_exists` (line 100)
- `has_active_reservation_exists` (line 109)
- `bed_computed_status_annotation` (line 180)
- `room_computed_status_annotation` (line 184)
- `shelter_bed_status_count_subquery` (line 270)
- `shelter_room_status_count_subquery` (line 275)

**Recommended**: Move to new `shelters/selectors/computed_status.py`. Keep only the thin `BedManager`/`RoomManager`/`ReservableStatusQuerySetMixin` wrappers in `managers.py`.

apps/betterangels-backend/common/permissions/utils.py
```diff
@@ -265,42 +268,48 @@ def permissioned_queryset(
...
         queryset = queryset.filter(
-            Exists(
-                Organization.objects.filter(
-                    pk=OuterRef(organization_field),
-                    users=user,
-                )
-            )
+            reduce(or_, (Q(Exists(Organization.objects.filter(pk=OuterRef(f), users=user))) for f in fields))
         )
     elif any_perm:
```
vecchp, line 308: [Should-fix] Duplicated reduce(or_, ...) expression -- extract helper

The `elif any_perm` and `else` branches build the same nested expression (~10 lines each).

**Extract a helper**:

```python
def _org_perm_exists_across_fields(user, app_label, codename, fields):
    return reduce(or_, (
        Q(Exists(
            Organization.objects.filter(pk=OuterRef(f))
            .filter(permission_groups__group__user=user)
            .filter(_perm_q(app_label, codename))
        ))
        for f in fields
    ))
```

**Then simplify**:
```python
# elif any_perm:
q = Q()
for perm_str in perms:
    app_label, codename = perm_str.split(".", 1)
    q |= _org_perm_exists_across_fields(user, app_label, codename, fields)
queryset = queryset.filter(q)

# else:
for perm_str in perms:
    app_label, codename = perm_str.split(".", 1)
    queryset = queryset.filter(
        _org_perm_exists_across_fields(user, app_label, codename, fields))
```

Saves ~20 lines of duplicated code.

apps/betterangels-backend/shelters/services/reservation.py
```diff
@@ -0,0 +1,119 @@
...
+
+    bed_id = data.get("bed_id")
+    room_id = data.get("room_id")
+
+    clients_data = data.pop("clients", None)
+    if not clients_data:
+        raise ValidationError("At least one client must be associated with a reservation.")
+
+    if bed_id:
+        shelter_id = bed_get(user=user, organization_id=organization_id, bed_id=bed_id).shelter_id
```
vecchp: [Should-fix] Redundant shelter_get call

`bed_get`/`room_get` already validate org membership via `permissioned_queryset`. The `shelter_get` + try/except block (lines 54-57) is redundant -- a bed can't belong to a shelter the user can't access.

**Before**:
```python
if bed_id:
    shelter_id = bed_get(...).shelter_id
elif room_id:
    shelter_id = room_get(...).shelter_id
else:
    raise ObjectDoesNotExist("A bed or room must be provided...")

try:
    shelter_get(user=user, organization_id=organization_id, shelter_id=shelter_id)
except Shelter.DoesNotExist:
    raise ObjectDoesNotExist("You do not have permission...")
```

**After**:
```python
if bed_id:
    bed_get(user=user, organization_id=organization_id, bed_id=bed_id)
elif room_id:
    room_get(user=user, organization_id=organization_id, room_id=room_id)
else:
    raise ObjectDoesNotExist("A bed or room must be provided...")
```

apps/betterangels-backend/shelters/managers.py
```diff
@@ -37,3 +58,220 @@ def open_at(
...
+
+    base = Q(maintenance_flag=False)
+
+    if status == status_enum.OCCUPIED:
+        return base & Q(has_checked_in_exists(fk_field))
+
+    if status == status_enum.RESERVED:
+        return base & Q(has_confirmed_or_overdue_exists(fk_field)) & ~Q(has_checked_in_exists(fk_field))
+
+    not_turnaround = ~in_turnaround_filter_q(fk_field)
```
vecchp: [Should-fix] Eager not_turnaround computation

`~in_turnaround_filter_q(fk_field)` on line 154 is computed for every status but only used in the `AVAILABLE` branch (line 160). This builds expensive correlated subqueries for OUT_OF_SERVICE, OCCUPIED, and RESERVED unnecessarily.

**Before**:
```python
not_turnaround = ~in_turnaround_filter_q(fk_field)
...
if status == status_enum.AVAILABLE:
    return base & not_turnaround & ~Q(has_active_reservation_exists(fk_field))
```

**After**:
```python
if status == status_enum.AVAILABLE:
    return base & ~in_turnaround_filter_q(fk_field) & ~Q(has_active_reservation_exists(fk_field))
```

apps/betterangels-backend/shelters/models/reservation.py
```diff
@@ -35,22 +38,112 @@ class Reservation(BaseModel):
...
+            elif (
+                self.status == ReservationStatusChoices.CHECKED_IN
+                and self._initial_status != ReservationStatusChoices.CHECKED_IN
+            ):
+                self.checked_in_at = timezone.now()
+
+        super().save(*args, **kwargs)
+        self._initial_status = self.status
+
     def __str__(self) -> str:
```
vecchp, line 86: [Should-fix] Duplicate shelter derivation -- add @property shelter

The pattern `self.bed.shelter if self.bed else self.room.shelter if self.room else "No Shelter"` appears in `__str__` (line 143) and `ReservationType.shelter` (outputs.py).

**Add to model**:
```python
@property
def shelter(self) -> Shelter | None:
    if self.bed_id:
        return self.bed.shelter
    if self.room_id:
        return self.room.shelter
    return None
```

Then simplify `__str__`:
```python
def __str__(self) -> str:
    shelter = self.shelter or "No Shelter"
    return f"Reservation #{self.pk} at {shelter} ({self.status})"
```

And `ReservationType.shelter`:
```python
def shelter(self, root) -> "OperatorShelterType":
    if root.shelter is None:
        raise ValueError(f"Reservation {root.pk} has neither bed nor room assigned.")
    return cast("OperatorShelterType", root.shelter)
```

Note: use `self.bed_id` not `self.bed` to avoid unnecessary lazy FK lookups.

apps/betterangels-backend/shelters/models/reservation.py
```diff
@@ -35,22 +38,112 @@ class Reservation(BaseModel):
...
+                )
+                if self.pk:
+                    conflicting_qs = conflicting_qs.exclude(pk=self.pk)
+                if conflicting_qs.exists():
+                    errors["room"] = "This room already has an active room-only reservation."
+
         if errors:
             raise ValidationError(errors)
 
+    def _is_bed_in_turnaround(self) -> bool:
```
vecchp: [Should-fix] _is_bed_in_turnaround / _is_room_in_turnaround re-query reservations

Both call `get_last_completed_checkout(reservable.reservations.all())`, hitting the DB per instance. If `clean()` runs both the turnaround check AND the conflict check (which also queries reservations), this is redundant DB work.

This is naturally fixed by moving validation to the service (see review above). In the service helper `_validate_reservable`, the reservations queryset can be evaluated once and reused for both checks. After the move, these two model methods are removed entirely.

apps/betterangels-backend/common/permissions/utils.py
```diff
@@ -233,6 +235,7 @@ def permissioned_queryset(
     perms: Sequence[str] | None = None,
     any_perm: bool = True,
     organization_field: str = "organization_id",
```
vecchp, line 259: [Should-fix] organization_fields parameter not documented

Add to the docstring between `organization_field` and `Returns`:

```
    organization_fields : list[str] | None
        Multiple field paths (OR'd together). Use when a model reaches
        its organization through more than one path (e.g. Reservation
        via ``bed__shelter__organization_id`` or
        ``room__shelter__organization_id``). Takes precedence over
        *organization_field*. Default ``None``.
```

apps/betterangels-backend/shelters/selectors/operator.py
```diff
@@ -131,6 +132,39 @@ def bed_queryset(
...
+    user: "User",
+    organization_id: str,
+    perms: list[str] | None = None,
+) -> "QuerySet[Reservation]":
+    """Scope *queryset* to *organization_id* where *user* belongs to the shelter's org.
+
+    Reservation links to a shelter via two optional paths
+    (``bed__shelter`` and ``room__shelter``), so the standard
+    ``permissioned_queryset`` helper (which expects a single
+    ``organization_field``) is not suitable.
```
vecchp: [Should-fix] reservation_queryset docstring now misleading

The current docstring says the standard helper *"is not suitable"* -- but `permissioned_queryset` now supports `organization_fields`.

**Before**:
> Reservation links to a shelter via two optional paths
> (``bed__shelter`` and ``room__shelter``), so the standard
> ``permissioned_queryset`` helper (which expects a single
> ``organization_field``) is not suitable.

**After**:
> Reservation links to a shelter via two optional paths
> (``bed__shelter`` and ``room__shelter``), so we pass
> ``organization_fields`` to ``permissioned_queryset`` to check both.

apps/betterangels-backend/shelters/migrations/0008_clean_up_models.py
```diff
@@ -0,0 +1,256 @@
+# Generated by Django 6.0.6 on 2026-06-18 21:08
+
+import django_choices_field.fields
+import pgtrigger.compiler
+import pgtrigger.migrations
+from django.conf import settings
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
```
vecchp, line 10: [Should-fix] No docstring about destructive migration

This migration drops `Bed.status`, `Bed.occupant`, `Room.status`, `Room.occupants`, and `Reservation.shelter` with **no RunPython data migration**. If any environment has data in these columns, it will be silently destroyed.

If this is intentional (clean-slate assumption), add a docstring:

```python
class Migration(migrations.Migration):
    """Drop stored status/occupant/shelter columns in favor of computed status.

    No RunPython data migration -- assumes a clean environment with no
    production data to preserve in the dropped columns.
    """
```

apps/betterangels-backend/shelters/models/shelter.py
```diff
@@ -211,69 +219,96 @@ def save(self, *args: Any, **kwargs: Any) -> None:
...
     type = TextChoicesField(choices_enum=BedTypeChoices, blank=True, null=True)
 
     class Meta:
         indexes = [
-            models.Index(fields=["shelter", "status"]),
+            models.Index(fields=["shelter"]),
         ]
 
+    @property
+    def computed_status(self) -> BedStatusChoices:
```
vecchp, line 265: [Nice-to-have] N+1 risk on computed_status fallback

When `_computed_status` annotation is missing (e.g., accessed outside `with_computed_status()`), the property falls back to:

```python
active = self.reservations.filter(status__in=ACTIVE_RESERVATION_STATUSES)
return compute_bed_status(
    last_checkout=get_last_completed_checkout(self.reservations.all()),
    active_reservation_statuses={r.status for r in active},
)
```

In a list context without annotation, this is N+1. No code change needed -- just always use `with_computed_status()` for querysets. The fallback is there for single-object paths (tests, admin detail view).

apps/betterangels-backend/shelters/tests/test_selectors.py
```diff
@@ -211,6 +217,7 @@ def _dt(y: int, m: int, d: int) -> datetime.datetime:
 _STATUS_KEYS = ("available", "occupied", "reserved", "out_of_service")
 
 
+@skip("status removed from Bed and Room")
```
vecchp: [Nice-to-have] @skip on ReportBedStatusCountsTestCase

```python
@skip("status removed from Bed and Room")
class ReportBedStatusCountsTestCase(TestCase):
```

Should be either restored with a computed-status equivalent in a follow-up PR, or removed entirely if no longer relevant.

apps/betterangels-backend/shelters/models/shelter.py
```diff
@@ -211,69 +219,96 @@ def save(self, *args: Any, **kwargs: Any) -> None:
...
+    objects = RoomManager()
+
     _clone_linked_m2m_fields = [
         "accessibility",
         "demographics",
         "funders",
         "pets",
     ]
-
+    _clone_excluded_fields = [
```
vecchp, line 294: [Nice-to-have] Room.clone() now excludes notes and status_notes

```python
_clone_excluded_fields = [
    "last_cleaned", "last_cleaned_inspected",
    "maintenance_flag",
    "notes",          # NEWLY excluded -- previously cloneable
    "status_notes",   # NEWLY excluded
]
```

Previously `notes` was cloneable. Confirm this is intentional -- cloned rooms will lose their custom notes. If unintentional, remove `"notes"` and `"status_notes"` from the list.

apps/betterangels-backend/shelters/models/shelter.py
```diff
@@ -211,69 +219,96 @@ def save(self, *args: Any, **kwargs: Any) -> None:
...
+
     _clone_linked_m2m_fields = [
         "accessibility",
         "demographics",
         "funders",
         "medical_needs",
         "pets",
     ]
-    _clone_excluded_fields = ["occupant"]
+    _clone_excluded_fields = [
```
vecchp, line 234: [Nice-to-have] Bed.clone() now excludes maintenance_flag

```python
_clone_excluded_fields = [
    "last_cleaned", "last_cleaned_inspected",
    "maintenance_flag",  # NEWLY excluded -- cloned beds always AVAILABLE
    "status_notes",
]
```

Cloned beds always start with `maintenance_flag=False` (default, AVAILABLE). This is good behavior -- cloned beds should be fresh. Worth a brief comment: `# Excluded so cloned beds always start AVAILABLE`.

apps/betterangels-backend/shelters/managers.py
```diff
@@ -37,3 +58,220 @@ def open_at(
...
+
+def bed_computed_status_annotation() -> Case:
+    return computed_status_case("bed_id", BedStatusChoices)
+
+
+def room_computed_status_annotation() -> Case:
+    return computed_status_case("room_id", RoomStatusChoices)
+
+
+class ReservableStatusQuerySetMixin:
```
vecchp, line 50: [Nice-to-have] cast(QuerySet, self) typing workaround

Several methods in `ReservableStatusQuerySetMixin` use `cast(QuerySet, self)` to silence MRO typing issues. A `Generic[T]` mixin would be cleaner but this works. Low priority.

apps/betterangels-backend/shelters/types/filters.py
```diff
@@ -222,21 +239,69 @@ class BedFilter(CommonBedRoomFilterMixin):
...
+
+@strawberry_django.filter_type(models.Reservation)
+class ReservationFilter:
+    id: Optional[ID]
+    room_id: Optional[ID]
+    bed_id: Optional[ID]
+    status = make_in_filter("status", ReservationStatusChoices)
+
+    @strawberry_django.filter_field
+    def shelter_id(self, info: Info, value: Optional[ID], prefix: str) -> Q:
```
vecchp, line 295: [Nice-to-have] Asymmetry: filter resolves to shelter, queryset to organization

`ReservationFilter.shelter_id` uses:
```python
return Q(**{f"{prefix}bed__shelter_id": value}) | Q(**{f"{prefix}room__shelter_id": value})
```
This filters by a **specific shelter ID**.

Meanwhile `reservation_queryset` uses:
```python
organization_fields=["bed__shelter__organization_id", "room__shelter__organization_id"]
```
This scopes to an **organization**. Different paths for different purposes -- both correct. Worth noting for future readers but no change needed.

apps/betterangels-backend/shelters/tests/test_reservation_models.py
```diff
@@ -0,0 +1,213 @@
...
+
+    # --- double-booking prevention: room-only ---
+
+    def test_cannot_create_second_room_only_reservation(self) -> None:
+        """Room-only (bed=null) is unique per room for active reservations."""
+        baker.make(Reservation, room=self.room_2, bed=None, status=ReservationStatusChoices.CONFIRMED)
+        with self.assertRaises(IntegrityError):
+            baker.make(Reservation, room=self.room_2, bed=None)
+
+    def test_multiple_bed_reservations_allowed_for_same_room(self) -> None:
```
vecchp, line 65: [Should-fix] Broken test: test_multiple_bed_reservations_allowed_for_same_room

`bed_2` is created in `room_2` (line 18 in `setUp`), but this test puts it into a reservation with `room=room_1` -- the bed is in a room it doesn't belong to. This only passes because the missing validation (#1, #2) hasn't been fixed.

**Fix the test**:
```python
def test_multiple_bed_reservations_allowed_for_same_room(self) -> None:
    """When both room and bed are set, the room constraint does not apply."""
    bed_2_in_room_1 = baker.make(Bed, shelter=self.shelter, room=self.room_1, name="Bed-2")
    baker.make(Reservation, room=self.room_1, bed=self.bed_1, status=ReservationStatusChoices.CONFIRMED)
    baker.make(Reservation, room=self.room_1, bed=bed_2_in_room_1, status=ReservationStatusChoices.CONFIRMED)
    self.assertEqual(Reservation.objects.filter(room=self.room_1).count(), 2)
```

**Also add a cross-shelter rejection test**:
```python
def test_rejects_bed_and_room_from_different_shelters(self) -> None:
    other_shelter = baker.make(Shelter)
    other_bed = baker.make(Bed, shelter=other_shelter)
    with self.assertRaises(ValidationError) as ctx:
        Reservation(bed=other_bed, room=self.room_1).clean()
    self.assertIn("same shelter", str(ctx.exception))
```

vecchp (changes requested):
## Comprehensive inline review — 21 findings

### 🔴 Must-fix (4)
1. Missing bed/room shelter consistency validation
2. Missing bed↔room belonging check
3. HackSoft: business logic in `save()` → move to service
4. HackSoft: complex validation in `clean()` → move to service

### 🟡 Should-fix (8)
5. Domain logic in `managers.py` → move to selectors
6. Duplicated `reduce(or_, ...)` in `permissioned_queryset`
7. Redundant `shelter_get` in `reservation_create`
8. Eager `not_turnaround` computation
9. Duplicate shelter derivation → add `@property`
10. Turnaround helpers re-query reservations
11. `_perm_q` is private but imported cross-module
12. Missing `organization_fields` docstring
13. Misleading `reservation_queryset` docstring
14. No docstring about destructive migration
21. Broken test exposes missing validation

### 🟢 Nice-to-have (5)
15-20. N+1 awareness, skipped test, clone exclusions, typing workaround, filter asymmetry

### ✅ Confirmed fixed
- `assert` → `ValueError` in `ReservationType.shelter`
- IN_TURNAROUND filter aligned with `_rooms_in_turnaround_subquery`
- `last_cleaned` comparison in turnaround check

**Note**: `_perm_q` at `common/permissions/utils.py:188` is prefixed with underscore but imported in `shelters/managers.py`. A public alias `perm_filter` already exists — managers should use that, or rename `_perm_q` to `perm_q`.

vecchp (approved):

PR #2153, opened 2026-06-19
feat(backend): shelter operator user management

vecchp:
## Summary
Backend changes for shelter operator user management (squashed from `sdb-178-shelter-operator-user-management`).

Includes: accounts (permissions, emails, extensions, schema, services, types), settings, urls, shelters (groups, managers, open_at, schema, services, tests), common (org_types, tests), reports, schema.graphql, docs.

## Related
- Frontend: #TBD (sdb-178-shelter-operator-frontend)


apps/betterangels-backend/accounts/types.py
```diff
@@ -247,6 +247,14 @@ class OrganizationMemberType(UserBaseType):
     def member_role(self, info: Info) -> OrgRoleEnum:
         return OrgRoleEnum(getattr(self, "_member_role", OrgRoleEnum.MEMBER.value))
 
+    @strawberry_django.field
+    def is_org_owner(self, info: Info) -> bool | None:
+        """Whether this member is the organization owner."""
+        value = getattr(self, "_is_org_owner", None)
+        if value is None:
+            return None
```
tglaz: any reason to not make it a pure `bool` and return `False` here?

apps/betterangels-backend/common/middleware.py
```diff
@@ -0,0 +1,44 @@
...
+
+    References:
+        Django Time Zones documentation:
+        https://docs.djangoproject.com/en/5.0/topics/i18n/timezones/#selecting-the-current-time-zone
+    """
+
+    def __init__(self, get_response: Callable[[HttpRequest], HttpResponse]) -> None:
+        self.get_response = get_response
+
+    def __call__(self, request: HttpRequest) -> HttpResponse:
```
tglaz: i'm curious if we actually want this - though seems like used with report dates?

always wary of setting any TZ info on the backend - it's nice to know everything is UTC always 
- can pass TZ explicitly in BE
- the FE can read a timezone cookie

apps/betterangels-backend/shelters/selectors.py
```diff
@@ -0,0 +1,248 @@
...
+from shelters.enums import BedStatusChoices, DayOfWeekChoices, ScheduleTypeChoices, StatusChoices
+
+if TYPE_CHECKING:
+    from accounts.models import User
+    from django.contrib.auth.base_user import AbstractBaseUser
+    from django.contrib.auth.models import AnonymousUser
+    from shelters.models import Bed, Room, Shelter
+
+
+def report_bed_status_counts(
```
tglaz: 1. i'd suggest adding the word `daily` into the name
2. i'm not sure this is the right place for it
- feels more of a reporting query than a selector
- perhaps `/shelters/reporting/get_daily_bed_status_counts.py`, else `/shelters/reporting.py`
- some of the report queries may get large and may be nice to split up - especially for parallel test files

apps/betterangels-backend/shelters/selectors.py
```diff
@@ -0,0 +1,248 @@
...
+    *, shelter: "Shelter", start_date: datetime.date, end_date: datetime.date
+) -> list[dict[str, Any]]:
+    """
+    Returns daily bed status counts for each day in the range.
+
+    Each BedEvent is valid from its ``pgh_created_at`` until the *next*
+    event for the same bed (or forever if no next event).  A ``bed.remove``
+    event ends the bed's lifecycle.  By annotating each event with its
+    successor's timestamp we can answer "what was the status on day X?"
```
tglaz: "what was the status `at the end of day` X?"

apps/betterangels-backend/shelters/selectors.py
```diff
@@ -0,0 +1,248 @@
...
+        )
+        .exclude(pgh_label="bed.remove")
+        .annotate(next_event_at=Subquery(next_event_subq))
+        .order_by("pgh_created_at")
+        .values("pgh_obj_id", "status", "pgh_created_at", "next_event_at")
+    )
+
+    n_days = (end_date - start_date).days + 1
+
+    def _counts(day: datetime.date) -> dict[str, Any]:
```
tglaz: should we define the returned type?

apps/betterangels-backend/shelters/selectors.py
```diff
@@ -0,0 +1,248 @@
...
+    """
+    from shelters.models import BedEvent  # type: ignore[attr-defined]
+
+    end_of_range = datetime.datetime.combine(end_date, datetime.time.max, tzinfo=datetime.timezone.utc)
+
+    # Subquery: the *next* BedEvent (by pgh_created_at) for the same bed.
+    next_event_subq = (
+        BedEvent.objects.filter(
+            pgh_obj_id=OuterRef("pgh_obj_id"),
+            pgh_created_at__gt=OuterRef("pgh_created_at"),
```
tglaz: could we have events for same bed with same `pgh_created_at`?
edge case perhaps, but figured i'd ask and see if can be consequential. 
probably the only scenario I'd think of is where status is changed at same time as bed.remove somehow, so the next event will be missed?

apps/betterangels-backend/shelters/selectors.py
```diff
@@ -0,0 +1,248 @@
...
+        .exclude(pgh_label="bed.remove")
+        .annotate(next_event_at=Subquery(next_event_subq))
+        .order_by("pgh_created_at")
+        .values("pgh_obj_id", "status", "pgh_created_at", "next_event_at")
+    )
+
+    n_days = (end_date - start_date).days + 1
+
+    def _counts(day: datetime.date) -> dict[str, Any]:
+        eod = datetime.datetime.combine(day, datetime.time.max, tzinfo=datetime.timezone.utc)
```
tglaz: Is the new TimezoneMiddleware intended to affect these date boundaries? 
i'd lean towards passing TZ into the function explicitly

apps/betterangels-backend/shelters/selectors.py
```diff
@@ -0,0 +1,248 @@
...
+            "available": counts.get(BedStatusChoices.AVAILABLE, 0),
+            "occupied": counts.get(BedStatusChoices.OCCUPIED, 0),
+            "reserved": counts.get(BedStatusChoices.RESERVED, 0),
+            "out_of_service": counts.get(BedStatusChoices.OUT_OF_SERVICE, 0),
+        }
+
+    return [_counts(start_date + datetime.timedelta(days=n)) for n in range(n_days)]
+
+
+def shelter_list(
```
tglaz: perhaps just be explicit in the name and call it `approved_shelter_list`?

apps/betterangels-backend/shelters/selectors.py
```diff
@@ -0,0 +1,248 @@
...
+    """
+    from shelters.models import Shelter
+
+    queryset = queryset.filter(status=StatusChoices.APPROVED)
+    if user and user.is_authenticated and hasattr(user, "has_perm") and user.has_perm(Shelter.perms.VIEW_PRIVATE):
+        return queryset
+    return queryset.filter(is_private=False)
+
+
+def admin_shelter_list(queryset: "QuerySet[Shelter]", *, user: "User") -> "QuerySet[Shelter]":
```
tglaz: maybe rename to `user_organization_shelter_list`?

`admin` can have different meanings. for example, would this work for BA "super" admins who could view all orgs? not how perms would work there - via a single `VIEW_ALL_ORGANIZATIONS` permission or more granular

apps/betterangels-backend/shelters/selectors.py
```diff
@@ -0,0 +1,248 @@
...
+            Q(day=None) | Q(day=day),
+            Q(start_date=None) | Q(start_date__lte=date),
+            Q(end_date=None) | Q(end_date__gte=date),
+        )
+    )
+
+    return queryset.filter(is_open).exclude(has_active_exception)
+
+
+def reservation_status_change_counts(
```
tglaz: similar to the `report_bed_status_counts` comment about perhaps moving to a `reporting` dir

apps/betterangels-backend/shelters/selectors.py
```diff
@@ -0,0 +1,248 @@
...
+    # to a single Schedule row.
+    from shelters.models import Schedule
+
+    is_open = Exists(
+        Schedule.objects.filter(
+            shelter=OuterRef("pk"),
+            schedule_type=schedule_type,
+            is_exception=False,
+            start_time__lte=time,
+            end_time__gte=time,
```
tglaz: not sure if we need to handle this, but...
does this need to work with schedules where end_time is after midnight? 
not sure if we allow a shelter's hours be from 5pm - 2am.
in that case this could happen `end_time < time`

apps/betterangels-backend/shelters/selectors.py
```diff
@@ -0,0 +1,248 @@
...
+            Q(day=None) | Q(day=day),
+            Q(start_date=None) | Q(start_date__lte=date),
+            Q(end_date=None) | Q(end_date__gte=date),
+        )
+    )
+
+    # Step 2: exclude shelters with an active exception covering *dt*.
+    #   - Full-day:  start_time IS NULL  → closed all day.
+    #   - Partial:   start_time <= time AND end_time >= time  → closed now.
+    covers_now = Q(start_time__isnull=True) | Q(start_time__lte=time, end_time__gte=time)
```
tglaz: should end_time be inclusive or `end_time__gt=time`?

tglaz (changes requested):

apps/betterangels-backend/shelters/open_at.py
```diff
@@ -39,27 +66,24 @@ def shelters_open_at(
...
             Q(end_date=None) | Q(end_date__gte=date),
         )
     )
 
     # Step 2: exclude shelters with an active exception covering *dt*.
     #   - Full-day:  start_time IS NULL  → closed all day.
-    #   - Partial:   start_time <= time AND end_time >= time  → closed now.
-    covers_now = Q(start_time__isnull=True) | Q(start_time__lte=time, end_time__gte=time)
+    #   - Partial:   same time+day logic as above, including overnight.
+    covers_now = Q(start_time__isnull=True) | time_day
```
tglaz: is this valid?

> Q(start_time__isnull=True) bypasses the day check. Since day=None means “every day” and a populated day means a specific weekday, should full-day exceptions also use Q(day=None) | Q(day=day)? As written, a full-day Monday exception with an active date range would exclude the shelter on every weekday.

this will currently fail:

```python
def test_open_at_full_day_exception_respects_day(self) -> None:
    shelter = Shelter.objects.create(name="Exception Shelter")

    Schedule.objects.create(
        shelter=shelter,
        schedule_type=ScheduleTypeChoices.OPERATING,
        day=None,
        start_time=datetime.time(0, 0),
        end_time=datetime.time(23, 59),
    )

    Schedule.objects.create(
        shelter=shelter,
        schedule_type=ScheduleTypeChoices.OPERATING,
        day=DayOfWeekChoices.MONDAY,
        start_time=None,
        end_time=None,
        is_exception=True,
    )

    monday = datetime.datetime(2026, 3, 2, 12, 0)
    tuesday = datetime.datetime(2026, 3, 3, 12, 0)

    self.assertNotIn(shelter, Shelter.objects.open_at(monday))
    self.assertIn(shelter, Shelter.objects.open_at(tuesday))
```

apps/betterangels-backend/shelters/selectors/operator.py
```diff
@@ -57,7 +57,7 @@ def operator_shelter_list(
     user: "User",
     organization_id: str,
 ) -> "QuerySet[Shelter]":
-    """Filter to shelters belonging to *organization_id* that *user* is a member of."""
+    """Filter to shelters in *organization_id* that *user* belongs to."""
```
mikefeldberg: undo change

mikefeldberg: undo change

mikefeldberg (approved):
just one docstring that got probably got reverted during the merge resolution

tglaz (approved):

PR #2155, opened 2026-06-19
feat(frontend): teams pagination, drawer redesign, table alignment, stale org cleanup

vecchp:
## Frontend: Dynamic Teams

Based on `pv-selah-update-teams-backend`.

### Admin App
- Teams page: offset pagination via GraphQL, `tableClassName`, sortable headers, slug column removed
- AddTeamDrawer: redesigned with Input component, disabled state
- Table component: `tableClassName` prop, right-aligned last column
- useActiveOrgState: stale org cleanup, localStorage sync
- useAllauthLogin: send email instead of username
- TeamPermissions enum values updated to PascalCase (Add/Change/Delete/View)
- Removed PageLayout wrapper from Teams page

### Expo App
- teamId/teamIds migration for dynamic teams
- FilterTeams component with search
- UserTeamPreference: team selector with persistence
- Notes/Tasks screens updated for `teams.results` offset pagination
- TaskForm team selector, NoteForm team picker
- Generated GraphQL types updated throughout

### Files Changed
89 files across `apps/betterangels-admin/`, `apps/betterangels/`, `libs/expo/`, `libs/react/`


apps/betterangels-admin/src/app/Layout/AppSidebar.tsx
```diff
@@ -125,6 +126,16 @@ export function AppSidebar(props: IProps) {
             Reports
           </Sidebar.Link>
         )}
+        {hasPermission(TeamPermissions.View) && (
+          <Sidebar.Link
+            to="/teams"
+            isActive={location.pathname === '/teams'}
```
mikefeldberg, line 137: maybe a good time to implement a routePaths.ts. e.g., [libs/react/shelter/src/lib/constants/routePaths.ts](https://github.com/BetterAngelsLA/monorepo/blob/main/libs/react/shelter/src/lib/constants/routePaths.ts). `/teams` is used four times in three places (as are the other paths). Also ok to leave for follow up since it's oos

libs/expo/betterangels/src/lib/ui-components/TaskForm/TaskForm.tsx
```diff
@@ -37,6 +39,8 @@ export function TaskForm(props: TProps) {
     mode: 'onChange',
   });
 
+  const { data } = useQuery<TeamsQuery>(TeamsDocument);
```
mikefeldberg, line 42: this query is broken throughout because the org header isn't being set. `info.context.request.organization_id` is `None`. Probably `useActiveOrgState.ts` 


---

## Analysis

There are **2 real bugs** and **1 dead-code issue** in `useActiveOrgState.ts`:

### Bug 1 (Critical — the root cause of your header problem): `useState` initializer clears valid localStorage when `organizations` is empty

```typescript
// Line 18-39: useState initializer
() => {
    const stored = localStorage.getItem(storageKey);
    if (stored && organizations.some((o) => o.id === stored)) {
      return stored;                        // ✅ valid → return
    }
    if (stored) {
      localStorage.removeItem(storageKey);  // ❌ BUG: deletes valid org
    }
    const fallback = organizations[0]?.id;
    ...
}
```

When `organizations` starts empty (`[]`) during async loading:
1. Reads `'org2'` from localStorage (user's real preference)
2. `[].some(...)` → `false` → assumes stored org is **stale**
3. **Deletes `'org2'` from localStorage**
4. Returns `undefined`

Then when organizations load (`[{id:'org1'}, {id:'org2'}]`):
- The `useEffect` re-runs, finds localStorage empty, falls back to `organizations[0]?.id` → `'org1'`
- **User loses their `'org2'` preference and is silently switched to `'org1'`**

Meanwhile, any GraphQL requests that fire between the delete and the `useEffect` re-write will lack the `X-Organization-ID` header entirely, because `orgLink` reads from localStorage.

### Bug 2 (Dead code): Unreachable reconciliation check

```typescript
// Line 48-54 in useEffect
const stored = localStorage.getItem(storageKey);
if (stored && organizations.some((o) => o.id === stored)) {
    setActiveOrgIdState(stored);
    if (stored !== localStorage.getItem(storageKey)) {  // ❌ NEVER true
      localStorage.setItem(storageKey, stored);
    }
    return;
}
```

`stored` was just retrieved from the same key — the comparison is always equal.

---

## Proposed Fix

In the `useState` initializer, **don't treat an empty `organizations` list as confirmation that stored orgs are stale**. Only clear localStorage when we have organizations loaded and can actually validate:

```typescript
// Before clearing, confirm organizations is non-empty so we know
// the stored org is truly stale — not just that data hasn't loaded yet.
if (stored && organizations.length > 0 && !organizations.some((o) => o.id === stored)) {
    localStorage.removeItem(storageKey);
}
```

And also re-write localStorage when the stored value is already valid (belt-and-suspenders sync):

```typescript
if (stored && organizations.some((o) => o.id === stored)) {
    localStorage.setItem(storageKey, stored);  // ensure sync
    return stored;
}
```

Remove the dead code on lines 52-54.

Would you like me to switch to ACT MODE and implement these changes?

apps/betterangels-admin/src/app/routes/appRoutes.tsx
```diff
@@ -33,7 +34,11 @@ export const routeChildren: RouteObject[] = [
   },
   {
     path: '/teams',
-    element: <TeamsPage />,
+    element: (
+      <PermissionGuard permission={TeamPermissions.View}>
```
mikefeldberg, line 38: should have been caught on backend review: `teams` query isn't behind `HasOrgPerm`

PR #2162, opened 2026-06-23
[BP] Add BarChart wrapper component

Angus-Sun:
## Summary

Adds a reusable `BarChart` wrapper component over `@ant-design/plots` Column with design-system styling applied by default.

## Changes

- New `BarChart` component in `libs/react/components`
- Storybook stories demonstrating single-series and stacked usage

## Features

- Design-system defaults: Poppins font, dashed gridlines, styled axis labels/titles
- `chartTitle` prop renders a title above the chart
- `showViewToggle` renders a Count/Percentage pill toggle
- `onViewChange(mode)` callback — return different data/axis/scale config per mode
- HTML color legend auto-renders for multi-series (`colorField`) charts
- Hover darkens the active bar; shared tooltip across a column

### Single-series

```tsx
<BarChart
  chartTitle="Daily Occupancy"
  showViewToggle
  onViewChange={(mode) => (mode === 'percentage' ? { data: pctData, scale: { y: { domain: [0, 100], nice: false } } } : {})}
  data={countData}
  xField="date"
  yField="count"
  axis={{
    x: { title: 'Date' },
    y: { title: 'Number of Beds Occupied', tickCount: 6 },
  }}
  scale={{
    x: { padding: 0.4 },
    y: { nice: true },
  }}
  style={{ fill: '#008CEE' }}
  tooltip={{ title: 'date' }}
/>
```

### Stacked (multi-series)

Pass `colorField`, `stack: true`, and `scale.color` with explicit `domain`/`range` to control color order. The component will auto-render a color legend above the chart.

Use `style: { stroke: '#ffffff', lineWidth: 1, inset: 0.1 }` to add a white separator between stacked segments.

```tsx
<BarChart
  chartTitle="Bed Status"
  showViewToggle
  onViewChange={(mode) => (mode === 'percentage' ? { data: pctData, scale: { y: { domain: [0, 100], nice: false } } } : {})}
  data={countData}
  xField="date"
  yField="count"
  colorField="status"
  stack
  scale={{
    color: {
      domain: ['Occupied', 'Available', 'Reserved', 'Out of Service'],
      range: ['#008CEE', '#05B428', '#FF7B00', '#F64949'],
    },
    x: { padding: 0.4 },
    y: { nice: true },
  }}
  axis={{
    x: { title: 'Date' },
    y: { title: 'Status by Count', tickCount: 6 },
  }}
  style={{
    stroke: '#ffffff',
    lineWidth: 1,
    inset: 0.1,
  }}
  tooltip={{ title: 'date' }}
/>
```

> **Note:** Always provide `scale.color.domain` explicitly when using `colorField` — if omitted, color order is inferred from the order values appear in `data`, which may not be consistent with real API responses.

## Testing

View stories in Storybook under **Web Components → BarChart**:

- `VerticalBarChart` — single-series with toggle
- `StackedBarChart` — multi-color stacked with legend and toggle

### Stacked Bar Chart — Count view
Screenshot 2026-06-22 194646

### Stacked Bar Chart — Count view with tooltip

Screenshot 2026-06-22 194711

### Stacked Bar Chart — Percentage view
Screenshot 2026-06-22 194733

### Single-series Bar Chart — Count view with tooltip
Screenshot 2026-06-22 194807

### Single-series Bar Chart — Percentage view
Screenshot 2026-06-22 194824


libs/react/components/src/lib/BarChart/BarChart.tsx
```diff
@@ -0,0 +1,339 @@
...
+    },
+  };
+
+  const defaultAxis = (defaultConfig.axis ?? {}) as {
+    x?: object;
+    y?: object;
+  };
+  const configAxis = (config.axis ?? {}) as { x?: object; y?: object };
+
+  const mergedConfig = {
```
frank731: Wondering if this behaviour can be done using something like mergeDeep from import { mergeDeep } from 'remeda'; to cut down a lot of this merge code

libs/react/components/src/lib/BarChart/BarChart.tsx
```diff
@@ -0,0 +1,339 @@
...
+      : [];
+
+  return (
+    <div
+      className={className}
+      style={{ display: 'flex', flexDirection: 'column', width: '100%' }}
+    >
+      {hasHeader && (
+        <div
+          style={{
```
frank731: Instead of using inline styles, we usually use Tailwind classes, look at components in `/libs/react/shelter-operator/src/lib/components` as reference

libs/react/components/src/lib/BarChart/BarChart.tsx
```diff
@@ -0,0 +1,339 @@
...
+                  {label}
+                </span>
+              ))}
+            </div>
+          )}
+        </div>
+      )}
+      <div style={{ flex: 1, minHeight: 0 }}>
+        <Column
+          key={`${viewMode}-${fontsReady ? 'fonts-ready' : 'fonts-loading'}`}
```
frank731: sucks that we need to make a loaded font check in this component, what happens when this isn't here?

frank731 (changes requested):
Component looks good! couple things need to refactor though, mainly that our components live in `/libs/react/shelter-operator/src/lib/components` so files will need to be moved over (should also show in our shelter operator storybook folder) and that we use Tailwind for styling

libs/react/components/src/lib/BarChart/BarChart.tsx
Angus-Sun: Oops, I think in earlier iterations it would mess up the formatting if Poppins hadn't loaded yet. It seems to be fine now though. I'll remove the check

frank731 (approved):

PR #2163, opened 2026-06-23
refactor(shelters): use cyclic-interval modulo arithmetic for schedule open-at queries

vecchp:
## Summary

Replaces the `is_normal`/`is_overnight` branching in `shelters_open_at` with precomputed database columns and modular arithmetic: a single, branchless condition handles normal, overnight, and week-wrapping schedules identically.

## Problem

The current `_time_and_day_condition` function has 6 branches to handle the same-day vs next-day distinction for overnight schedules (e.g. 6 PM – 2 AM). Each call site must pass both `day` and `yesterday` parameters. This is error-prone and hard to reason about.

## Solution

Two `GENERATED ALWAYS ... STORED` columns on the `Schedule` model, kept in sync by the database:

- **`start_week_minutes`** — week-minute offset (0–10079), or daily offset (0–1439) for every-day schedules
- **`duration_minutes`** — positive duration via `(end − start + 1440) % 1440`, handling overnight naturally

The open-at query becomes a single modulo check:

```
(query − start_week_minutes + modulus) % modulus < duration_minutes
```

No `is_normal`, no `is_overnight`, no `day`/`yesterday` parameters.

## Changes

| File | Change |
|---|---|
| `shelters/models/schedule.py` | Add `start_week_minutes` / `duration_minutes` `GeneratedField` columns + `_time_to_minutes` helper |
| `shelters/open_at.py` | Rewrite `shelters_open_at` with `Mod`/`LessThan` — zero branches |
| `shelters/services/utils.py` | No changes needed (generated columns are DB-managed) |
| `migrations/0008_*` | Add generated columns to `schedule` and `scheduleevent` tables |

## Testing

All 276 shelter tests pass.


apps/betterangels-backend/shelters/models/schedule.py
```diff
@@ -3,13 +3,91 @@
...
+    """
+    day_offset = Case(
+        When(day="monday", then=Value(0 * DAILY_MINUTES)),
+        When(day="tuesday", then=Value(1 * DAILY_MINUTES)),
+        When(day="wednesday", then=Value(2 * DAILY_MINUTES)),
+        When(day="thursday", then=Value(3 * DAILY_MINUTES)),
+        When(day="friday", then=Value(4 * DAILY_MINUTES)),
+        When(day="saturday", then=Value(5 * DAILY_MINUTES)),
+        When(day="sunday", then=Value(6 * DAILY_MINUTES)),
+        default=Value(0),
```
tglaz, line 58: minor naming, but this makes `start_week_minutes` a bit untrue, as offset is relative to midnight (not Monday midnight)
if want to, options can be `start_offset_minutes` or `start_cycle_minutes`....

apps/betterangels-backend/shelters/open_at.py
```diff
@@ -21,15 +26,62 @@ def shelters_open_at(
...
+    Uses precomputed ``start_week_minutes`` / ``duration_minutes`` columns
+    with modular arithmetic so that overnight schedules (e.g. 6 PM – 2 AM)
+    and normal schedules (9 AM – 5 PM) are handled by the same branchless
+    condition — no is_normal / is_overnight distinction needed.
+
+    **How the modulo check works**
+
+    Every schedule is represented as a cyclic interval
+    ``[start_week_minutes, start_week_minutes + duration_minutes)`` on a
+    10080-minute clock (Monday 00:00 = 0, Sunday 23:59 = 10079).  For a
```
tglaz, line 39: minor: but may be worth adding something like:

```
Start time is inclusive, end time is exclusive. Close time of 15:00 means closed at 15:00
```

apps/betterangels-backend/shelters/open_at.py
```diff
@@ -21,15 +26,62 @@ def shelters_open_at(
...
+    ) & Q(
+        LessThan(
+            Mod(Value(query_wm) - F("start_week_minutes") + WEEK_MINUTES, WEEK_MINUTES),
+            F("duration_minutes"),
+        ),
+    )
+    timed_covers = every_day_window | day_specific_window
+
+    # Full-day schedules (start_time IS NULL) match purely by day.
+    full_day_covers = Q(start_time__isnull=True) & (Q(day__isnull=True) | Q(day=day_enum))
```
tglaz, line 86: we ignore `end_time` here - which works.
do we have a constraint that `start_time` and `end_time` are either both populated or both null?

apps/betterangels-backend/shelters/models/schedule.py
```diff
@@ -3,13 +3,91 @@
...
 from .lookups import Demographic
 from .shelter import Shelter
 
+DAILY_MINUTES = 24 * 60  # 1440
+
+# ── Helpers ───────────────────────────────────────────────────────────────
+
+
+def _time_to_minutes(time_field: str) -> ExpressionWrapper:
+    """``EXTRACT(HOUR FROM time) * 60 + EXTRACT(MINUTE FROM time)`` as integer.
```
tglaz, line 20: probably nothing to do - but our schedules inherently ignore seconds+
pretty expected, but still an assumption we could document at the model

not sure if there is a good boundary to truncate start/end time so we avoid any downstream issues related to timestamp having seconds - we handle this here, but we have to remember it in any future functions

tglaz (changes requested):

tglaz (approved):

PR #2167, opened 2026-06-24
feat(frontend): teams pagination, drawer redesign, table alignment, stale org cleanup (SDB-178)

vecchp:
## Frontend: Dynamic Teams

Based on `pv-selah-update-teams-backend`.

### Admin App
- Teams page: offset pagination via GraphQL, `tableClassName`, sortable headers, slug column removed
- AddTeamDrawer: redesigned with Input component, disabled state
- Table component: `tableClassName` prop, right-aligned last column
- useActiveOrgState: stale org cleanup, localStorage sync
- useAllauthLogin: send email instead of username
- TeamPermissions enum values updated to PascalCase (Add/Change/Delete/View)
- Removed PageLayout wrapper from Teams page

### Expo App
- teamId/teamIds migration for dynamic teams
- FilterTeams component with search
- UserTeamPreference: team selector with persistence
- Notes/Tasks screens updated for `teams.results` offset pagination
- TaskForm team selector, NoteForm team picker
- Generated GraphQL types updated throughout

### Shelter Operator
- Role selection required with validation error rendering (AddUserForm)

PR #2174, opened 2026-06-25
docs: add AI instructions and styleguides for Python/React

vecchp:
## Summary

Add a unified AI instruction system with clear separation of concerns:

- **`docs/ai-instructions.md`** — canonical entry point for all AI tools (Cline, Copilot)
- **`docs/styleguides/react.md`** — React/Frontend conventions (components, state, errors, code style, GraphQL, imports)
- **`docs/styleguides/python.md`** — Python/Django conventions (services, selectors, testing, type checking)
- **`docs/tooling.md`** — common commands and GitHub MCP bootstrap
- **`.clinerules`** — slimmed to pointer (was ~50 lines of duplication)
- **`.github/copilot-instructions.md`** — new, same pointer pattern

### Key features
- Single source of truth — no duplication between Cline and Copilot configs
- Iterative learning rule — AI proactively suggests styleguide additions during sessions
- Clear authority: Python guide points to full HackSoftware Django-Styleguide as canonical source
- Rules extracted from real PR review patterns (SDB-178)

### Related
- PR #2167 (SDB-178) — styleguide rules informed by review comments on that PR

## Summary by Sourcery

Introduce a centralized AI instruction entrypoint and language-specific styleguides for the monorepo.

New Features:
- Add a canonical AI instructions document that all AI assistants reference for project conventions.
- Add dedicated React/frontend and Python/Django styleguides to document component, architecture, and coding conventions.
- Document common tooling commands and GitHub MCP auto-bootstrap behavior for the development environment.
- Add Copilot-specific instructions that delegate to the shared AI instructions document.

Enhancements:
- Simplify existing Cline configuration to point to the shared AI instructions instead of duplicating rules.

Documentation:
- Create AI instructions documentation as the single source of truth for AI coding assistants across the monorepo.
- Add React/frontend styleguide covering components, state management, error handling, GraphQL usage, and import structure.
- Add Python/Django styleguide summarizing architecture, testing, and type-checking practices with a pointer to the canonical external guide.
- Document key Nx-based commands and tooling notes, including GitHub MCP auto-bootstrap fallback steps.


docs/styleguides/python.md
```diff
@@ -0,0 +1,24 @@
+# Python / Django Styleguide
+
+**Primary reference:** The backend follows the [HackSoftware Django-Styleguide](https://github.com/HackSoftware/Django-Styleguide).
```
tglaz: I wonder how Cline will use this, in the sense that we provide a URL and curious when it'd actually look it up and how often. Something to confirm I guess.

Think the URL does belong here, though.
But for Ai agents, technically they should just be aware of it and not have to "read" or "fetch" anything with each prompt.

docs/styleguides/python.md
```diff
@@ -0,0 +1,24 @@
...
+**When working on backend code, read the full styleguide first.** The summary below is a quick reference —
+the canonical rules are in the HackSoftware guide.
+
+## Architecture
+
+- **Services** (`services.py`) — all business logic lives here (not in models or views)
+- **Selectors** (`selectors.py`) — read-only queries
+- **APIs / serializers** — thin views; delegate logic to services
```
tglaz: think we use "resolvers"?

- **API layer (GraphQL resolvers / REST views & serializers)** — thin; delegate business logic to services.
- **API layer (GraphQL resolvers / REST serializers)** — thin; delegate business logic to services.

docs/styleguides/react.md
```diff
@@ -0,0 +1,46 @@
+# React / Frontend Styleguide
+
+## Component Conventions
+
+- **No `export default`** — always use named exports. The only exceptions are routing-related files that require it.
+  - Component logic goes in named files (e.g., `TeamsPage.tsx`).
+  - Index files (`index.ts`) should only re-export: `export { TeamsPage } from './TeamsPage'`.
+  - Never put component logic directly in an index file.
+- **Every lib component must accept `className`** — with few exceptions.
```
tglaz: perhaps can loosen it a bit to
- **Every lib component `should` accept `className` prop** — with few exceptions, such as very local/non-shared components.

docs/styleguides/react.md
```diff
@@ -0,0 +1,46 @@
...
+- **Use `mergeCss` not `clsx`** — `mergeCss` from `@monorepo/react/shared` combines `clsx` + `twMerge`, so overridden Tailwind classes are properly removed. Pass an array of class values: `mergeCss(["text-sm", className])`.
```
tglaz: can probably drop the explanation.
- it's a weird balance today between info for devs and ai
  - ai: want short prompts
  - devs:  they still get the instruction - but can also assume they probably never read this anway :) 

- **Use `mergeCss` not `clsx`** — `mergeCss` from `@monorepo/react/shared` combines `clsx` + `twMerge`.

docs/styleguides/react.md
```diff
@@ -0,0 +1,46 @@
...
+- **No `className = ''` defaults** — `mergeCss` handles falsy values; defaulting to empty string is unnecessary.
```
tglaz: perhaps too granular and can skip (but also if someone does it it breaks nothing)
- the `className= ''` is largely a legacy copy/paste thing that should go away

docs/styleguides/react.md
```diff
@@ -0,0 +1,46 @@
...
+- **Don't stringify class conditions** — use `condition && 'class'` pattern inside `mergeCss()`, not template literals with empty strings.
```
tglaz: same comment as above - perhaps a bit granular
- doesn't hurt to keep, but not really a convention - just how the code works

docs/styleguides/react.md
```diff
@@ -0,0 +1,46 @@
...
+
+## State & Data
+
+- **Prefer custom hooks** for data fetching (e.g., `useOrgTeams`) over raw `useQuery` calls in components.
+  - Hooks should handle loading, error, and empty states internally.
```
tglaz: hooks should just return the error - the consumer should decide what to do with it
- the exception would be when you have a local hook that's not exported (typically there to abstract the functionality from the UI)

docs/styleguides/react.md
```diff
@@ -0,0 +1,46 @@
...
+  - **Co-locate the `.graphql` file with the hook** — the query document and its `__generated__/` output should live in the same directory as the hook (e.g., `hooks/useOrgTeams/teams.graphql`). Consumers import the hook, never the GQL document directly.
+  - Delete the old `.graphql` and `__generated__/` from the previous location after co-locating.
```
tglaz: so perhaps just nitpicking here - but as we add more to "conventions", which we will, we may want to avoid things that are just standard - not sure we need a convention to remove unused files

- if the goal is to remind AI to do things that it tends to forget, then that's different. Though this will still yield mixed results with AI - and you get diminishing results with more prompt instructions

docs/styleguides/react.md
```diff
@@ -0,0 +1,46 @@
...
+
+## Error Handling
+
+- **Never trust backend error messages** to display to users. Log the actual error, show a generic user-friendly message unless the error matches a known, expected key.
+  - `console.error(err)` then show a generic message like `"Sorry, something went wrong. Please try again."`
+- Use `toError()` from `@monorepo/react/shared` for error normalization (in non-user-facing contexts).
+
+## Code Style
+
+- **Avoid nested ternaries in JSX** — use flat `condition && <Component />` patterns instead of `a ? (b ? c : d) : e`.
```
tglaz: this is what I use - results vary :)

```
- Avoid ternary operators unless very basic. Use early returns instead.
  - "Very basic" means both branches are simple literals or single short identifiers — readable in one glance.
  - As soon as either branch contains a function call, template literal with expressions, or any non-trivial computation, use `if/else`.
  - In TSX, use guards for conditional rendering or separate components if gets complicated.
  - Example of an acceptable ternary: `const color = isError ? 'red' : 'gray';`
```

again, unfortunately no great tslint solution for this, which is ok. To me the guideline is basic `legibility` and not any strict rule

docs/styleguides/react.md
```diff
@@ -0,0 +1,46 @@
...
+- **Avoid type casting** (`as`) when possible — make function signatures accept the correct types.
+  - If a callback receives `string | null`, the handler should accept `string | null` — don't cast at the call site.
+  - Use `??` (nullish coalescing) instead of `||` for defaults where `0` or `''` are valid values.
+- **Use `Partial<Record<>>`** for label/display mappings rather than full records with optional keys.
```
tglaz, line 37: this may just be very specific 
- probably would work better with an example
- but ok either way

docs/ai-instructions.md
```diff
@@ -0,0 +1,54 @@
...
+- See `docs/tooling.md` for the full command reference.
+
+## Before Committing Code
+
+**Always run these three checks before committing any changes:**
+
+```bash
+ynx-generate     # regenerate GraphQL schema + types
+ynx-lint         # lint all affected projects
+ynx-typecheck    # typecheck all affected projects
```
tglaz, line 50: does ynx-precommit do all of these?

tglaz (approved):

docs/ai-instructions.md
vecchp: it does but i also like to give it individual commands to run so that it can loop through them as necessary when fixing.  When I've give it uber commands it usually just blindly runs it every time when it wasn't necessary.

PR #2176, opened 2026-06-26
[BP] - create function to export reports to csv

c4bae:
## Summary

Adds server-side CSV export helpers for shelter reporting metrics, including separate CSV generation for daily occupancy, daily bed status, reservation metrics, and average days to occupancy. Bundles all metric CSVs into a ZIP for download with relevant filenames.

## Testing

- Added exporter tests for CSV row generation, per-metric CSV output, ZIP creation, and full shelter metrics ZIP export.

## Summary by Sourcery

Add CSV and ZIP export utilities for shelter occupancy reporting metrics.

New Features:
- Provide helpers to convert shelter occupancy, bed status, reservation, and average-days-to-occupancy metrics into CSV files.
- Bundle multiple metric CSV outputs into a single ZIP archive for download.

Tests:
- Add unit tests covering generic row-to-CSV conversion, per-metric CSV generation, ZIP file creation, and end-to-end shelter metrics ZIP export.


apps/betterangels-backend/reports/export_to_csv.py
```diff
@@ -0,0 +1,126 @@
+import csv
```
frank731, line 8: file can prob be renamed to csv export or smth like that dont want to balloon this file with all the other exporters when they get added

apps/betterangels-backend/reports/export_to_csv.py
```diff
@@ -0,0 +1,126 @@
...
+
+from shelters.types.reporting import (
+    DailyBedStatusMetricsType,
+    DailyOccupancyMetricsType,
+    ReservationMetricsType,
+    ShelterOccupancyMetricsType,
+)
+
+
+def csv_files_to_zip(files: dict[str, str]) -> bytes:
```
frank731, line 22: note for the future, likely will have very similar logic for xlsx export (or we can fit all tables into one file now technically), may need to reuse this logic which would require renaming probably

apps/betterangels-backend/reports/export_to_csv.py
```diff
@@ -0,0 +1,126 @@
...
+    headers = ["start_date", "end_date", "shelter_id", "avg_days_to_occupancy"]
+
+    rows = [
+        {"start_date": start_date, "end_date": end_date, "shelter_id": shelter_id, "avg_days_to_occupancy": avg_days}
+    ]
+
+    return rows_to_csv(rows, headers)
+
+
+def metrics_to_zip(metrics: ShelterOccupancyMetricsType) -> bytes:
```
frank731, line 122: https://app.notion.com/p/uwblueprintexecs/Export-Report-to-CSV-37210f3fb1dc80aea14fe9b1439a8b44?v=27010f3fb1dc81a981f8000c1d31ec81&source=copy_link

updated the ticket, possibly after you already read it, but we now need to take in a list of metrics to return now that we can select a choice of metrics https://www.figma.com/design/I9q9YZrrRrK596fWPOChWJ/Demo-File?node-id=4405-6552&t=sjVfxCLYZFvLHV32-0, may need to make an enum for each metric type or smth if it doesn't exist already (i don't think it does yet)

apps/betterangels-backend/reports/exporters.py
```diff
@@ -0,0 +1,126 @@
...
+    return rows_to_csv(rows, headers)
+
+
+def metrics_to_zip(metrics: ShelterOccupancyMetricsType) -> bytes:
+    shelter_id = str(metrics.shelter_id)
+    start_date = metrics.start_date
+    end_date = metrics.end_date
+
+    files = {
+        "daily_occupancy_metrics.csv": daily_occupancy_metrics_to_csv(shelter_id, metrics.daily_occupancy),
```
frank731: maybe add date to the file names (can also add shelter name but idk how hard that is to fetch here)

frank731 (commented):
Looks good! The main thing is that we now need to support taking in a selection of metrics to export as well.

apps/betterangels-backend/reports/exporters.py
```diff
@@ -0,0 +1,126 @@
+import csv
+import zipfile
+from datetime import date
+from io import BytesIO, StringIO
+from typing import Any, Optional
```
vecchp: **nit:** `from typing import Any, Optional` — `Optional` appears to be used only in `avg_days_to_occupancy_to_csv`. If you drop `Optional` from the annotation in favor of `float | None` (Python 3.10+), you can remove this import.

```suggestion
-from typing import Any, Optional
+from typing import Any
```

apps/betterangels-backend/reports/export_to_csv.py
```diff
@@ -0,0 +1,126 @@
+import csv
+import zipfile
+from datetime import date
+from io import BytesIO, StringIO
+from typing import Any, Optional
+
+from shelters.types.reporting import (
```
vecchp, line 14: **design:** This file imports exclusively from `shelters.types.reporting`, creating a `reports → shelters` dependency for Strawberry/GraphQL types. If the GraphQL schema evolves independently, it could silently break CSV output.

Consider keeping the same *shape* but decoupling from Strawberry:

```python
# Instead of importing @strawberry.type classes directly,
# define a local contract the exporter owns:
from dataclasses import dataclass
from datetime import date

@dataclass
class DailyOccupancyRow:
    date: date
    occupied_count: int
    total_beds: int
    occupancy_pct: float

def daily_occupancy_metrics_to_csv(
    shelter_id: str,
    metrics: list[DailyOccupancyRow],  # no longer tied to Strawberry
) -> str: ...
```

Or keep as-is if this coupling is intentional — just flagging it.

apps/betterangels-backend/reports/export_to_csv.py
```diff
@@ -0,0 +1,126 @@
...
+
+    writer.writeheader()
+
+    for row in rows:
+        writer.writerow({header: row.get(header, "") for header in headers})
+
+    return output.getvalue()
+
+
+def daily_occupancy_metrics_to_csv(shelter_id: str, metrics: list[DailyOccupancyMetricsType]) -> str:
```
vecchp, line 44: **nit:** `date` objects are implicitly stringified to `YYYY-MM-DD` by `csv.writer`. This works but is not obvious. Making it explicit improves readability:

```suggestion
             "date": metric.date.isoformat(),
```

Apply the same `.isoformat()` call in all `*_to_csv` functions that include date fields.

apps/betterangels-backend/reports/exporters.py
```diff
@@ -0,0 +1,126 @@
...
+    ]
+
+    return rows_to_csv(rows, headers)
+
+
+def avg_days_to_occupancy_to_csv(shelter_id: str, start_date: date, end_date: date, avg_days: Optional[float]) -> str:
+    headers = ["start_date", "end_date", "shelter_id", "avg_days_to_occupancy"]
+
+    rows = [
+        {"start_date": start_date, "end_date": end_date, "shelter_id": shelter_id, "avg_days_to_occupancy": avg_days}
```
vecchp: **bug:** When `avg_days` is `None`, the CSV will contain the literal string `"None"`. Consumers shouldn't have to parse Python's `None` representation.

```suggestion
         {"start_date": start_date, "end_date": end_date, "shelter_id": shelter_id,
-         "avg_days_to_occupancy": avg_days}
+         "avg_days_to_occupancy": avg_days if avg_days is not None else ""}
```

Also, please add a test for the `None` case — the current test only covers `4.5`.

apps/betterangels-backend/reports/tests/test_exporters.py
```diff
@@ -1,9 +1,242 @@
...
+                "start_date",
+                "end_date",
+                "shelter_id",
+                "check_in_overdue",
+                "cancelled",
+                "checked_in",
+                "check_in_overdue_to_checked_in",
+            ],
+            ["2026-06-01", "2026-06-30", "shelter-1", "3", "2", "11", "1"],
+        ]
```
vecchp, line 259: **missing test:** No test for `avg_days_to_occupancy_to_csv` when `avg_days=None`. This is the most likely edge case in production (e.g., a brand-new shelter with no occupancy history).

Suggested test:

```python
def test_avg_days_to_occupancy_to_csv_with_none(self) -> None:
    csv_content = avg_days_to_occupancy_to_csv(
        "shelter-1",
        date(2026, 6, 1),
        date(2026, 6, 30),
        None,
    )

    rows = list(csv.reader(StringIO(csv_content)))

    assert rows == [
        ["start_date", "end_date", "shelter_id", "avg_days_to_occupancy"],
        ["2026-06-01", "2026-06-30", "shelter-1", ""],
    ]
```

vecchp (commented):
## Review: `BP]-report-to-csv`

Thanks for this! The overall design is clean and the tests are solid. A few comments below.

### Summary
- **Design concerns 1 (cross-app dependency), 2 (django-import-export reuse), and 3 (integration layer)** are all fine as-is for this PR — no changes needed on those fronts. The integration layer is a follow-up PR, and the raw `csv.DictWriter` approach is appropriate since the data source is Strawberry types, not Django querysets.

### Items to address before merge

1. **`avg_days_to_occupancy_to_csv` with `None`** — When `avg_days` is `None`, the CSV will contain the literal string `"None"`. Should emit an empty string instead, and a test should cover this case.

2. **Coupling to Strawberry types** — `metrics_to_zip` and all `*_to_csv` functions accept Strawberry `@strawberry.type` objects directly. This means changes to the GraphQL schema could silently break CSV output. Consider one of:
   - Accepting plain dataclasses or `TypedDict` instead
   - Or at minimum, documenting this coupling is intentional

3. **Implicit date serialization** — `rows_to_csv` relies on `csv.writer` to stringify `date` objects, which happens to produce `YYYY-MM-DD`. An explicit `.isoformat()` call in each `*_to_csv` function would make this intent clearer.

### Nitpick

- Consider moving `exporters.py` to `shelters/exporters.py` since it only deals with shelter types — but this is optional, keeping it in `reports/` is also reasonable since it's the reports app's export pipeline.

apps/betterangels-backend/reports/export_to_csv.py
```diff
@@ -0,0 +1,145 @@
...
+def metrics_to_zip(metrics: ShelterOccupancyMetricsType) -> bytes:
+    shelter_id = str(metrics.shelter_id)
+    start_date = metrics.start_date
+    end_date = metrics.end_date
+
+    start_str = start_date.strftime("%Y%m%d")
+    end_str = end_date.strftime("%Y%m%d")
+
+    files = {
+        f"{start_str}_{end_str}_daily_occupancy_metrics.csv": daily_occupancy_metrics_to_csv(
```
frank731, line 131: if you want you could add shelter name this but we would prob do that outside once we make the export helper/manager and can get in shelter name there

frank731 (approved):

PR #2179, opened 2026-06-26
[BP] Add demographics filter and date-range guards to reservation status selector

Anahat07:
- Extends `reservation_status_change_counts` in `shelters/selectors/reports.py`
- Inputs are now `shelter`, `start_date`, `end_date`, and an optional `demographics` filter (keyword-only).
- `start_date` / `end_date` are plain `date` objects, inclusive at both ends.
- Adds a one-year range guard and an `end < start` guard (raises `ValueError`).
- `demographics` narrows counts to reservations whose bed *or* room currently serves any of the requested `DemographicChoices`. `ALL` and empty lists are treated as "no narrowing".
- New `ReservationStatusChangeCounts` TypedDict for typed callers.
- Compatible with the model changes from #2147 (no more `Reservation.shelter` FK — resolved via `bed__shelter` / `room__shelter`).

### DIFF
--- a/apps/betterangels-backend/shelters/selectors/reports.py
+++ b/apps/betterangels-backend/shelters/selectors/reports.py
@@ -3,12 +3,18 @@
 import datetime
 from collections import Counter
 from itertools import takewhile
-from typing import Any, TypedDict
+from typing import TYPE_CHECKING, Any, Iterable, TypedDict
 
 import pghistory
 from django.db.models import Count, OuterRef, Q, Subquery, TextField
 from django.db.models.functions import Cast
-from shelters.enums import BedStatusChoices
+from shelters.enums import BedStatusChoices, DemographicChoices
+
+if TYPE_CHECKING:
+    from shelters.models import Shelter
+
+
+MAX_REPORT_RANGE_DAYS = 366  # one (leap) year, inclusive of both endpoints
 
 
 class DailyBedCounts(TypedDict):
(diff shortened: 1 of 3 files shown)


apps/betterangels-backend/shelters/selectors/reports.py
```diff
@@ -3,12 +3,18 @@
...
 from django.db.models import Count, OuterRef, Q, Subquery, TextField
 from django.db.models.functions import Cast
-from shelters.enums import BedStatusChoices
+from shelters.enums import BedStatusChoices, DemographicChoices
+
+if TYPE_CHECKING:
+    from shelters.models import Shelter
+
+
+MAX_REPORT_RANGE_DAYS = 366  # one (leap) year, inclusive of both endpoints
```
frank731: sorry that the ticket reqs were poorly worded, when I said they should support reports up to a year in the past, I didn't mean that the reporting itself had a cap of a year i just meant that it should be tested that it is able to support a year

apps/betterangels-backend/shelters/selectors/reports.py
```diff
@@ -94,28 +100,129 @@ def _counts(day: datetime.date) -> DailyBedCounts:
     return [_counts(start_date + datetime.timedelta(days=n)) for n in range(n_days)]
 
 
-def reservation_status_change_counts(
-    shelter_id: int,
-    start_date: datetime.datetime,
-    end_date: datetime.datetime,
-) -> dict[str, int]:
+class ReservationStatusChangeCounts(TypedDict):
```
frank731: I think we can directly use `ReservationMetricsType `here, should save redeclaring

apps/betterangels-backend/shelters/selectors/reports.py
```diff
@@ -94,28 +100,129 @@ def _counts(day: datetime.date) -> DailyBedCounts:
...
+        values.append(value)
+    return values or None
+
+
+def reservation_status_change_counts(
+    *,
+    shelter: "Shelter",
+    start_date: datetime.date,
+    end_date: datetime.date,
+    demographics: Iterable[DemographicChoices | str] | None = None,
```
frank731: demographic filtering isn't a priority for now, currently not done in design either, can do demographic filtering later

frank731 (changes requested):
I forgot to update the ticket after we decided that demographic filters aren't required at the moment, super sorry about that. I would keep this branch somewhere until design finalizes what demographic filtering would look like and on a seperate branch have the selector return a `ReservationMetricsType ` and reverify that the selector is working properly and update tests against that

apps/betterangels-backend/shelters/selectors/reports.py
vecchp: **flag:** These TypedDict keys (`STATUS_TO_CHECK_IN_OVERDUE`, `STATUS_TO_CANCELLED`, etc.) don't match `ReservationMetricsType` in `shelters/types/reporting.py` (`check_in_overdue`, `cancelled`, etc.). A mapping layer will be needed in the resolver/service that bridges these. Not blocking for this PR.

apps/betterangels-backend/shelters/selectors/reports.py
```diff
@@ -94,28 +100,129 @@ def _counts(day: datetime.date) -> DailyBedCounts:
...
+    from pghistory's ``pgh_diff`` payload).
+    """
+
+    STATUS_TO_CHECK_IN_OVERDUE: int
+    STATUS_TO_CANCELLED: int
+    STATUS_TO_CHECKED_IN: int
+    STATUS_OVERDUE_TO_CHECKED_IN: int
+
+
+def _normalize_demographics(demographics: Iterable[DemographicChoices | str] | None) -> list[str] | None:
```
vecchp: **nit:** `_normalize_demographics` both normalizes values AND decides whether to filter (returns `None` = "no filter"). Consider `_resolve_demographics_filter`. The `ALL` short-circuit is documented but means `[ALL, SENIORS]` silently ignores the non-ALL values.

apps/betterangels-backend/shelters/selectors/reports.py
```diff
@@ -94,28 +100,129 @@ def _counts(day: datetime.date) -> DailyBedCounts:
...
+        value = demo.value if isinstance(demo, DemographicChoices) else demo
+        if value == DemographicChoices.ALL.value:
+            # ALL means: do not narrow.  Short-circuit so callers can pass it
+            # alongside other choices without accidentally widening the filter.
+            return None
+        values.append(value)
+    return values or None
+
+
+def reservation_status_change_counts(
```
vecchp, line 101: **praise:** Keyword-only arguments, proper type annotations, comprehensive docstring with Args/Raises/Notes. Textbook HackSoft selector. 👏

apps/betterangels-backend/shelters/selectors/reports.py
```diff
@@ -94,28 +100,129 @@ def _counts(day: datetime.date) -> DailyBedCounts:
...
+    Notes:
+        Demographic filtering uses the bed/room's *current* demographics.
+        ``Bed.demographics`` / ``Room.demographics`` are not historically
+        tracked by pghistory (M2M through tables aren't included on the
+        ``BedEvent`` / ``ReservationEvent`` tables), so we cannot
+        reconstruct demographic membership as it existed at event time.
     """
     from shelters.models import Reservation
 
+    if end_date < start_date:
```
vecchp, line 137: **styleguide note:** Per the HackSoft styleguide, input validation typically lives in an API's `InputSerializer`, not in the selector. Since this selector has no API yet (it'll be called from a GraphQL resolver), having guards here protects all callers. When the resolver is built, consider migrating these checks to the interface layer. Not blocking.

apps/betterangels-backend/shelters/selectors/reports.py
```diff
@@ -94,28 +100,129 @@ def _counts(day: datetime.date) -> DailyBedCounts:
...
+        datetime.time.min,
+        tzinfo=datetime.timezone.utc,
+    )
+
+    reservations = Reservation.objects.filter(
+        Q(bed__shelter_id=shelter.pk) | Q(room__shelter_id=shelter.pk),
+    )
+
+    demo_values = _normalize_demographics(demographics)
+    if demo_values is not None:
```
vecchp: **optimization:** The demographics filter uses M2M JOINs + `.distinct()`: `Q(bed__demographics__name__in=...) | Q(room__demographics__name__in=...)`. This creates up to 4 extra JOINs (bed→through→Demographic, room→through→Demographic) and requires `.distinct()` because a bed with 2 matching demographics produces duplicate reservation rows.

Consider `Exists` subqueries instead — no `.distinct()` needed and the DB can use semi-join (stops at first match):

```python
from django.db.models import Exists, OuterRef

if demo_values is not None:
    bed_demo = Bed.demographics.through.objects.filter(
        bed_id=OuterRef("bed_id"),
        demographic__name__in=demo_values,
    )
    room_demo = Room.demographics.through.objects.filter(
        room_id=OuterRef("room_id"),
        demographic__name__in=demo_values,
    )
    reservations = reservations.filter(
        Q(Exists(bed_demo)) | Q(Exists(room_demo))
    )
```

Nice-to-have — current approach is correct. The test coverage is already in place to catch regressions if you refactor.

apps/betterangels-backend/shelters/selectors/reports.py
```diff
@@ -129,3 +236,9 @@ def reservation_status_change_counts(
             filter=Q(pgh_diff__status__0="check_in_overdue", pgh_diff__status__1="checked_in"),
         ),
     )
+    return {
+        "STATUS_TO_CHECK_IN_OVERDUE": aggregated["STATUS_TO_CHECK_IN_OVERDUE"] or 0,
```
vecchp: **praise:** `or 0` on each aggregate key is correct. Django's `Count(..., filter=...)` returns `None` (not `0`) on empty rowsets, and `aggregate()` returns `dict[str, int | None]` despite the annotation. Common footgun — well caught.

apps/betterangels-backend/shelters/tests/test_selectors.py
```diff
@@ -5,54 +5,76 @@
...
     # -- helpers -------------------------------------------------------------
 
-    def _make_reservation(self, statuses: list[ReservationStatusChoices], bed: Bed | None = None) -> Reservation:
-        """Create a reservation on a fresh bed then apply each status in order.
-
-        Each reservation gets its own bed to avoid the
-        unique-active-reservation-per-bed constraint. Each ``save`` that changes
-        ``status`` fires a ``reservation.status_change`` event via the pghistory
-        trigger.
+    def _make_reservation(
```
vecchp, line 42: **nit:** The helper always creates a new bed when `bed=None` (`bed=bed or baker.make(...)`). This is correct for test isolation but means each reservation gets a unique bed by default. The demographic tests that pass an explicit `bed=` work around this. Worth a brief comment in the docstring.

apps/betterangels-backend/shelters/tests/test_selectors.py
```diff
@@ -167,43 +191,196 @@ def test_no_events_returns_zero_counts(self) -> None:
...
+        result = reservation_status_change_counts(
+            shelter=self.shelter,
+            start_date=self.start_date,
+            end_date=self.end_date,
+            demographics=[DemographicChoices.SINGLE_MEN],
+        )
+
+        self.assertEqual(result["STATUS_TO_CHECKED_IN"], 0)
+
+    def test_demographics_filter_matches_via_room_when_bed_absent(self) -> None:
```
vecchp: **praise:** Good edge case — room-only reservations (`bed=None`) exercising the `Q(room__shelter_id=...)` path. This validates the OR fallback in the demographics filter.

apps/betterangels-backend/shelters/tests/test_selectors.py
```diff
@@ -167,43 +191,196 @@ def test_no_events_returns_zero_counts(self) -> None:
...
+        result = reservation_status_change_counts(
+            shelter=self.shelter,
+            start_date=self.start_date,
+            end_date=self.end_date,
+            demographics=[DemographicChoices.SENIORS, DemographicChoices.SINGLE_MEN],
+        )
+
+        self.assertEqual(result["STATUS_TO_CHECKED_IN"], 2)
+
+    def test_demographics_filter_does_not_double_count(self) -> None:
```
vecchp: **praise:** Critical test. Without `.distinct()` in the selector, a bed tagged with both `SENIORS` and `SINGLE_MEN` would match both Q conditions and be counted twice. This test proves the guard works.

vecchp (commented):
## Review: `anahat/reservation-status-selector`

Solid work — well-structured selector with thorough test coverage and good docs.

### ✅ What's good

- **TypedDict return** — `ReservationStatusChangeCounts` replaces bare `dict[str, int]` for type safety
- **Keyword-only args** — `*` prevents accidental positional mixups (both dates are `date`)
- **Date semantics are right** — inclusive end date, widened internally to `end_date + 1 day`. Matches `reports/views.py`
- **Input validation** — no more silent zero-returns. `ValueError` for inverted dates and out-of-range windows
- **`.distinct()` on demographics** — prevents double-counting when both bed and room match
- **`or 0` on aggregate** — handles the Django gotcha where conditional `Count` returns `None` on empty rowsets
- **Comprehensive docstring** — documents all 4 buckets, args, raises, and the known M2M history limitation
- **7 demographic test cases** — matching bed, non-matching, room-only, string values, OR semantics, no double-count, ALL sentinel, empty list
- **Only 2 queries** — reservation ID resolution + event aggregation. Can't do it in 1 without denormalizing pghistory

### 🏗️ HackSoft Styleguide Check

| Rule | Status |
|---|---|
| Selectors in `<app>/selectors.py` | ✅ `shelters/selectors/reports.py` |
| Read-only queries | ✅ Zero writes |
| Keyword-only arguments | ✅ |
| Type-annotated | ✅ Full annotations + TypedDict |
| `<entity>_<action>` naming | ✅ `reservation_status_change_counts` |
| Selector handles filtering | ✅ Demographics filter in selector, not caller |

One styleguide note: input validation (`ValueError` for date range) lives in the selector rather than an API serializer. Per the styleguide, validation belongs at the interface layer. Since this selector has no API yet (awaiting a GraphQL resolver), the guards here protect all callers. When the resolver is built, consider migrating these checks up.

### ⚠️ Items

1. **Field name disconnect with `ReservationMetricsType`** — selector keys are `STATUS_TO_CHECK_IN_OVERDUE`, `STATUS_TO_CANCELLED`, etc. but `shelters/types/reporting.py`'s `ReservationMetricsType` uses `check_in_overdue`, `cancelled`, etc. A mapping layer is needed in the resolver/service. Not blocking — just flagging.

2. **`Exists` optimization for demographics** — the current M2M JOIN + `.distinct()` approach is correct but could be more efficient with `Exists` subqueries (see inline comment). Nice-to-have, not a blocker.

3. **Service layer coming** — per the styleguide, a `shelters/services/reports.py` should compose these selectors when `ShelterOccupancyMetricsType` is wired up (follow-up PR). The service would own field-name mapping, input validation, and orchestrate all reporting selectors into one call. Not needed in this PR — the selector is correctly scoped as a standalone data-fetching function.

frank731 (approved):

PR #2181, opened 2026-06-26
chore: upgrade TypeScript 6.0 + Expo SDK 56 (DEV-2435, DEV-2432)

vecchp:
## Summary

Upgrade TypeScript from ~5.9.2 to ^6.0.0 and Expo from SDK 55 to SDK 56, including all necessary config, type, and code migrations.

### TypeScript 6.0
- **Config**: `moduleResolution: node` → `bundler`, `baseUrl` removed, `esModuleInterop: false` removed
- **Module alignment**: 11 projects `module: commonjs` → `esnext` per [TS bundler guidance](https://www.typescriptlang.org/docs/handbook/modules/guides/choosing-compiler-options.html)
- **CSS imports**: `noUncheckedSideEffectImports: false` (correct for bundler projects)
- **e2e**: added `"types": ["node"]` (TS 6 defaults `types` to `[]`)
- **Type fixes**: replaced private `@apollo/client` import with `FieldPolicy['keyArgs']`, explicit return type on `getQueryPolicyFactory`, CJS `import = require()` → ESM

### Expo SDK 56
- **Packages**: `expo@^56.0.0`, `react-native@0.85.3`, `react@19.2.3`, all SDK packages at 56.x
- **RN 0.85**: `StyleSheet.absoluteFillObject` → `absoluteFill`
- **Colors**: `enum` → `const` object — eliminates `as string` casts with RN 0.85's narrowed `string` types
- **Expo Router 56**: dropped `NativeStackNavigationOptions` return types (Stack.Screen accepts plain objects), replaced `@expo/vector-icons` → `@react-native-vector-icons/ant-design`, eliminated prop-drilled router

## Verification

- ✅ `nx affected -t lint` — 20/20 projects pass
- ✅ `nx affected -t typecheck` — all projects pass
- ✅ `expo install --check` — dependencies up to date
- ✅ No forced casts, no private API imports, no unnecessary files


libs/react/shelter/tsconfig.spec.json
```diff
@@ -4,6 +4,7 @@
     "outDir": "../../../dist/out-tsc",
     "module": "esnext",
     "moduleResolution": "Bundler",
```
tglaz: passes checks and not sure if it's an issue, but technically should be lowercase `bundler`
- think there are 3 files with this

tglaz:
probably no big deal, but `tools/tsconfig.tools.json` may need an update

```
 Inherited from base:  "moduleResolution": "bundler"
 Explicitly set:        "module": "commonjs"        ← conflict!
 Also:                  "target": "es5"             ← outdated
```

tglaz:
the `outDir` in `libs/expo/shared/utils/tsconfig.lib.json` seems to have an inconsistent path depth

has `"outDir": "../../dist/out-tsc"`
should be `"outDir": "../../../../dist/out-tsc"`?

libs/ba-platform/tsconfig.json
```diff
@@ -0,0 +1,19 @@
+{
+  "compilerOptions": {
+    "composite": true,
+    "declaration": true,
+    "jsx": "react-jsx",
```
tglaz, line 5: I'd just double check other `tsconfig.json` files
in our expo libs we seem mix these 3: `react-jsx`, `react` and `react-native`

i'd confirm, but think should use `react-jsx` for all of them

tglaz (approved):

PR #2187, opened 2026-06-26
reservations frontend 2 of 2 (SDB-251)

mikefeldberg:
SDB-251


libs/react/shelter-operator/src/lib/components/occupants/OccupantsView.tsx
```diff
@@ -0,0 +1,79 @@
...
+            bedName: r.bed?.name ?? null,
+            startDate: r.startDate ?? null,
+          };
+        });
+      });
+  }, [data]);
+
+  return (
+    <div>
+      <OccupantTable rows={rows} loading={loading} />
```
tglaz: typically i'm always for smaller files, but a bit weird weird to separate rows from columns.
I'd expect `OccupantTable` to include both, unless rows and columns were separate files

libs/react/shelter-operator/src/lib/components/reservations/api/clientQueries.graphql
```diff
@@ -0,0 +1,16 @@
+query SearchClients($search: String!, $pagination: OffsetPaginationInput) {
```
tglaz: BP introduced this `api` pattern, and i'm not totally against it, but we don't really use it.
We use hooks, so having a `useSearchClient` hook with its own gql docs inside would probably be more consistent

similar for reservation CRUD etc
but not a strict blocker

libs/react/shelter-operator/src/lib/components/reservations/api/reservationFormInput.ts
```diff
@@ -0,0 +1,43 @@
+import type {
+  CreateReservationInput,
+  ReservationClientInput,
+  UpdateReservationInput,
+} from '../../../apollo/graphql/__generated__/types';
+import type { ReservationFormData } from '../reservation-form/formTypes';
+
+const sanitizeString = (value?: string | null): string | undefined => {
```
tglaz: `sanitize` has a different meaning.

perhaps `normalizeString` or `toInputString`?

libs/react/shelter-operator/src/lib/components/reservations/components/ClientSearchInput.tsx
```diff
@@ -0,0 +1,240 @@
...
+import { formatClientDisplayName } from '@monorepo/react/shared';
+import { Star, X } from 'lucide-react';
+import { useEffect, useMemo, useRef, useState } from 'react';
+import {
+  SearchClientsDocument,
+  type SearchClientsQuery,
+  type SearchClientsQueryVariables,
+} from '../api/__generated__/clientQueries.generated';
+
+const DEBOUNCE_MS = 300;
```
tglaz: 300 is long. i'd trim it to 100 or 150
but not a huge issue

libs/react/shelter-operator/src/lib/components/reservations/reservation-form/sections/ReservationFormSection.tsx
```diff
@@ -0,0 +1,137 @@
...
+                field.onChange(option ? option.value : null);
+              }}
+            />
+          )}
+        />
+        <p className="text-sm text-gray-600">
+          Optional. At least one of Room or Bed must be selected.
+        </p>
+
+        {bedRoomError ? (
```
tglaz: if do `{bedRoomError &&....` then no need for ternary
assuming bedRoomError is a value || bool || undefined

libs/react/shelter-operator/src/lib/components/reservations/reservation-form/sections/ReservationFormSection.tsx
```diff
@@ -0,0 +1,137 @@
...
+              required
+            />
+          )}
+        />
+
+        <Controller
+          name="notes"
+          control={control}
+          render={({ field }) => (
+            <TextAreaField
```
tglaz: can use base-ui `Input` with type=paragraph (forget how they set up textarea -not great, but works)

libs/react/shelter-operator/src/lib/components/NavBar/breadcrumbs.ts
```diff
@@ -0,0 +1,429 @@
+import { useQuery } from '@apollo/client/react';
```
tglaz, line 1: seems to work, so good as is

but curious if eventually we'd replace it with a simple header that has either shelter name or another title if applicable.
just cause it's a lot of code to maintain, and technically between shelter name in header, selected sidebar or tab link and page/form title the use should know where they are - but more of a product question

libs/react/shelter-operator/src/lib/components/occupants/OccupantTable.tsx
```diff
@@ -0,0 +1,137 @@
...
+}: OccupantTableProps) {
+  const columns: TableColumn<OccupantRow>[] = useMemo(
+    () => [
+      {
+        key: 'client',
+        label: 'Client',
+        width: '1.5fr',
+        cellClassName: 'min-w-0',
+        sortValue: (row) => row.clientName,
+        filterValue: (row) => row.clientName,
```
tglaz, line 59: probably a separate ticket, but seems like `useReservations` uses a paginated query, so any sort/filter may need to be handled server-side

FYI: i created this related ticket earlier: [SDB-254: SO User Management table Role filters should query via api](https://betterangels.atlassian.net/browse/SDB-254)

tglaz, line 59: if wanna do a quick fix, the table filter popover seems to need background color set 

Image

libs/react/shelter-operator/src/lib/components/occupants/OccupantTable.tsx
```diff
@@ -0,0 +1,137 @@
...
+  rowStyle,
+  trailingColumnWidth,
+}: OccupantTableProps) {
+  const columns: TableColumn<OccupantRow>[] = useMemo(
+    () => [
+      {
+        key: 'client',
+        label: 'Client',
+        width: '1.5fr',
+        cellClassName: 'min-w-0',
```
tglaz: cell in Table should already have this class i think?

libs/react/shelter-operator/src/lib/components/occupants/OccupantTable.tsx
```diff
@@ -0,0 +1,137 @@
...
+        sortValue: (row) => row.clientName,
+        filterValue: (row) => row.clientName,
+        render: (row) => (
+          <span className="truncate text-sm font-medium text-gray-900">
```
tglaz, line 61: not sure you need this `span`
adding classes to `cellClassName` will do it

and doesn't seem to truncate anyway

Image

libs/react/shelter-operator/src/lib/components/occupants/OccupantTable.tsx
```diff
@@ -0,0 +1,137 @@
...
+        label: 'Room / Bed',
+        width: '1.5fr',
+        cellClassName: 'text-sm text-gray-700',
+        sortValue: (row) => (row.roomName ?? '') + (row.bedName ?? ''),
+        filterValue: (row) => (row.roomName ?? '') + (row.bedName ?? ''),
+        render: (row) => {
+          const { roomName, bedName } = row;
+          if (roomName && bedName) {
+            return (
+              <span>
```
tglaz: not blocking, but if just returning a string may be better to one of these:
```<>{roomName} &middot; {bedName}</>``` 
```return `${roomName} &middot; ${bedName}`;```

the reason is that when wrapping with a `span` the way css gets applied may change - not an issue here, but in general.

File: libs/react/shelter-operator/src/lib/components/occupants/OccupantTable.tsx (hunk 8 above)
Comment (mikefeldberg) [reply libs/react/shelter-operator/src/lib/components/occupants/OccupantTable.tsx:59]: agreed that sort & filter should be done server-side. This was a quick copy paste of Paul's [client-side add-in](https://github.com/BetterAngelsLA/monorepo/issues/2154). I assumed this was a quick add for presentation purposes and would be replaced with "real" sort & filter in a follow-up

File: libs/react/shelter-operator/src/lib/components/reservations/reservation-form/ReservationForm.tsx [hunk 12]
```diff
@@ -0,0 +1,309 @@
...
+
+  const { data: bedsData } = useQuery<GetBedsQuery, GetBedsQueryVariables>(
+    GetBedsDocument,
+    {
+      variables: { shelterId },
+      skip: !shelterId,
+    }
+  );
+
+  const allBeds = bedsData?.beds.results ?? [];
```
tglaz: probably wanna wrap in `useMemo` since it's a dependency in other useMemos

libs/react/shelter-operator/src/lib/components/reservations/reservation-form/ReservationForm.tsx
```diff
@@ -0,0 +1,309 @@
...
+            readOnlyFields={effectiveReadOnlyFields}
+            selectedClients={selectedClients}
+            primaryClientId={watchedPrimaryClientId}
+            onAddClient={handleAddClient}
+            onRemoveClient={handleRemoveClient}
+            onSetPrimary={handleSetPrimary}
+          />
+
+          <Form.Actions
+            onPrimaryClick={() => handleSubmit(submitReservation)()}
```
tglaz: think can just `onPrimaryClick={handleSubmit(submitReservation)}`

libs/react/shelter-operator/src/lib/components/reservations/reservation-form/ReservationForm.tsx
```diff
@@ -0,0 +1,309 @@
...
+
+        <form
+          onSubmit={handleSubmit(submitReservation)}
+          className="space-y-6"
+          data-testid="reservation-form"
+        >
+          <ReservationFormSection
+            control={control}
+            errors={errors}
+            bedOptions={bedOptions.map(({ value, label }) => ({
```
tglaz, line 278: it just remaps, so this should work: `bedOptions={bedOptions}`

libs/react/shelter-operator/src/lib/components/occupants/OccupantTable.tsx
mikefeldberg: changing filter to open down rather than up fixed this. The issue wasn't background but rendering behind the tabs

libs/react/shelter-operator/src/lib/components/reservations/reservation-form/ReservationForm.tsx
```diff
@@ -0,0 +1,309 @@
...
+  onCancel,
+}: ReservationFormProps) {
+  const isEditMode = Boolean(reservationId);
+  const initialBedIdRef = useRef(initialData?.bedId ?? null);
+
+  const defaults = createEmptyReservationFormData();
+
+  const methods = useForm<ReservationFormData>({
+    resolver: zodResolver(formSchema),
+    defaultValues: initialData ?? defaults,
```
tglaz, line 55: probably should test in Dev, but `initialData` is fetched, so in edit more the first mount may not have all the data, in which case may need to watch for changes and reset form with initialData

libs/react/shelter-operator/src/lib/components/reservations/reservation-form/ReservationForm.tsx
```diff
@@ -0,0 +1,309 @@
...
+    control,
+  });
+
+  const [selectedClients, setSelectedClients] = useState<SelectedClient[]>(
+    () => initialSelectedClients ?? []
+  );
+
+  const handleAddClient = useCallback(
+    (client: SelectedClient) => {
+      const newIds = watchedClientIds.includes(client.id)
```
tglaz, line 80: can improve by using `getValues` in these useCallbacks as this is only triggered by user (doesn't need to react), so don't risk new renders when watched dependency changes.
allows you to remove the `watchedClientIds`

```
 const { getValues } = methods; 

const handleAddClient = useCallback(
  (client: SelectedClient) => {
    const { clientIds, primaryClientId } = getValues();

```

libs/react/shelter-operator/src/lib/components/reservations/ReservationsView.tsx
```diff
@@ -0,0 +1,191 @@
...
+
+  const closeCheckinConfirmation = useCallback(() => {
+    setCheckinConfirmation({ isOpen: false, reservationId: null });
+  }, []);
+
+  const handleStatusUpdate = useCallback(
+    async (
+      reservationId: string,
+      status: ReservationStatusChoices,
+      action: 'checkin' | 'complete' | 'cancel',
```
tglaz: since using `'checkin' | 'complete' | 'cancel'` twice - may just define the type

libs/react/shelter-operator/src/lib/pages/dashboard/components/create-shelter-form/CreateShelterForm.tsx
```diff
@@ -2,7 +2,7 @@ import { useMutation } from '@apollo/client/react';
 import { Button } from '@monorepo/react/components';
```
tglaz, line 2: this can just be deleted at this point. think just need to update a link when there are no shelters to point to new form
i've a cleanup ticket for that

libs/react/shelter-operator/src/lib/pages/reservations/EditReservationPage.tsx
```diff
@@ -0,0 +1,82 @@
...
+  const navigate = useNavigate();
+  const location = useLocation();
+  const { shelterId, reservationId } = useParams();
+  const { reservation, loading, error } = useReservation(
+    reservationId ? reservationId : ''
+  );
+
+  const reservationsPath = shelterManageReservationsRoute(shelterId ?? '');
+
+  const state = (location.state ?? {}) as ReservationLocationState;
```
tglaz: should work, but a bit unsafe

libs/react/shelter-operator/src/lib/pages/reservations/EditReservationPage.tsx
```diff
@@ -0,0 +1,82 @@
...
+import { ManageFormPageLayout } from '../../components/manage-form-page-layout';
+import { useReservation } from '../../hooks/useReservation';
+import { shelterManageReservationsRoute } from '../../routing';
+
+interface ReservationLocationState {
+  bedId?: string | null;
+  roomId?: string | null;
+}
+
+export function EditReservationPage() {
```
tglaz: this does create + edit, so maybe should just be `ReservationFormPage`
or the page would call ``ReservationForm`

libs/react/shelter-operator/src/lib/pages/reservations/EditReservationPage.tsx
```diff
@@ -0,0 +1,82 @@
...
+    reservationId ? reservationId : ''
+  );
+
+  const reservationsPath = shelterManageReservationsRoute(shelterId ?? '');
+
+  const state = (location.state ?? {}) as ReservationLocationState;
+
+  const defaults = createEmptyReservationFormData();
+  let initialData: ReservationFormData | undefined;
+  const readOnlyFields: ('bedId' | 'roomId')[] = [];
```
tglaz: the `initialData` + `readOnlyFields` could better be defined using `useMemo`
since this does edit, initial data may change - or come after first mount, so watching for it and memoizing may safer and cleaner

tglaz (changes requested):

libs/react/shelter-operator/src/lib/pages/reservations/EditReservationPage.tsx
tglaz: probably for later, or i'm just misreading
but how does that data gets there. where is bedId and roomId get set on state?

tglaz (approved):

PR #2190, opened 2026-06-27
feat(shelter-operator): date-range filter foundation — logic + state (PR1)

julianchen24:
**📚 Stack — date-range filter** (merge order, top → bottom):
1. **#2190 — foundation 👈 (1/6)** · 2. #2191 — Calendar · 3. #2192 — YearGrid · 4. #2193 — presets · 5. #2194 — popover · 6. #2195 — toolbar

---

## Problem

The operator Overview/reporting dashboard can't be scoped to a date range. The filter that fixes this is built from several UI pieces (a presets dropdown and a custom calendar) that must agree on one range and one default — without a shared contract they drift.

## Solution

Establishes the shared vocabulary the rest of the feature builds on: the named presets (Last 7 / 30 Days, Month-To-Date, Year-To-Date, Custom), the rule that turns each preset into a concrete range, and one piece of shared state that defaults to **Last 30 Days**. Every later PR reads and writes this single source of truth, so the dropdown and calendar can never disagree. No UI ships here.

### DIFF
--- a/libs/react/shelter-operator/src/lib/components/date-range-filter/presets.ts
+++ b/libs/react/shelter-operator/src/lib/components/date-range-filter/presets.ts
@@ -1,42 +1,16 @@
 import { startOfMonth, startOfYear, subDays } from 'date-fns';
 import type { DateRange, DateRangePreset } from './types';
 
-/** Human-readable labels for each preset (drives the dropdown options). */
 export const PRESET_LABELS: Record<DateRangePreset, string> = {
   LAST_7_DAYS: 'Last 7 Days',
   LAST_30_DAYS: 'Last 30 Days',
-  // Figma's label read "Month-To-Year"; treated as a typo. "Month-To-Date"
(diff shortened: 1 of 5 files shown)


libs/react/shelter-operator/src/lib/components/date-range-filter/presets.ts
```diff
@@ -0,0 +1,30 @@
...
+
+export const DEFAULT_PRESET: DateRangePreset = 'LAST_30_DAYS';
+
+export function resolvePreset(
+  preset: DateRangePreset,
+  today: Date = new Date()
+): DateRange {
+  switch (preset) {
+    case 'LAST_7_DAYS':
+      return { from: subDays(today, 6), to: today };
```
frank731, line 20: Should this floor to the nearest date? Date range type doesn't care about time so I doubt this is an issue but maybe keep in mind

frank731 (approved):

PR #2191, opened 2026-06-27
feat(shelter-operator): single-month range Calendar grid (PR2)

julianchen24:
**📚 Stack — date-range filter**: #2190 → **#2191 (2/6)** → #2192 → #2193 → #2194 → #2195

---

## Problem

The custom picker needs a calendar surface that matches the design and app tokens — range band, endpoint circles, neutral day hover, month navigation, and a clickable month/year caption.

## Solution

A presentational, fully-controlled single-month range calendar (the popover later shows two side by side); the parent drives selection, and clicking within a range adjusts the nearest endpoint. Day hover uses the neutral design-system grey.

## How to test

Storybook → `Date Range Filter/Calendar`. Pick a range; page the month; click inside a selected range and confirm it moves an endpoint.


package.json
julianchen24, line 1: careful and yarn.lock of this when merging

apps/betterangels-backend/shelters/selectors/reports.py
```diff
@@ -99,83 +95,43 @@ def _counts(day: datetime.date) -> DailyBedCounts:
```
frank731, line 410: rn we just have them all as datetime (probably should be date though now that i look at it) can either remove this diff or do a full refactor of all selectors and the resolver [here](https://github.com/BetterAngelsLA/monorepo/commit/a46455b425af77892edf22a44ae7cb078971b16f)

frank731 (dismissed):

libs/react/shelter-operator/src/lib/components/date-range-filter/Calendar.tsx
```diff
@@ -0,0 +1,112 @@
...
+  onSelect?: (range: RdpDateRange | undefined) => void;
+  month?: Date;
+  onMonthChange?: (month: Date) => void;
+  onMonthLabelClick?: (month: Date) => void;
+  className?: string;
+}
+
+const dayCell = 'relative h-9 w-9 p-0 text-center';
+const dayButton =
+  'h-9 w-9 rounded-full text-sm text-[#383B40] transition-colors hover:bg-[#F4F6FD]';
```
StockerMC: Several of these hex values repeat elsewhere in the folder (#008CEE in all three of Calendar/YearGrid/DateRangeCalendar, plus #F4F6FD, #747A82, #0071C0 across two of them). I think it'd be cleaner to lift the shared ones into one constants file so a palette tweak is a single edit.

libs/react/shelter-operator/src/lib/components/date-range-filter/Calendar.tsx
```diff
@@ -0,0 +1,127 @@
...
+  onMonthChange,
+  onMonthLabelClick,
+  className,
+}: CalendarProps) {
+  return (
+    <CalendarContext.Provider value={{ onMonthChange, onMonthLabelClick }}>
+      <DayPicker
+        mode="range"
+        selected={selected}
+        onSelect={onSelect}
```
Anahat07: The PR description says clicking within an existing range should adjust the nearest endpoint, but it looks like we’re relying directly on react-day-picker’s range selection here. In v9.14.0, for a completed June 10-18 range, clicking June 11 keeps June 10 as from and changes to to June 11 rather than moving the nearer start endpoint to June 11. Could we handle the nearest-endpoint behaviour explicitly before calling onSelect? It would also be helpful to add a test covering clicks near both endpoints.

Anahat07 (dismissed):
The calendar structure and colour cleanup look good overall. I found one behavioural issue with selecting a date inside an existing range: the current behaviour doesn’t appear to adjust the nearest endpoint as described in the PR. I left an inline comment with a reproducible example. Once that’s addressed, this looks good from my side.

libs/react/shelter-operator/src/lib/components/date-range-filter/Calendar.tsx
StockerMC: Good catch, and the repro is exact. Confirmed in react-day-picker 9.14.0's `addToRange`: for a completed range the `isAfter(date, from)` branch fires for any click after the start, so it always moves `to` and never the nearer endpoint.

Fixed by adjusting the closer endpoint before calling `onSelect`, rather than passing react-day-picker's result straight through. Clicks outside the range still extend the side they fall on.

Also added `Calendar.test.tsx` covering clicks near both endpoints, plus before/after the range and the empty-selection case. Reverting the fix fails all five.

I've taken this stack over from Julian, so further changes will come from me.

package.json
StockerMC: Checked this. react-day-picker and everything it pulls in are still in the lockfile, and the 62 lines that commit dropped were unrelated leftovers (`@csstools/*`, `dateformat`) that main doesn't have either.

apps/betterangels-backend/shelters/selectors/reports.py
StockerMC: This file isn't in the PR anymore, the diff is frontend only now. Mind re-reviewing so it's not blocking?

libs/react/shelter-operator/src/lib/components/date-range-filter/Calendar.tsx
StockerMC: Fixed. Heads up that the first attempt used `text-[${COLOR}]`, which silently broke the colors: Tailwind scans source as plain text, so those classes never got generated. It exports full class strings now.

vecchp:
Adding `react-day-picker` pulls a second major version of `date-fns` into the workspace. The
lockfile on this branch has both:

```
"date-fns@npm:^3.6.0"   -> 3.6.0     (the repo's own pin)
"date-fns@npm:^4.1.0"   -> 4.4.0     (via react-day-picker 9.14.0)
```

`main` has only 3.6.0 today.

Two things follow from that, and I don't think either is a blocker — but they're worth a
decision rather than arriving by accident:

1. **Both copies ship.** They're separate module instances, so bundling `react-day-picker`
   means bundling date-fns twice.
2. **`Date` objects cross between them.** `Calendar.tsx` and `dateRangeCalendarUtils.ts` import
   from the v3 copy, and `react-day-picker` hands you `Date`s produced by v4 internals. That's
   fine as long as it stays plain `Date`s — v4's headline change is the `TZDate`/`in` option for
   timezone-aware arithmetic, and a `TZDate` flowing into a v3 helper would not behave the way
   the call site reads. Nothing here does that, so this is a "watch for it" rather than a bug.

Cheapest resolution is probably to bump the root pin to `^4.1.0` so there's one copy. It's a
major, so `format`/`parse` call sites across the repo want a look first — v4 kept the API
surface but changed some defaults.

Alternatively `react-day-picker` exposes its own date library adapter, so there may be a way to
hand it the v3 instance instead. I haven't checked whether that path is workable on 9.x.

Either way, worth a note in the PR body about which choice you made, since the duplicate is
invisible from the diff — it only shows in `yarn.lock`.

vecchp (approved):

PR #2192, opened 2026-06-27
feat(shelter-operator): YearGrid year picker (PR3)

julianchen24:
**📚 Stack — date-range filter**: #2190 → #2191 → **#2192 (3/6)** → #2193 → #2194 → #2195

---

## Problem

Reaching a far-off month with next/prev arrows is slow; the design offers a year grid to jump straight to a year.

## Solution

A year picker the calendar caption opens — pick a year and the view jumps there, keeping the month. The year window is a fixed 2021–2036 to match the design, with a marked follow-up to make it dynamic once design confirms the range.

## How to test

Storybook → `Date Range Filter/YearGrid`.

## Summary by Sourcery

Add a reusable year picker for quickly selecting a year in the date-range filter.

New Features:
- Add an accessible year-selection grid with selectable years anchored around the current year.
- Expose the YearGrid component and provide Storybook examples for selected and unselected states.

Enhancements:
- Centralize shared accent styling for selected and current calendar controls.

Tests:
- Add coverage for year ranges, selection state, current-year marking, default date behavior, and selection callbacks.

### DIFF
--- a/libs/react/shelter-operator/src/lib/components/date-range-filter/YearGrid.tsx
+++ b/libs/react/shelter-operator/src/lib/components/date-range-filter/YearGrid.tsx
@@ -1,0 +1,65 @@
+import { mergeCss } from '@monorepo/react/shared';
+
+// TODO(date-range-filter): these years are hard-coded to match Figma
+// (a fixed 4x4 grid). Replace with a dynamic/relative window — e.g. derived
+// from today or the data's earliest year — once design confirms the range.
+// See docs/plans/date-range-filter-plan.md (Phase 0 decision #2).
+const FIRST_YEAR = 2021;
+const LAST_YEAR = 2036;
+
+const YEARS = Array.from(
+  { length: LAST_YEAR - FIRST_YEAR + 1 },
+  (_, i) => FIRST_YEAR + i
+);
+
+export interface YearGridProps {
+  /** The year currently selected (filled). */
+  selectedYear?: number;
+  /** The year to outline as "current" (defaults to today's year). */
+  currentYear?: number;
+  /** Fired when a year cell is chosen. */
+  onSelect: (year: number) => void;
+  className?: string;
+}
+
+/**
+ * 4x4 grid of selectable years. Presentational leaf — the selected year is
+ * filled (`primary-main`) and the current year is outlined.
+ */
+export function YearGrid({
+  selectedYear,
+  currentYear = new Date().getFullYear(),
+  onSelect,
+  className,
+}: YearGridProps) {
+  return (
+    <div
+      role="grid"
+      className={mergeCss(['grid grid-cols-4 gap-2 font-sans', className])}
+    >
+      {YEARS.map((year) => {
+        const isSelected = year === selectedYear;
+        const isCurrent = year === currentYear;
+        return (
+          <button
+            key={year}
+            type="button"
+            role="gridcell"
+            aria-selected={isSelected}
+            onClick={() => onSelect(year)}
+            className={mergeCss([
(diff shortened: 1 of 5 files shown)


libs/react/shelter-operator/src/lib/components/date-range-filter/YearGrid.tsx
```diff
@@ -0,0 +1,53 @@
...
+
+export function YearGrid({
+  selectedYear,
+  currentYear = new Date().getFullYear(),
+  onSelect,
+  className,
+}: YearGridProps) {
+  return (
+    <div
+      role="grid"
```
StockerMC: role="grid" with gridcell children but no role="row" in between is an incomplete grid pattern (gridcell needs a row parent). I'd either add the row wrapper or drop the grid roles and leave them as plain buttons.

StockerMC: Fixed. Dropped the grid roles rather than adding a row wrapper, and switched to `aria-pressed` since `aria-selected` isn't valid on a plain button.

libs/react/shelter-operator/src/lib/components/date-range-filter/colors.ts
```diff
@@ -28,6 +28,22 @@ export const HOVER_BG = 'hover:bg-neutral-99';
...
+export const ACCENT_BG = 'bg-[#008CEE]';
+
+/** Darker primary blue – hover on a selected control */
+export const ACCENT_HOVER_BG = 'hover:bg-[#0071C0]';
+
+/** Primary accent blue ring marking the current day or year */
+export const ACCENT_RING = 'ring-[#008CEE]';
+
+/*
+ * The same three scoped to react-day-picker's day cell, which renders its
```
frank731, line 41: is this comment needed

libs/react/shelter-operator/src/lib/components/date-range-filter/YearGrid.tsx
```diff
@@ -0,0 +1,66 @@
...
+export interface YearGridProps {
+  selectedYear?: number;
+  currentYear?: number;
+  onSelect: (year: number) => void;
+  className?: string;
+}
+
+export function YearGrid({
+  selectedYear,
+  currentYear = new Date().getFullYear(),
```
frank731, line 32: is this fine to do inline as a default?

this gets highlighted with 
"Do not call 'new Date()' during render. Components and hooks must be pure. Move this call into an event handler, effect, or state initializer."

frank731 (commented):
the current year fetch is a bit suspicious but otherwise lgtm

libs/react/shelter-operator/src/lib/components/date-range-filter/YearGrid.tsx
StockerMC: Good catch. It's `@eslint-react/purity` that flags it (react-hooks' own purity rule ignores `new Date()`), warn-level, which is why CI stayed green.

Moved it to a first-render `useState` initializer, which is one of the remedies the rule names, and kept the prop optional. Also added a test for the no-prop path, since that's the one DateRangeCalendar actually uses and it had no coverage.

libs/react/shelter-operator/src/lib/components/date-range-filter/colors.ts
StockerMC: Mostly not, no. The file header already says every entry has to be a full class string with variants included, so that part was restating it. Cut it down to the one line it was actually carrying: that the `[&>button]:` prefix exists because react-day-picker renders the day button inside the cell that takes the class.

PR #2193, opened 2026-06-27
feat(shelter-operator): presets dropdown wired to shared state (PR5)

julianchen24:
**📚 Stack — date-range filter**: #2190 → #2191 → #2192 → **#2193 (4/6)** → #2194 → #2195

---

## Problem

Most users want a common window (last 30 days, year-to-date) without hand-picking dates.

## Solution

A presets dropdown that resolves the chosen window to a concrete range and updates the shared filter, so the calendar reflects it too. "Custom" is the exception — it doesn't resolve to a range; it signals the toolbar to open the calendar for hand-picking, and nothing commits until the user saves.

## How to test

Storybook → `Date Range Filter/DateRangePresetDropdown`.

## Summary by Sourcery

Wire date-range presets into shared state while preserving custom calendar selection behavior.

New Features:
- Add a date-range preset dropdown that updates the shared filter for standard presets and delegates custom selection to the toolbar.

Enhancements:
- Expose the preset dropdown through the date-range filter component exports.

Tests:
- Add Storybook coverage and tests for shared-state updates, custom selection behavior, and optional handlers.

### DIFF
--- a/libs/react/shelter-operator/src/lib/components/date-range-filter/DateRangePresetDropdown.tsx
+++ b/libs/react/shelter-operator/src/lib/components/date-range-filter/DateRangePresetDropdown.tsx
@@ -1,0 +1,66 @@
+import { useAtom } from 'jotai';
+import { Dropdown, toDropdownOptions, toDropdownValue } from '../base-ui/dropdown';
+import type { DropdownOption } from '../base-ui/dropdown';
+import { dateRangeFilterAtom } from './dateRangeFilterAtom';
+import { PRESET_LABELS, resolvePreset } from './presets';
+import type { DateRangePreset } from './types';
+
+export interface DateRangePresetDropdownProps {
+  /**
+   * Called when the user picks "Custom". The toolbar opens the calendar in
+   * response; the atom's range is committed later, on Save. The atom preset is
+   * intentionally not changed here.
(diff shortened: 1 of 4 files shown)


libs/react/shelter-operator/src/lib/components/date-range-filter/DateRangePresetDropdown.tsx
```diff
@@ -0,0 +1,47 @@
...
+  className,
+}: DateRangePresetDropdownProps) {
+  const [filter, setFilter] = useAtom(dateRangeFilterAtom);
+
+  const value = toDropdownValue(displayPreset ?? filter.preset, PRESET_LABELS);
+
+  function handleChange(option: DropdownOption<DateRangePreset> | null) {
+    if (!option) return;
+
+    if (option.value === 'CUSTOM') {
```
StockerMC, line 28: The CUSTOM branch is the only real logic here and isn't covered yet. I think one test case would be worth it.

StockerMC: Added `DateRangePresetDropdown.test.tsx`, three cases covering the CUSTOM branch.

PR #2194, opened 2026-06-27
feat(shelter-operator): custom-range calendar popover (PR4)

julianchen24:
**📚 Stack — date-range filter**: #2190 → #2191 → #2192 → #2193 → **#2194 (5/6)** → #2195

---

## Problem

Users need to pick an arbitrary date range both by clicking a calendar and by typing exact dates — without the two drifting out of sync, without bad input corrupting the range, and with the two month panes staying in a sensible order.

## Solution

The custom-range popover: two months with a per-side year picker, plus typable From/To fields. Typing and clicking drive the same draft so they always agree; slashes fill in as you type (`06012026` → `06/01/2026`), and an invalid or out-of-order date is flagged inline and never committed. The two panes stay ordered — navigating the start past the end pushes the end forward, and vice-versa. The range applies on Save; Cancel discards. Styling follows the design: neutral day hover, grey Cancel, tighter calendar gap, and no divider above the actions.

## How to test

Storybook → `Date Range Filter/DateRangeCalendar`:
- Type `05/20/0000` (or `13/40/2026`) → field turns red, nothing commits.
- Type `06012026` → slashes fill in automatically.
- Pick a range on the grid → the From/To fields update to match.
- Page the start month past the end → the end month follows (and vice-versa).

## Summary by Sourcery

Add a synchronized custom date-range calendar for selecting, typing, validating, and committing arbitrary date ranges.

New Features:
- Add a custom date-range calendar with dual month views, per-side year selection, editable From/To fields, and draft-based Save/Cancel behavior.

Bug Fixes:
- Prevent invalid, incomplete, or out-of-order typed dates from being committed and keep the two calendar months in chronological order.

Enhancements:
- Support automatic date-input masking and synchronization between calendar selections and typed date fields.
- Allow menu panels to size to their content when left-aligned.
- Export the date-range calendar component and its props for reuse.

Documentation:
- Add Storybook examples for populated and empty date-range calendar states.

Tests:
- Add utility and interaction coverage for date parsing, input masking, range validation, month coupling, Escape dismissal, and commit behavior.

### DIFF
--- a/libs/react/shelter-operator/src/lib/components/date-range-filter/DateRangeCalendar.tsx
+++ b/libs/react/shelter-operator/src/lib/components/date-range-filter/DateRangeCalendar.tsx
@@ -9,8 +9,10 @@ import { usePortalPosition } from '../base-ui/dropdown/usePortalPosition';
 import { Calendar, type RdpDateRange } from './Calendar';
 import {
   computeFieldCommit,
+  coupleMonths,
   defaultMonths,
   formatDate,
+  maskDateInput,
   toDomain,
   toRdp,
 } from './dateRangeCalendarUtils';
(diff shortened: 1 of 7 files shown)


libs/react/shelter-operator/src/lib/components/date-range-filter/DateRangeCalendar.tsx
```diff
@@ -0,0 +1,305 @@
...
+          onClick={() => (open ? close() : openPopover())}
+          className="ml-auto shrink-0"
+        >
+          <CalendarIcon className="h-6 w-6 text-[#676767]" />
+        </button>
+      </div>
+
+      {open &&
+        createPortal(
+          <MenuPanel
```
StockerMC, line 285: Popover closes on backdrop click, scroll, and resize, but not on Escape, and there's no dialog role or focus handling. I think it'd be good to add an Escape handler and role="dialog".

StockerMC: Added an Escape handler that also restores focus to the toggle, plus `role="dialog"`. Covered in `DateRangeCalendar.test.tsx`.

PR #2195, opened 2026-06-27
feat(shelter-operator): date-range filter toolbar + dashboard mount (PR6)

julianchen24:
image
image
image
image

**📚 Stack — date-range filter**: #2190 → #2191 → #2192 → #2193 → #2194 → **#2195 (6/6)**

---

## Problem

The operator Overview/reporting dashboard has no date filter — users can't scope what they're viewing.

## Solution

Assembles the pieces into the reporting toolbar — a custom-range field beside the presets dropdown (plus a placeholder, disabled Export Data action) — and mounts it on the dashboard. Picking a preset or saving a custom range updates one shared filter; a hand-picked range shows as "Custom." Feeding the selected range into the actual dashboard queries is follow-up work.

## How to test

Run the operator app → open a shelter's Overview dashboard → exercise the toolbar (preset select, custom range via the field). Export Data is intentionally disabled (separate feature).

## Summary by Sourcery

Enable date-range filtering controls in the shelter operator reporting dashboard while preserving the existing export workflow.

New Features:
- Add functional date-range calendar and preset controls to the shelter operator reporting toolbar, with custom selections reflected as a shared filter.
- Mount the reporting filter toolbar on the dashboard with the existing export action.

Enhancements:
- Ensure the Tailwind sans-serif font uses the branded Poppins font stack.

Tests:
- Update report filter bar tests to seed the shared date-range state while exercising the real date controls.


libs/react/shelter-operator/src/lib/pages/dashboard/ShelterDashboardPage.tsx
```diff
@@ -78,8 +79,16 @@ export default function ShelterDashboardPage({ tab }: { tab: ShelterTab }) {
         items={TAB_ITEMS}
       />
 
+      {tab === 'overview' && (
+        <>
+          <div className="px-6 pt-4">
+            <DateRangeFilterBar />
```
StockerMC: The bar writes dateRangeFilterAtom, but OverviewView never reads it, so the filter doesn't affect the data yet. Is that intentional for this PR? A TODO would help if the wiring lands later.

StockerMC: Withdrawing this: it's out of date. Main has since wired `dateRangeFilterAtom` into `ReportsView`, which passes `range.from`/`range.to` into `useShelterOccupancyMetrics`, so the filter does drive the data now.

The mount here is the part that needs changing. Main rewrote the dashboard down to a single `reports` tab, so the `tab === 'overview'` block this sits in no longer exists. The real target is `ReportFilterBar` in `ReportsView`, which is a placeholder for exactly this component.

I'm taking this stack over and will handle it.

vecchp:
Heads up before this gets rebased — #2301 changed the file this PR deletes.

`ReportFilterBar.tsx` is no longer just the two placeholders. #2301 added a working **Export Data** button and its modal to it, and `ReportsView` now renders `<ReportFilterBar shelterId={shelterId} />` rather than `<ReportFilterBar />`.

This PR deletes that file (`0+/61-`) and swaps the render for `<DateRangeFilterBar className="flex-wrap" />`, so a straight rebase drops the export button without a conflict to warn you.

Reconciliation, roughly:

- `DateRangeFilterBar` takes `shelterId?: string` and `ReportsView` passes it through.
- Carry the export button and modal across into `DateRangeFilterBar`, right of the date controls — the layout slot is the same one the placeholders occupied.
- The pieces are already in `components/reports/`: `ExportDataModal.tsx` and `useShelterMetricsExport.ts`. The button needs `shelterId`, the `range` from `dateRangeFilterAtom` (which you are already reading), and `disabled={!shelterId}`.
- `ReportFilterBar.test.tsx` covers the export flow and asserts the request URL, so point it at `DateRangeFilterBar` and it will tell you whether the wiring survived.

Nothing here changes what this PR is doing — replacing the placeholders with the real controls is still the goal. It just needs to bring the export button along.

PR #2198, opened 2026-06-29
display real capacity in shelter dashboard (SDB-258)

mikefeldberg:
SDB-258

## Summary by Sourcery

Update shelter dashboard to base capacity display and sorting on detailed bed status instead of total/available beds.

New Features:
- Show unavailable vs total beds using full beds-by-status data in the shelter table.

Enhancements:
- Introduce a structured beds-by-status type and propagate it from GraphQL through the dashboard into the shelter table.
- Replace ambiguous 'N/A' capacity label with clearer messaging when no bed availability data is present.


libs/react/shelter-operator/src/lib/components/ShelterTable.tsx
```diff
@@ -106,13 +100,16 @@ export function ShelterTable({
         width: '1.2fr',
         cellClassName: 'whitespace-nowrap text-gray-700',
         render: (shelter) => {
-          const totalBeds = shelter.totalBeds ?? 0;
-          const reservedBeds = computeReservedBeds(shelter);
+          const totalBeds = sumAllBeds(shelter.bedsByStatus);
+          const unavailableBeds =
```
tglaz: we compute this twice - here and in `getRowObject` we pass to Table
could add a `getCapacityStats` function that'd return both totalBeds + unavailableBeds

libs/react/shelter-operator/src/lib/components/ShelterTable.tsx
```diff
@@ -29,14 +29,8 @@ type ShelterTableProps = {
...
 
-function computeReservedBeds(shelter: Shelter): number | null {
-  const totalBeds = shelter.totalBeds ?? 0;
-  if (totalBeds === 0) return null;
-  return Math.min(
-    Math.max(totalBeds - (shelter.availableBeds ?? 0), 0),
-    totalBeds
-  );
-}
+const sumAllBeds = (bedsByStatus: Shelter['bedsByStatus']): number =>
```
tglaz: doesn't `Shelter` have `totalBeds` prop?
not sure if it's computed on server, but this should be something we know and/or use?
unless some statuses are excluded from count i suppose

tglaz (changes requested):

libs/react/shelter-operator/src/lib/components/ShelterTable.tsx
mikefeldberg: doh

libs/react/shelter-operator/src/lib/components/ShelterTable.tsx
mikefeldberg: `totalBeds` is the temporary int field we put in to show capacity. But you're right, we should have one, so I changed `bedsByStatus` to `bedCounts` and added `total` to it.

libs/react/shelter-operator/src/lib/components/overview/OverviewView.tsx
```diff
@@ -33,8 +33,9 @@ export function OverviewView({ shelterId }: { shelterId: string }) {
   });
 
   const shelter = data?.operatorShelter;
-  const bedsByStatus = shelter?.bedsByStatus;
-  const roomsByStatus = shelter?.roomsByStatus;
+  console.log('OverviewView shelter:', shelter);
```
tglaz: straggling console log

libs/react/shelter-operator/src/lib/components/ShelterTable.tsx
```diff
@@ -100,13 +101,14 @@ export function ShelterTable({
         width: '1.2fr',
         cellClassName: 'whitespace-nowrap text-gray-700',
         render: (shelter) => {
-          const totalBeds = sumAllBeds(shelter.bedsByStatus);
-          const unavailableBeds =
-            totalBeds - (shelter.bedsByStatus.available ?? 0);
+          console.log('Rendering capacity for shelter:', shelter);
```
tglaz: console log

tglaz (approved):

PR #2200, opened 2026-06-30
Shelter create FE validation

tglaz:
- default to empty string on Shelter Create
- update Input to accept dataType=url
- run rhf validation onBlur consistently - keep submit btn clickable
- add url and domain regex to align closely with with Django URLValidator
- add regex tests and enable running them
  - fix sanitizeHtml tests
- tsconfig.json updates

## Summary by Sourcery

Tighten front-end validation and submission behavior across shelter profile forms and wire up shared React library testing configuration.

New Features:
- Add basic website URL regex to shared static regex utilities for reuse in form validation.
- Introduce Jest configuration and Nx test target for the react-shared library.

Bug Fixes:
- Prevent saving invalid shelter photos by disabling the Save action until the photo form is valid.
- Adjust create shelter payload to always send a non-undefined description string.

Enhancements:
- Strengthen email and website validation messages in the shelter basic info form for clearer user feedback.
- Standardize react-hook-form validation modes to onBlur for most shelter profile segment forms and keep schedules onSubmit with documentation explaining the choice.
- Rename the shared React library project to react-shared and align its tsconfig setup and references with the workspace base configuration.
- Rename the empty-string regex to nonBlank to better reflect its semantics.

Tests:
- Add unit tests around shared regex utilities to validate email, phone, and website patterns.


vecchp (approved):
LGTM — great work on the validation pipeline and URL handling. Two non-blocking comments/questions:

**1. URL normalization only on blur (suggestion)**

The `normalizeUrlScheme` call in `Input.handleBlur` auto-prepends `https://` on blur, which is nice UX. But it is the only place normalization happens. If blur does not fire before submit (edge case, programmatic submit, etc.), the raw bare domain hits validation and is rejected.

Consider also normalizing in `toCreateInput` as a safety net, so the blur behavior is a UX convenience rather than a correctness requirement:

```ts
website: formData.website
  ? normalizeUrlScheme(formData.website) ?? undefined
  : undefined,
```

Non-blocking — in practice the blur fires before the click handler in React 18.

**2. `z.email()` vs shared `Regex.email` (question)**

`Regex.email` is defined in the shared module (stricter — requires a dot in the domain) but `formSchema` uses Zod's built-in `.email()` (more permissive — allows `user@localhost`). Was the intent to eventually use the shared regex for consistency with the custom URL/domain patterns, or is `z.email()` fine here?

PR #2202, opened 2026-06-30
feat(permissions): migrate unified permissions to Expo RN

vecchp:
## Summary

Migrates the unified permissions system to the Expo RN app and consolidates shared user provider logic into ba-platform.

### Changes

**Expo RN (`betterangels`):**
- Added `permissions` field to UserProvider GraphQL query
- Created `asyncStorageAdapter` for orgLink in React Native
- Added `links` support to Expo ApolloClientProvider
- Wired `ActiveOrgProvider` + `orgLink` into Expo root layout

**ba-platform (shared):**
- Created `createUserProvider` factory — encapsulates Apollo query lifecycle, auth detection, and ActiveOrgProvider embedding
- Created pre-configured `currentUserProvider` using `CurrentOrgUserDocument`
- Added `CurrentUser` type for admin/dashboard apps

**shelter-web:** Already fully migrated (orgLink + permissions + ActiveOrgProvider in OperatorApp)

**ba-admin:** Refactored to use the shared factory (60 fewer lines)

### Deleted dead code
- ba-admin UserProvider.tsx, useUser.ts, UserContext.ts
- Expo UserContext.ts, hooks/user/useUser.ts

### Type Checks
All packages pass `tsc --noEmit` and `nx typecheck`.


libs/expo/betterangels/src/lib/screens/Interactions/index.tsx
```diff
@@ -3,8 +3,8 @@ import { SearchBar } from '@monorepo/expo/shared/ui-components';
 import { ElementType, useCallback, useState } from 'react';
 import { StyleSheet, View } from 'react-native';
 import { NoteType, toNoteFilter } from '../../apollo';
-import useUser from '../../hooks/user/useUser';
-import { TUser } from '../../providers/user/UserContext';
+import { useUser } from '../../providers/user/UserProvider';
+import { TUser } from '../../providers/user/UserProvider';
```
mikefeldberg: surprised this wasn't autocorrected on save. same in `InteractionsHmis/index.tsx`

libs/ba-platform/src/lib/react/providers/user/createUserProvider.tsx
```diff
@@ -0,0 +1,184 @@
...
+          setUser(parseUser(res.data?.currentUser));
+        }
+      },
+      // eslint-disable-next-line react-hooks/exhaustive-deps
+      [],
+    );
+
+    useEffect(() => {
+      if (!loading) {
+        updateUser({ data, errors: error ? [error] : undefined });
```
mikefeldberg, line 152: robot suggestion: 

`error` is wrapped as `[error]` but typed inconsistently

When `error` exists (an `ApolloError`), you wrap it in an array:  `errors: error ? [error] : undefined`. But `ApolloError` has `.message` and `.extensions` properties at the top-level (not per-error). This works because `isUnauthenticated` reads `.message` / `.extensions` — but the caller's `isUnauthenticated` in the Expo provider checks `e.extensions?.['code']`, which will be `undefined` on a raw `ApolloError` (the code lives in `error.graphQLErrors[i].extensions`). This may be a latent bug: if the network request fails with a GraphQL-level auth error, the Expo app's `isUnauthenticated` check might not trigger correctly because you're passing the top-level `ApolloError` object rather than the individual `graphQLErrors`.

 Recommend changing line 141 to:

`errors: error?.graphQLErrors ?? undefined`

vecchp: Good catch on the typing concern, but this is actually an Apollo v4 difference — `ApolloError` was [removed in v4](https://www.apollographql.com/docs/react/v4-migration#apolloerror). The `error` from `useQuery` is now typed as `ErrorLike` (`{ message, name, stack? }`), which has no `graphQLErrors` property. Trying `error?.graphQLErrors` gives:

```
TS2339: Property 'graphQLErrors' does not exist on type 'ErrorLike'
```

The current pattern `error ? [error] : undefined` is the correct approach for v4. At runtime the error *is* a `GraphQLError` from the `graphql` package, which does carry `extensions`, so the Expo `isUnauthenticated` check (`e.extensions?.['code']`) works correctly with optional chaining.

That said, the `isUnauthenticated` contract is intentionally permissive (`extensions?` is optional) to support both the Expo provider (which checks `extensions.code`) and the admin/shelter providers (which check `message` text). I verified this works correctly in the Expo preview build.

mikefeldberg (approved):

PR #2206, opened 2026-06-30
Fix shelter photo cache sync after upload

tglaz:
* stabilize `_typename` to avoid using magic strings that can break
  * was the cause of this phot breakage - couldn't update apollo cache
* update `createShelterTypePolicies` to accept extra policies (for shelter-operator)
  * largely needed avoid circular deps, but also good to define separately


vecchp (dismissed):
Great work. The `Extract<>` pattern for `__typename` constants is especially nice — catches renames at build time instead of runtime.

Two small suggestions inline, none blocking.

apps/shelter-web/src/main.tsx
```diff
@@ -1,13 +1,14 @@
...
+import { orgLink } from '@monorepo/ba-platform';
 import { createApolloClient } from '@monorepo/react/shared/apollo';
 import {
   ApiConfigProvider,
   ShelterFeatureControlProvider,
   UserProvider,
   createShelterTypePolicies,
 } from '@monorepo/react/shelter';
-import { orgLink } from '@monorepo/ba-platform';
+import { createOperatorTypePolicies } from '@monorepo/react/shelter-operator';
```
vecchp, line 11: `createOperatorTypePolicies()` returns `{}` — a no-op right now. This import creates a `shelter-web` → `shelter-operator` dependency for no current benefit.

```diff
-import { createOperatorTypePolicies } from '@monorepo/react/shelter-operator';
```

And below:

```diff
   typePolicies: createShelterTypePolicies({
     isDevEnv,
-    extraPolicies: createOperatorTypePolicies(),
   }),
```

When operator policies arrive, they can be passed through a shared interface in `ba-platform` instead of creating this dependency direction.

libs/react/shelter-operator/src/lib/apollo/cachePolicies/createOperatorTypePolicies.ts
```diff
@@ -0,0 +1,17 @@
+import { TypePolicies } from '@apollo/client';
```
vecchp, line 1: Nice scaffolding for operator-specific type policies.

When real policies are needed, the `OPERATOR_SHELTER_TYPENAME` import can be uncommented and used as the type policy key:

```ts
import { OPERATOR_SHELTER_TYPENAME } from '../../hooks/useShelterOperatorProfile/constants';

export function createOperatorTypePolicies(): TypePolicies {
  return {
    [OPERATOR_SHELTER_TYPENAME]: {
      keyFields: ['id'],
    },
  };
}
```

The `extraPolicies` extension point on `createShelterTypePolicies` is a clean design — operator policies get merged into the shared cache policy without forking.

vecchp (approved):
Great work. The `Extract<>` pattern for `__typename` constants is especially nice — catches renames at build time instead of runtime.

Two small suggestions inline, none blocking.

apps/shelter-web/src/main.tsx
tglaz: think we'll want to update really soon - so I left it
need to add `createShelterApolloCachePolicyRegistry` as technically the operator `users query`, and others, are paginated.

tglaz: would have done those but want to take a look at updating the code to generate those policies. 
the existing functions are a bit (or rather very) convoluted partly because of type safety, but with hooks exporting `_typename` as a safe const, may be able to simplify it quite a bit. but wasn't the time to mess with it yet

PR #2209, opened 2026-07-03
Handle server errors in UI for ShelterBasicInfo (SDB-241)

tglaz:
### Handle server errors in UI for ShelterProfile updates
SDB-241

* shelter-operator/codegen`
  * run new custom plugin: `tools/codegen/plugins/operation-meta-plugin.cjs`
  * add `nonOptionalTypename` (generates __typename as required -it already is)
  * generate new types (in shelter-operator only)
* @monorepo/ba-platform
  * add `getFieldErrorsOrThrow` in `/graphql/responseUtils` dir
  * define `BaError` in `/errors/BaError.ts
    * it's meant as a custom error where we can "trust" the message
* base-ui
  * add `resolveErrorMessage` util with tests
  
**Next:**
* export static const for `OperationInfo` __typename in ba-platform lib
* add tests for `getFieldErrorsOrThrow` and apply elsewhere in shelter-operator code
  * a bit more cleanup to `responseUtils`
  * handle extension errors - and extension field errors
* add `operation-meta-plugin` plugin to run in other `codegen` - possibly via separte ticket


vecchp (approved):
## Overall: ✅ Approve

This is the right approach for type-safe error handling against strawberry-django's `OperationInfo` union pattern. The `operation-meta-plugin` codegen approach is genuinely good — replacing string-literal `__typename` checks with compile-time constants is the correct long-term direction here.

Nice that your "Next" list already covers extension errors, meta plugin rollout to other codegen targets, tests for `getFieldErrorsOrThrow`, and `responseUtils` cleanup. Those all make sense as follow-up.

A few small inline suggestions below — none blocking, but worth addressing before the pattern spreads to more forms.

### Design notes (discuss, no changes needed)

**Throw vs return.** This uses throw-semantics while the Expo app's equivalent uses return-semantics. Both valid — just something to standardize when consolidating.

**Two error paths.** OperationInfo (handled here ✅) and extensions-level errors (planned in "Next" ✅). The `ba-platform` library is on track to replace both Expo and Admin `extract*` utilities once both paths are covered.

**strawberry-django best practice #3.** The [official docs](https://strawberry.rocks/docs/django/guide/error-handling#best-practices) recommend error codes for programmatic client handling: `raise ValidationError("...", code="OUT_OF_STOCK")`. `OperationMessage.code` exists for exactly this reason. Adding `code` to `FieldError` now aligns with upstream guidance and unblocks code-based error handling later.

libs/ba-platform/src/lib/apollo/graphql/responseUtils/types.ts
```diff
@@ -0,0 +1,5 @@
+/** A field-level validation error returned from the API */
```
vecchp: Suggestion: add optional `code` field. Strawberry-django [best practice #3](https://strawberry.rocks/docs/django/guide/error-handling#best-practices) recommends error codes for programmatic client handling. `OperationMessage.code` already exists in the schema.

```suggestion
export type FieldError = {
  field: string;
  message: string;
  /** Error code from the API, if set. */
  code?: string;
};
```

libs/ba-platform/src/lib/apollo/graphql/responseUtils/partitionMessages.ts
```diff
@@ -0,0 +1,35 @@
...
+    if (msg.__typename !== 'OperationMessage') {
+      continue;
+    }
+
+    if (
+      msg.kind === OperationMessageKind.Validation &&
+      msg.field &&
+      (!fieldSet || fieldSet.has(msg.field))
+    ) {
+      fieldErrors.push({ field: msg.field, message: msg.message });
```
vecchp: Suggestion: thread `msg.code` through so callers can use it for programmatic error handling instead of string-parsing messages.

```suggestion
fieldErrors.push({ field: msg.field, message: msg.message, code: msg.code ?? undefined });
```

libs/react/shared/src/lib/utils/forms/applyFieldErrors.ts
```diff
@@ -0,0 +1,16 @@
+import { FieldValues, Path, UseFormSetError } from 'react-hook-form';
+
+type FieldError = { field: string; message: string };
```
vecchp, line 3: Non-blocking: this local `FieldError` type duplicates the one in `@monorepo/ba-platform`. If the dependency graph allows, consider importing it instead. Otherwise add a comment noting the intentional duplication — same type defined in two places will drift.

libs/react/shelter-operator/src/lib/components/base-ui/input/utils/resolveErrorMessage.ts
```diff
@@ -0,0 +1,79 @@
+import type { InputDataType } from '../types';
+
+const validationPatterns: Record<InputDataType, RegExp> = {
+  email: /^[^\s@]+@[^\s@]+\.[^\s@]+$/,
+  'phone-number': /^[\d\s\-+()]*\d[\d\s\-+()]*$/,
+  number: /^-?\d+(\.\d+)?$/,
+  time: /^([01]\d|2[0-3]):([0-5]\d)$/,
+  date: /^\d{4}-(?:0[1-9]|1[0-2])-(?:0[1-9]|[12]\d|3[01])$/,
+  // TODO: import Regex.url from ts-lib once created - import issues otherwise
+  url: /^https?:\/\/[\w.-]+\.[a-zA-Z]{2,63}(\/\S*)?(\?\S*)?$/i,
```
vecchp, line 10: Suggestion: resolve the TODO about importing Regex.url. This PR already fixed the canonical Regex.url in libs/react/shared/src/lib/static/regex.ts (same TLD fix). If there is a genuine import ordering issue, consider extracting the shared regex to a constants file that both can import. Duplicated regexes drift over time.

libs/ba-platform/src/lib/apollo/graphql/responseUtils/getFieldErrorsOrThrow.ts
```diff
@@ -0,0 +1,58 @@
...
+  fields?: string[];
+};
+
+export function getFieldErrorsOrThrow(
+  params: GetFieldErrorsOrThrowParams
+): FieldError[] {
+  const { response, operationKey, successTypename, fields } = params;
+
+  if (response.errors?.length) {
+    throw new BaError(response.errors[0].message);
```
vecchp: Suggestion: consider logging all transport errors, not just the first. If the server returns multiple, the rest are silently dropped.

```suggestion
if (response.errors?.length) {
  if (response.errors.length > 1) {
    console.warn('[getFieldErrorsOrThrow] Multiple transport errors:', response.errors);
  }
  throw new BaError(response.errors[0].message);
}
```

tglaz: the error message generation got updated in this open PR: #2211 
the messages are concatenated there - but i'll take a follow up look to confirm and make sure the tests reflect the correct message produced

libs/ba-platform/src/lib/apollo/graphql/responseUtils/partitionMessages.ts
tglaz: i'll probably do that - but in a follow up PR
i'll need to review how these codes are organized in FE and also apply them across the board.
but will need to still review where they'd be defined.

not sure want react/shared libs importing these codes (or any code) from ba-platform lib - but will have to review this still

libs/ba-platform/src/lib/apollo/graphql/responseUtils/types.ts
tglaz: answered above

libs/react/shared/src/lib/utils/forms/applyFieldErrors.ts
tglaz: this was intentional
1. `react/shared` lib probably should't import from `ba-platform`
2. its' a local type that's simple enough that it doesn't matter - it's defined explicitly as a type but cold also be inlined in the function signature - ts will handle either way (no need for any dependency)

libs/react/shelter-operator/src/lib/components/base-ui/input/utils/resolveErrorMessage.ts
tglaz: i didn't want to add another lib in this PR - prefer to handle that separately (which will trigger a bunch of changes - but none critical or immediate)
all this file change does is to move the code into a util and sync the regex

PR #2211, opened 2026-07-06
update getFieldErrorsOrThrow with docs and tests (SDB-241)

tglaz:
### Handle server errors in UI
SDB-241

* udpate `getFieldErrorsOrThrow`
  * will throw unless:
    * no errors
    * all errors match requested `fields`
* add BE docs for `GraphQL Error Handling`
  * docs/graphql_errors.md

#### Test:

```
  getFieldErrorsOrThrow
    returns []
      success
        ✓ when __typename matches successTypename
    returns FieldError[]
      with extensions.errors
        all errors match fields filter
          ✓ recognized errorCode: returns mapped message
          ✓ unrecognized errorCode: falls back to server message
          ✓ no errorCode or message: returns "Invalid value"
          ✓ multiple errors: returns all mapped messages
      with OperationInfo
        VALIDATION kind: all messages match fields filter
          ✓ single message: returns message
          ✓ multiple messages: returns matching messages
    throws BaError
      BaPermissionError
        with response errors
          UNAUTHENTICATED
            ✓ with message: throws BaPermissionError with server message (4 ms)
            ✓ missing message: throws default BaPermissionError
        with OperationInfo
          PERMISSION kind
            ✓ with message: throws BaPermissionError with server message (1 ms)
            ✓ with field: throws BaPermissionError with contextual message
            ✓ missing message: throws BaPermissionError with default msg
    throws Error
      with extensions.errors
        with field
          ✓ non-field error present: throws Error (1 ms)
          ✓ no fields matching filter: throws Error
        without field
          ✓ throws Error with fallback message
      with response errors
        NOT_FOUND
          ✓ throws Error with the message (1 ms)
        with server message
          ✓ throws Error with server message
        empty
          ✓ throws Error with fallback
        with multiple messages
          ✓ joins server error messages (1 ms)
      with OperationInfo
        VALIDATION kind
          ✓ non-VALIDATION message present: throws Error
          ✓ no fields matching filter: throws Error
          ✓ without field: throws Error
        ERROR kind
          ✓ with field: throws Error (1 ms)
          ✓ without field: throws Error (2 ms)
        INFO kind
          ✓ with field: throws Error
          ✓ without field: throws Error (1 ms)
        WARNING kind
          ✓ with field: throws Error
          ✓ without field: throws Error
      edge cases
        ✓ empty fields array: throws Error (1 ms)
        ✓ no data: throws Error
        ✓ unexpected __typename: throws Error
        ✓ OperationInfo with no messages: throws Error with fallback
```


libs/ba-platform/src/lib/apollo/graphql/response/utils/composeErrorMessage.ts
```diff
@@ -0,0 +1,21 @@
+import { DEFAULT_GQL_ERROR_MESSAGE } from '../../constants';
+
+/**
+ * Joins non-empty messages with "; " separators into a single error string.
+ * Falls back to `DEFAULT_GQL_ERROR_MESSAGE` when all messages are empty.
+ *
+ * @param messages - Messages from `response.errors` to join.
+ */
+export function composeErrorMessage(
+  messages: (string | null | undefined)[]
```
Davit-BetterAngels: messages: (string | null | undefined)[] = []

libs/ba-platform/src/lib/apollo/graphql/response/utils/filterOperationMessages.ts
```diff
@@ -0,0 +1,30 @@
...
+): OperationMessage[] {
+  if (!opInfo) {
+    return [];
+  }
+
+  const messagesByKind = opInfo.messages.filter((m) => m.kind === kind);
+
+  if (!fields?.length) {
+    return messagesByKind;
+  }
```
Davit-BetterAngels: if (fields === undefined) {
  return messagesByKind;
}

if (!fields.length) {
  return [];
}

libs/ba-platform/src/lib/apollo/graphql/response/getFieldErrorsOrThrow.ts
```diff
@@ -0,0 +1,182 @@
...
+    // Extensions with field-level errors are recoverable — return for form display
+    if (extensionFieldErrors.length) {
+      return extensionFieldErrors;
+    }
+  }
+
+  // Top-level errors (non-extension)
+  if (response.errors?.length) {
+    const errorMessages = response.errors?.map((e) => e.message) ?? [];
+    const errMesssage = composeErrorMessage(errorMessages);
```
vecchp: 🐍 sss, seems intentional but why?

libs/ba-platform/src/lib/apollo/graphql/response/getFieldErrorsOrThrow.ts
```diff
@@ -0,0 +1,182 @@
...
+  if (extensionErrors.length) {
+    const matchedExtensionErrors = filterExtensionErrors({
+      errors: extensionErrors,
+      key: 'field',
+      filters: fields,
+    });
+
+    // Some extension errors didn't match (no field, or field not in filter)
+    if (matchedExtensionErrors.length !== extensionErrors.length) {
+      throw new Error('An unexpected error occurred.');
```
vecchp: This line is repeated 3 times in the same file maybe a put into a shared constant?

libs/ba-platform/src/lib/apollo/graphql/response/getFieldErrorsOrThrow.ts
```diff
@@ -0,0 +1,182 @@
...
+  // Everything else → throw
+  // If messages exist but no field errors matched (e.g. fields filter
+  // excluded everything), use a generic fallback rather than leaking an
+  // uncorrelated message from the server.
+  if (fields.length && operationInfo?.messages?.length) {
+    throw new Error('An unexpected error occurred.');
+  }
+
+  const errorMessages = response.errors?.map((e) => e.message) ?? [];
+  const errMessage = composeErrorMessage(errorMessages);
```
vecchp: We seem to be doing this twice in the function?

libs/ba-platform/src/lib/apollo/graphql/response/getFieldErrorsOrThrow.test.ts
```diff
@@ -0,0 +1,565 @@
...
+          kind: m.kind,
+          field: m.field ?? null,
+          message: m.message ?? '',
+        })),
+      },
+    },
+  };
+}
+
+function resonseWithErors(
```
vecchp: resonseWithErors?

libs/ba-platform/src/lib/apollo/graphql/response/getFieldErrorsOrThrow.ts
```diff
@@ -0,0 +1,182 @@
...
+  params: GetFieldErrorsOrThrowParams
+): FieldError[] {
+  const { response, operationKey, successTypename, fields } = params;
+
+  if (!fields.length) {
+    console.error(
+      '[getFieldErrorsOrThrow] requires a non-empty fields argument.'
+    );
+
+    throw new Error('An unexpected error occurred.');
```
Davit-BetterAngels: make this constant ?

Davit-BetterAngels (approved):

libs/ba-platform/src/lib/apollo/graphql/response/getFieldErrorsOrThrow.ts
```diff
@@ -0,0 +1,182 @@
...
+import { BaApiErrorCode } from '../types';
+import { getOperationInfo } from './getOperationInfo';
+import { isUnauthenticatedError } from './isUnauthenticatedError';
+import type { FieldError, GraphQLResponse } from './types';
+import { composeErrorMessage } from './utils/composeErrorMessage';
+import { filterExtensionErrors } from './utils/filterExtensionErrors';
+import { filterOperationMessages } from './utils/filterOperationMessages';
+import { getExtensionErrors } from './utils/getExtensionErrors';
+
+const DEFAULT_ERROR_MESSAGE = 'Invalid value';
```
vecchp: We have `DEFAULT_ERROR_MESSAGE` `DEFAULT_GQL_ERROR_MESSAGE` and then the hard-coded 3x `An unexpected error occurred.`.  Is this necessary? If not can you document the differences somewhere?

libs/ba-platform/src/lib/apollo/graphql/response/getFieldErrorsOrThrow.ts
```diff
@@ -0,0 +1,182 @@
...
+    const hasNonValidationMessages = operationInfo?.messages.some(
+      (m) =>
+        m.kind !== OperationMessageKind.Validation ||
+        !m.field ||
+        (m.field && !fields.includes(m.field))
+    );
+
+    if (hasNonValidationMessages) {
+      throw new Error('An unexpected error occurred.');
+    }
```
vecchp: I believe this is the case

The `!m.field` clause means a VALIDATION message with `field: null` (a legitimate non-field validation error from Django's full_clean()) will set `hasNonValidationMessages = true`

Server returns `OperationInfo` with two messages:
* `{ kind: 'VALIDATION', field: 'name', message: 'Required' }`
* `{ kind: 'VALIDATION', field: null, message: 'Form is invalid' }`

Called with `fields = ['name']`. Tracing the logic:

1. `validationErrors` → `[{ field: 'name', message: 'Required' }]`  the non-field message is correctly excluded by filterOperationMessages)
2. `hasNonValidationMessages` iterates all messages, hits the non-field one:
kind !== VALIDATION → false
!null → true ← short-circuits
3. `Throws `An unexpected error occurred.` instead of returning the recoverable field error

Suggested fix: exclude `!m.field` from the predicate for `VALIDATION` messages, or adjust the intent: only flag messages from a different kind (`ERROR`, `INFO`, `WARNING`) as non-validation. Something like:

```ts
const hasNonValidationMessages = operationInfo?.messages.some(
  (m) =>
    m.kind !== OperationMessageKind.Validation ||
    (m.kind === OperationMessageKind.Validation &&
      m.field &&
      !fields.includes(m.field))
);
```

libs/ba-platform/src/lib/apollo/graphql/response/getFieldErrorsOrThrow.ts
tglaz: `composeErrorMessage` will concatenate error message into single string - if that's what you were referring to

libs/ba-platform/src/lib/errors/BaError.ts
```diff
@@ -2,7 +2,7 @@
  * A known BetterAngels error
  */
 export class BaError extends Error {
-  constructor(message: string) {
+  constructor(message = 'Sorry, something went wrong.') {
```
vecchp, line 5: Another default it seems?

libs/ba-platform/src/lib/apollo/graphql/response/getFieldErrorsOrThrow.ts
tglaz: yeah, I already have a `DEFAULT_GQL_ERROR_MESSAGE` const - will update

vecchp:
**Great work on this — the docs are excellent and the test coverage is thorough. Did a broader audit of Strawberry error handling + frontend consumption across the monorepo. A few FYIs surfaced for follow-up (out of scope for this PR):**

**Backend:**
- `validate_client_profile_data` raises `GraphQLError` instead of `ValidationError`, bypassing Strawberry's `_handle_exception`. `OperationMessage.code` already supports machine-readable codes — the standard Strawberry path could replace the custom extensions path and eliminate the need for the frontend to handle two different response shapes.
- `PermissionError` (Python built-in) is used instead of `PermissionDenied` (Django) in several mutations — not caught by `_handle_exception`, already flagged as "likely a bug" in the docs.

**Frontend:**
- `BaPermissionError` is thrown but never caught distinctively by any consumer — no redirect-to-login flow on the web side. The React web clients also have no Apollo Link error middleware (unlike Expo which has one).
- 5 shelter-operator segments (`Details`, `Ecosystem`, `Policies`, `Services`, `Schedules`) still have `// TODO: handle specific OperationInfo field errors via SDB-241` and discard `OperationInfo` entirely.
- Bed/Room/Reservation forms manually check `__typename === 'OperationInfo'` and only read the first message — never use `getFieldErrorsOrThrow`.
- `ShelterWeb UserProvider` does substring matching on error messages (`e.message.includes('User is not logged in.'`) instead of using `isUnauthenticatedError`.
- Two parallel error-handling systems exist: `getFieldErrorsOrThrow` (ba-platform) vs `extractOperationErrors` (expo), with separate `isUnauthorizedError` implementations.

libs/ba-platform/src/lib/apollo/graphql/response/getFieldErrorsOrThrow.ts
```diff
@@ -0,0 +1,182 @@
...
+    if (matchedExtensionErrors.length !== extensionErrors.length) {
+      throw new Error('An unexpected error occurred.');
+    }
+
+    const extensionFieldErrors: FieldError[] = matchedExtensionErrors.flatMap(
+      (e) => {
+        if (!e.field) {
+          return [];
+        }
+
```
vecchp, line 80: **`flatMap` with `!field` guard is dead code** — the preceding filter functions already guarantee string fields:

- **Extension path (line 78):** `filterExtensionErrors({ key: 'field' })` only keeps elements where `typeof e.field === 'string' && filterSet.has(e.field)`. `!e.field` would only trigger if `e.field === ""`, which requires `fields` to contain `""` — impossible since `fields` comes from `Object.keys(formSchema.shape)`. Practically dead code.

- **OperationInfo path (line 143):** `filterOperationMessages` already does `m.field && fields.includes(m.field)`. The `m.field &&` truthiness check guarantees `!m.field` is always `false`. Strictly dead code.

Replace both `flatMap` with `.map()` and drop the guard.

libs/ba-platform/src/lib/apollo/graphql/response/getFieldErrorsOrThrow.ts
```diff
@@ -0,0 +1,182 @@
...
+ * 3. OperationInfo payload — VALIDATION messages matching `fields` →
+ *    `FieldError[]`; PERMISSION → `BaPermissionError`; anything else → `Error`.
+ */
+export function getFieldErrorsOrThrow(
+  params: GetFieldErrorsOrThrowParams
+): FieldError[] {
+  const { response, operationKey, successTypename, fields } = params;
+
+  if (!fields.length) {
+    console.error(
```
vecchp, line 53: **`console.error` + generic throw loses context** — the diagnostic message goes to `console.error` (suppressed in tests, invisible in production), then a different generic message is thrown. Merge the diagnostic into the throw so the error object carries the context:

```typescript
throw new Error('[getFieldErrorsOrThrow] requires a non-empty fields argument.');
```

libs/ba-platform/src/lib/apollo/graphql/response/getFieldErrorsOrThrow.ts
```diff
@@ -0,0 +1,182 @@
...
+        ];
+      }
+    );
+
+    // Extensions with field-level errors are recoverable — return for form display
+    if (extensionFieldErrors.length) {
+      return extensionFieldErrors;
+    }
+  }
+
```
vecchp, line 98: Three issues in this block:

**1. `composeErrorMessage` runs before the UNAUTHENTICATED check** — wasted string allocation when the error is unauthenticated. Swap the order: check `isUnauthenticatedError` first, compose the message only if needed.

**2. `find((e) => e.message)` picks the wrong error** — if multiple `response.errors` exist and the unauthenticated one has no message but another does, `BaPermissionError` gets thrown with a semantically wrong message (e.g. `"Not Found."`). Should filter on the UNAUTHENTICATED code:

```typescript
response.errors.find(
    (e) => e.extensions?.code === 'UNAUTHENTICATED' && e.message
)?.message || undefined
```

(Doesn't manifest today since the backend never returns multiple errors, but it's a landmine.)

**3. `|| undefined` after `?.message`** — `?.message` already produces `undefined` when `.find()` misses. The `|| undefined` only fires for empty-string messages (`""`), treating them as "no message." If that's intentional for empty-string fallback to the default, it's correct but not obvious — a comment would help.

Cleaned up:
```typescript
if (response.errors?.length) {
    if (isUnauthenticatedError(response.errors)) {
        throw new BaPermissionError(
            response.errors.find(
                (e) => e.extensions?.code === 'UNAUTHENTICATED' && e.message
            )?.message || undefined
        );
    }

    const errorMessages = response.errors.map((e) => e.message);
    throw new Error(composeErrorMessage(errorMessages));
}
```

libs/ba-platform/src/lib/apollo/graphql/response/utils/filterOperationMessages.ts
tglaz: i'm not gonna enforce a non-empty arr, in which case the current versions should be ok?

with passing `fields = []` it'd return nothing, which seems like a caller mistake.
no fields technically just means don't filter on fields?

```
if (!fields.length) {
return [];
}
```

libs/ba-platform/src/lib/apollo/graphql/response/getFieldErrorsOrThrow.ts
tglaz: they're all different, but i'll update the naming so makes more sense

libs/ba-platform/src/lib/errors/BaError.ts
tglaz: i'd keep this as is - 
Error should probably own its fallback message, and typically the caller would pass one in - else they don't care anyway

libs/ba-platform/src/lib/apollo/graphql/response/getFieldErrorsOrThrow.ts
tglaz: replaced it with a `filterRecoverableOperationMessages` 

```typescript
  const { recoverable, unrecoverable } = filterRecoverableOperationMessages(
    operationInfo?.messages ?? [],
    fields
  );
```

libs/ba-platform/src/lib/apollo/graphql/response/getFieldErrorsOrThrow.ts
tglaz: replaced with `filterRecoverableOperationMessages()` that removes this

libs/ba-platform/src/lib/apollo/graphql/response/getFieldErrorsOrThrow.ts
tglaz: i'll probably let that go for now - but kind of valid

libs/ba-platform/src/lib/apollo/graphql/response/getFieldErrorsOrThrow.ts
tglaz: i hoisted the `composedErrorMessage` definition

vecchp (approved):

PR #2214, opened 2026-07-07
open now include unknown and select all (SDB-259)

Davit-BetterAngels:
SDB-259

## Summary by Sourcery

Extend shelter search "Open Now" filters to support multi-select with a select-all option and an "include unknown" toggle, and wire these into filter state, pills, and type definitions.

New Features:
- Add multi-select checkbox group with select-all for "Open Now" schedule type filters.
- Introduce an "Include unknown" option for "Open Now" filtering with corresponding filter pill display.

Enhancements:
- Update shelter filter state handling and configuration to support the new open-now-related flags without impacting other property filters.


shiozakiyuko (commented):
Hi, @Davit-BetterAngels 

I noticed a few things. 

- The Select All works. 👍🏻
- I thought we decided to keep the Include Unknown in the checkbox like everything else, not outside of the checkbox group. Can we fix that?
- The results of filtering are inconsistent. Filtering by the Operating Hours are all good but the search results based on filtering Intake, Meal Services, and Staff Availability are not correct.  

image

mikefeldberg (commented):
afaict, this is purely a cosmetic change. `openNowIncludeNull` isn't included in query vars (or supported by the backend). Is there an accompanying backend ticket to wire this up to? Does it make sense to add this as a standalone change now, rather than as a follow up to SDB-260/#2215

tglaz:
closing.
handled via:
FE: #2386
BE: #2385

PR #2216, opened 2026-07-07
Inconsistent total of search result number (SDB-261, SDB-239)

Davit-BetterAngels:
## Summary

Fixes **SDB-261** (Inconsistent total of search result number) and **SDB-239** (persist map position after shelter navigation).

Navigating home → shelter detail → back now restores the **exact same viewport (center + zoom)** — it does not re-center on the last clicked pin — so the same pins stay visible and the result total stays consistent. Returning also **does not re-query the backend** when nothing changed (verified: 0 GraphQL POSTs on return, vs 2 before).

### Implementation
- **Fully controlled camera**: `Map` is driven by `center`/`zoom` props + `onCameraChange`; user pan/zoom and `fitBounds` results are reported back to the parent; `onMapIdle` reports the actual rendered bounds.
- **Viewport persistence (jotai)**: `savedMapViewportAtom` (module-scoped, type-safe) is written by `ShelterCard` on navigate and consumed once by the home init effect. The Map mounts directly at the saved camera.
- **Search-area persistence + skip-guard**: `shelterMapBoundsFilterAtom` persists the last-searched bounds so returning re-runs the query with identical variables (Apollo default `cache-first` = cache hit, no network). The location effect skips firing a search entirely when the restored bounds exactly match the last search (`sameMapBounds`) — and correctly falls through to a fresh search if the viewport genuinely changed (e.g. window resize).
- Restore by **center + zoom**, not `fitBounds(savedBounds)` — `fitBounds` picks a different zoom (padding), which shifted pins and changed the count.
- **Architecture**: all map/search orchestration extracted into a `useShelterMapSearch` hook (unit-tested); `HomePage` is a thin ~130-line presentation component.

### SDB-261 acceptance criteria
1. ✅ Returning preserves the exact same map coordinates, boundaries, and zoom level.
2. ✅ The total number of results/pins stays identical (search uses the actual restored viewport).
3. ✅ The map does not re-center or shift around the last clicked pin.

### SDB-239 acceptance criteria
1. ✅ HomePage map position persists when navigating to and back from shelter details.

### Tests (64)
- `useShelterMapSearch`: LA-center fallback search, geolocation → idle search, **skip-guard** (matching restored bounds → no re-search; differing bounds → re-search), map-area search.
- `ShelterCard`: saves center + zoom to the atom before navigating; still navigates without saving when the map isn't ready.
- `mapViewportFromMap`, `sameMapBounds`, `symmetricBoundsAroundPinCentroid` unit tests.
- Verified in the browser: restore adds **0** GraphQL POSTs (was 2); exact viewport + count restored; pan (no snap-back); search-this-area; name search (`fitBounds`).


shiozakiyuko (commented):
Hi @Davit-BetterAngels 

I'm still noticing inconsistent search results with this change. Specifically:

**Pin shifting**: Sometimes the map stays in the expected viewport, but the number of pins shifts.
**Map resetting**: Sometimes the map completely readjusts back to the full LA view, and the search result counts don't stay the same.

libs/react/shelter/src/lib/components/ShelterCard/ShelterCard.tsx
```diff
@@ -73,6 +74,12 @@ export function ShelterCard(props: TShelterCard) {
         zoom: map?.getZoom(),
       })
     );
+    if (currentBounds) {
+      sessionStorage.setItem(
+        'savedMapBounds',
```
tglaz: would be safer to save these sessionStorage keys somewhere as constants to avoid typos etc..

tglaz (approved):

shiozakiyuko (commented):
Hi @Davit-BetterAngels 

I don't see any changes unfortunately. I tried several different locations but it still shift to the center when I close the detail page. I also noticed that when I tap on a pin especially from the westside, it goes to downtown area when I close the detail page. When I zoomed in, the number of shelters weren't updated either.
Please check the video I attached. Thanks!

https://github.com/user-attachments/assets/babefa8e-ecda-42f8-975b-2b83cbee947c

shiozakiyuko (approved):
It looks good. Approved.

PR #2217, opened 2026-07-07
feat: consolidated presigned S3 upload pipeline — bump limits to 50 MB, add note attachments (DEV-2464)

vecchp:
DEV-2464

## Summary

Consolidates the presigned S3 upload flow across all 3 domains (shelter photos,
client documents, note attachments) into a shared, configuration-driven pipeline
in `common/services/attachment_upload.py`. Bumps upload limits from 10 MiB →
50 MiB (50 × 1024 × 1024) and adds full note attachment support.

> **35 files, +2,950 / −1,149 lines.** ~60% of additions are tests, ~8% are
> auto-generated TypeScript, ~6% is a design doc. ~780 lines of real code.

See [`docs/presigned-upload-pipeline.md`](https://github.com/BetterAngelsLA/monorepo/blob/feat/bump-upload-limit-50mb/docs/presigned-upload-pipeline.md) for the full architecture design doc.

### New: generic presigned upload pipeline
- `common/services/attachment_upload.py` — 3-phase pipeline (generate → upload → resolve)
- `common/services/upload_token.py` — stateless HMAC-signed upload tokens scoped to user + S3 key + service
- `common/services/types.py` — typed dataclasses: `AttachmentUploadConfig`, `GenerateUploadItem`, `ResolveUploadItem`
- `common/services/exceptions.py` — domain exceptions: `InvalidContentTypeError`, `InvalidUploadTokenError`, `S3KeyNotFoundError`

### Refactored: domain services use shared pipeline
- `clients/services/client_document.py` — delegates to generic pipeline
- `clients/services/client_profile_photo.py` — delegates to generic pipeline
- `shelters/services/shelter_photo.py` — delegates to generic pipeline; typed `ShelterPhotoResolveItem` preserves enum
- `notes/services.py` — new note attachment upload/resolve with org-scoped permissions

### New: note attachment GraphQL mutations
- `generateNoteAttachmentUploads` — generate presigned URLs + upload tokens for note attachments
- `resolveNoteAttachmentUploads` — validate tokens + S3 keys → create Attachment rows

### Settings
- Upload limits: all 4 env vars default to 50 MiB (50 × 1024 × 1024)
- `DATA_UPLOAD_MAX_MEMORY_SIZE` / `FILE_UPLOAD_MAX_MEMORY_SIZE` left at Django defaults (files go via presigned S3, never through request body)
- Upload limits evaluated after `.env` files are loaded so local overrides are respected
- Token expiry: 5 minutes

### Deprecation
- `Attachment.save()` override is deprecated in favor of the service layer; emits `DeprecationWarning`

## Ticket

[DEV-2464](https://betterangels.atlassian.net/browse/DEV-2464)

[DEV-2464]: https://betterangels.atlassian.net/browse/DEV-2464?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/betterangels-backend/shelters/schema.py
```diff
@@ -263,11 +251,12 @@ def resolve_shelter_photo_uploads(
         user = cast(User, get_current_user(info))
         org_id = get_current_organization(info)
 
+        photo_dicts = [strawberry.asdict(p) for p in data.photos]
         photos = shelter_photo.resolve_uploads(
             user=user,
             organization_id=org_id,
             shelter_id=data.shelter_id,
-            photos=data.photos,
+            photos=photo_dicts,
```
vecchp: Does make sense we are doing this pattern we lose type safety.

apps/betterangels-backend/shelters/services/shelter_photo.py
```diff
@@ -2,122 +2,69 @@
...
+)
 
 
 def create_presigned_uploads(
     *,
     user: User,
     organization_id: str,
     shelter_id: int | str,
-    uploads: Iterable[ShelterPhotoUploadItemInput],
+    uploads: Iterable[dict],
```
vecchp: dict seems like a bad idea here

apps/betterangels-backend/notes/services.py
```diff
@@ -296,3 +306,66 @@ def note_create(
...
+    service_name="note_attachment",
+    allowed_content_types=DEFAULT_DOCUMENT_CONTENT_TYPES | DEFAULT_IMAGE_CONTENT_TYPES,
+    max_file_size=settings.NOTE_ATTACHMENT_MAX_FILE_SIZE,
+)
+
+
+def create_note_attachment_presigned_uploads(
+    *,
+    user: User,
+    uploads: Iterable[dict],
```
vecchp: preferably not a dict here please.

apps/betterangels-backend/clients/services/client_document.py
```diff
@@ -3,132 +3,69 @@
...
-                "url": item["url"],
-                "fields": item["fields"],
-                "presigned_key": item["key"],
-                "upload_token": upload_token,
-            }
-        )
-
-    return AuthorizedPresignedUploadBatch(uploads=authorized_uploads)
+    """Generate presigned S3 URLs and upload tokens for client documents (Phase 1)."""
+    return generic_create_presigned_uploads(
```
vecchp: Do we really want to call these services generic? is there a better name?

apps/betterangels-backend/clients/services/client_document.py
```diff
@@ -3,132 +3,69 @@
...
-                permission_group.group,
-                attachment,
-                [
-                    Attachment.perms.DELETE,
-                    Attachment.perms.CHANGE,
-                ],
-            )
-
-            attached.append(attachment)
+    attached = generic_resolve_attachments(
```
vecchp: Do we really want to call these services generic? is there a better name?

apps/betterangels-backend/common/services/file_upload.py
```diff
@@ -0,0 +1,248 @@
...
+            }
+        )
+
+    return AuthorizedPresignedUploadBatch(uploads=authorized_uploads)
+
+
+# ── Phase 3: Validate batch ──────────────────────────────────────────────────
+
+
+def validate_upload_batch(
```
vecchp, line 132: phase 1 to phase 3?

apps/betterangels-backend/common/services/s3.py
```diff
@@ -9,7 +9,7 @@
 
 # generate_presigned_post
 DEFAULT_UPLOAD_EXPIRATION_SECONDS = 300
-DEFAULT_MAX_FILE_SIZE = 10_000_000
+DEFAULT_MAX_FILE_SIZE = 50_000_000  # 50 MB
```
vecchp: should be env maybe?

apps/betterangels-backend/common/services/s3.py
```diff
@@ -9,7 +9,7 @@
 
 # generate_presigned_post
 DEFAULT_UPLOAD_EXPIRATION_SECONDS = 300
```
vecchp: should be env maybe?

apps/betterangels-backend/shelters/services/shelter_photo.py
```diff
@@ -1,123 +1,112 @@
...
-        ):
-            raise ValueError(f"Invalid or expired upload token for '{photo.filename}'")
-
-        if not s3_key_exists(key=photo.presigned_key):
-            raise ValueError(f"File not found in storage for '{photo.filename}'")
+    Accepts typed ``ShelterPhotoResolveItem`` instances so that
+    ``photo_type`` arrives as a ``ShelterPhotoTypeChoices`` enum member
+    — no string parsing, no ``KeyError`` possible.
+    """
+    photo_list = list(photos)
```
vecchp, line 90: why?

apps/betterangels-backend/betterangels_backend/settings.py
```diff
@@ -78,18 +78,38 @@
     IMGPROXY_PATH_PREFIX=(str, ""),
     IMGPROXY_LOCAL_URL=(str, "http://localhost:8080"),
     IMGPROXY_LOCAL_MEDIA_URL=(str, "http://better-angels:8000/media/"),
+    SHELTER_PHOTO_MAX_FILE_SIZE=(int, 50 * 1024 * 1024),  # 50 MiB
+    NOTE_ATTACHMENT_MAX_FILE_SIZE=(int, 50 * 1024 * 1024),
+    CLIENT_DOCUMENT_MAX_FILE_SIZE=(int, 50 * 1024 * 1024),
```
tglaz, line 83: can probably also add these to `ba-platform - libs/uploads/constants.s`
or whatever location works

one of the bigger risks is that FE and BE size checks don't align - and then it's kind of a weird/silent error, cause users don't know or care about file size

apps/betterangels-backend/clients/tests/utils.py
```diff
@@ -619,6 +620,33 @@ def _resolve_client_profile_photo_upload_fixture(
             },
         )
 
+    def _delete_client_profile_photo_fixture(
```
tglaz, line 623: ...out of scope, especially as we do this everywhere
but think we're overloading the term `fixture` a bit - think it's typically related to static/reusable setup data

for a method/factory I'd just call it `_delete_client_profile_photo`

apps/betterangels-backend/clients/schema.py
```diff
@@ -664,20 +689,17 @@ def generate_client_document_uploads(
             [ClientProfile.perms.CHANGE],
         ).get(id=data.client_profile_id)
 
-        presigned_uploads = client_document.create_presigned_uploads(user=user, uploads=data.uploads)
+        uploads = [
+            GenerateUploadItem(
```
tglaz: minor, but as I read it, `GenerateUploadItem` reads as a function call
maybe `UploadGenerationItem` or `PresignedUploadRequest`.....

the convention of Verb+Noun for a type/class can get confusing
counter-example: `ShelterPhotoResolveItem`

apps/betterangels-backend/clients/schema.py
```diff
@@ -691,13 +713,23 @@ def resolve_client_document_uploads(
             [ClientProfile.perms.CHANGE],
         ).get(id=data.client_profile_id)
 
-        documents = client_document.resolve_upload(
+        documents = [
+            ResolveUploadItem(
```
tglaz: same comment as for `GenerateUploadItem` above
it just read like it's resolving an upload here

apps/betterangels-backend/shelters/services/shelter_photo.py
```diff
@@ -1,123 +1,113 @@
...
+            upload_token=self.upload_token,
+            filename=self.filename,
+            mime_type=self.mime_type,
+        )
 
 
-def _validate_content_type(content_type: str, filename: str) -> None:
-    if content_type not in ALLOWED_CONTENT_TYPES:
-        raise ValueError(f"Unsupported content_type: {content_type} for filename={filename}.")
+SHELTER_PHOTO_CONFIG = AttachmentUploadConfig(
```
tglaz: can probably move it to top of file

apps/betterangels-backend/common/services/types.py
```diff
@@ -1,13 +1,65 @@
...
 
 
-class AuthorizedPresignedUploadBatch(TypedDict):
+@dataclass(frozen=True)
+class AuthorizedPresignedUploadBatch:
     uploads: list[AuthorizedPresignedUpload]
+
+
+@dataclass(frozen=True)
+class AttachmentUploadConfig:
```
tglaz, line 19: think at some point (def not now) we may want to make each svc a package, thought that may be just me
and each package would own its types

apps/betterangels-backend/common/tests/test_file_upload_service.py
```diff
@@ -0,0 +1,453 @@
...
+    max_file_size=settings.S3_DEFAULT_PRESIGNED_MAX_FILE_SIZE,
+)
+
+
+# ---------------------------------------------------------------------------
+# _validate_content_type
+# ---------------------------------------------------------------------------
+
+
+class ValidateContentTypeTest(TestCase):
```
tglaz, line 52: I asked our chinese AI friend about some scenarios that we could possibly add.
this is the core of the entire PR, so I'd probably review and see which are appropriate
- I'd assume some may be overkill
- but these are tests for the common service, so if you're gonna overtest anything, this is it

**ai feedback** 

`validate_upload_batch` is a public function that domain services (e.g. `shelter_photo.py`) call directly, but it has no dedicated tests. Currently it's only exercised indirectly through `create_attachment_records`.

Cases worth adding:

- **validate_upload_batch** — empty uploads returns empty list (not raises)
- **validate_upload_batch** — item with empty/`None` mime_type skips content-type check but still validates token + S3 existence (the `if item.mime_type:` branch is untested)
- **validate_upload_batch** — content-type error fires before token validation runs (error precedence — useful for debugging UX)
- **validate_upload_batch** — token validation failure fires before S3 key check
- **validate_upload_batch** — fails on first bad item in a multi-item batch, never reaches later items (fail-fast ordering)

Lower-priority edge cases that could break in a refactor:

- **create_presigned_uploads** — S3 returns fewer uploads than requested (count mismatch — `zip()` in `resolve_uploads` would silently truncate)
- **create_attachment_records** — `content_object` is an unsaved model instance (no pk yet — would raise late with a confusing error)
- **create_attachment_records** — `strip_storage_location` receives a key without the expected prefix (e.g. storage backend migration changes the prefix)
- **create_attachment_records** — duplicate presigned keys in the same batch (two items resolving the same S3 object)

apps/betterangels-backend/common/models.py
```diff
@@ -89,11 +89,28 @@ def save(
...
+        .. deprecated::
+            This custom ``save()`` override is deprecated. MIME-type detection,
+            attachment type inference, and filename canonicalisation will move
+            to the service layer (see ``create_attachment_records`` in
+            ``common/services/attachment_upload.py``).  The presigned S3 upload
+            pipeline already enforces content types at the S3 level and passes
+            ``mime_type`` explicitly via ``direct_upload=True``.  This override
+            will be removed in a future release — ``Attachment`` will revert to
+            plain Django ``save()``.
         """
```
tglaz, line 103: up to you to reconcile (not sure):

```
.. deprecated:
    The MIME-type *detection* path (reading file bytes via ``magic``) is
    deprecated — the presigned S3 upload pipeline sets ``mime_type``
    explicitly via ``direct_upload=True``.  Attachment-type inference and
    filename canonicalisation still happen here and must be moved to
    ``create_attachment_records`` before this override can be removed.
```

apps/betterangels-backend/shelters/schema.py
```diff
@@ -231,25 +230,22 @@ def generate_shelter_photo_uploads(
         user = cast(User, get_current_user(info))
         org_id = get_current_organization(info)
 
-        presigned_uploads = shelter_photo.create_presigned_uploads(
+        uploads = [
+            GenerateUploadItem(
```
tglaz: the `ShelterPhotoResolveItem` used below seems better - does not read like an executable function

libs/react/shelter-operator/src/lib/components/ShelterProfile/segments/Media/Photos/components/ShelterImagesUpload/useShelterPhotoUpload/constants.ts
```diff
@@ -7,7 +7,7 @@ import type {
 // Must be <= the server-side max_file_size in the S3 presigned POST policy.
 // If this exceeds the server limit, the client check will pass but S3 will
 // reject the upload with EntityTooLarge, showing a confusing error to users.
-export const SHELTER_PHOTO_MAX_SIZE = 10_000_000;
+export const SHELTER_PHOTO_MAX_SIZE = 50 * 1024 * 1024;  // 50 MiB
```
tglaz: since it's same as default, 2 options:
1. define `SHELTER_PHOTO_MAX_SIZE` to reference the default
  * this would just sets up the pattern for someone to override, if they want
2. don't define it at all (rely on default)

tglaz (changes requested):

apps/betterangels-backend/common/services/file_upload.py
```diff
@@ -0,0 +1,182 @@
+"""
+Generic presigned S3 upload service.
```
tglaz, line 2: i'd probably prefer the name `file_upload.py`

* `common/services/file_upload.py`
   * I'd assume (correctly) that it handles all file uploads
   * the fact that it's related to `class Attachment(BaseModel)` is an implementation detail
* `common/services/attachment_upload.py`
   * makes me ask: what is an "attachment"?

apps/betterangels-backend/shelters/services/shelter_photo.py
tglaz: 🤷

tglaz (approved):

PR #2218, opened 2026-07-08
operator portal refactor 1 of 3 beds (SDB-264)

mikefeldberg:
SDB-264

PR 1 of 3

This PR updates bed operations and components and anything that touches them

full diff [here](https://github.com/BetterAngelsLA/monorepo/compare/SDB-264/op-portal-hooks?expand=1)


libs/react/shared/src/lib/utils/errors/extractOperationInfo.ts
```diff
@@ -0,0 +1,38 @@
...
+ * @param queryKey        The top-level mutation field name (e.g. `'cloneBed'`).
+ * @returns The `OperationInfo` payload, or `null` if none was found.
+ */
+export function extractOperationInfo(
+  responseOrData: unknown,
+  queryKey: string
+): {
+  __typename: 'OperationInfo';
+  messages: Array<{ kind: string; field?: string | null; message: string }>;
+} | null {
```
tglaz: should it return `OperationInfo | null`?

libs/react/shared/src/lib/utils/errors/extractOperationInfoMessage.ts
```diff
@@ -0,0 +1,24 @@
...
+  responseOrData: unknown,
+  queryKey: string
+): string | null {
+  const operationInfo = extractOperationInfo(responseOrData, queryKey);
+
+  if (!operationInfo || operationInfo.messages.length === 0) {
+    return null;
+  }
+
+  return operationInfo.messages[0].message || null;
```
tglaz: this will work, but difficult to tell if the message we show user is actually a relevant one.

FYI - this PR #2211 is in review and will be able to handle some of this work via a `getFieldErrorsOrThrow` function - can see behavior in the tests

libs/react/shelter-operator/src/lib/components/beds/bed-form/BedForm.tsx
```diff
@@ -100,53 +80,48 @@ export function BedForm({
...
-      if (!isEditMode) {
-        reset();
-      }
-      onSuccess?.();
-    } catch {
-      setSubmissionError('A network error occurred. Please try again.');
+    if (!isEditMode) {
+      reset();
     }
+    onSuccess?.();
```
tglaz: the if statement on onSuccess are outside of the try/catch block - so they'll always run
should they be inside try/catch or in a `.finally` block perhaps?

libs/react/shelter-operator/src/lib/hooks/useBeds/useBeds.ts
```diff
@@ -0,0 +1,25 @@
+import { useQuery } from '@apollo/client/react';
+import {
+  BedsDocument,
+  type BedsQuery,
+  type BedsQueryVariables,
+} from './__generated__/useBeds.generated';
+
+export type UseBedsResultType = NonNullable<BedsQuery['beds']['results']>;
```
tglaz: not major, may be simpler to just define the type for a singular item, such as `UseBedsResultIemType`, and then can just use `UseBedsResultIemType[]`
or can export both `UseBedsResultType` and `UseBedsResultIemType`

avoids doing `UseBedsResultType[number]` elsewhere, which is a bit awkward

libs/react/shelter-operator/src/lib/components/beds/BedsView.tsx
```diff
@@ -1,57 +1,78 @@
...
 const UNASSIGNED_ROOM_ID = 'unassigned-room';
 const UNASSIGNED_ROOM_LABEL = 'Unassigned';
 
+function toBedRow(
+  bedData: UseBedsResultType[number],
+  roomId: string,
+  roomAssignment: string
+): Bed {
+  return {
+    __typename: 'BedType',
```
tglaz: fyi, this #2209 should get merged soon, and I made `__typename` required in types (since it actually already is)
* that PR only covers `shelter-operator`, but will apply it across the board as a follow up

libs/react/shelter-operator/src/lib/components/beds/BedsView.tsx
```diff
@@ -1,57 +1,78 @@
...
+
+  const beds = useMemo<Bed[]>(() => {
+    const grouped = new Map<string, Bed[]>();
+
+    for (const bed of bedsData) {
+      const roomId = bed.room?.id ?? UNASSIGNED_ROOM_ID;
+      const roomAssignment = bed.room?.name ?? UNASSIGNED_ROOM_LABEL;
+      const roomGroup = grouped.get(roomId) ?? [];
+
+      roomGroup.push(toBedRow(bed, roomId, roomAssignment));
```
tglaz: any reason to not just use BedType in BedTable and do `roomGroup.push(toBedRow(bed))`?
everything seems like defaults or derived values - could possiby simplify

libs/react/shelter-operator/src/lib/components/beds/BedsView.tsx
```diff
@@ -239,22 +207,35 @@ export function BedsView({ shelterId }: { shelterId: string }) {
...
-      {actionError && (
+      {(cloneError || deleteError || updateError) && (
         <div
-          className="mx-4 mt-4 rounded-md border border-red-200 bg-red-50 px-4 py-3 text-sm text-red-800"
+          className="mx-4 mt-4 flex items-start rounded-md border border-red-200 bg-red-50 px-4 py-3 text-sm text-red-800"
           role="alert"
         >
-          {actionError}
+          <span className="flex-1">
+            {cloneError || deleteError || updateError}
```
tglaz: would we ever have more than 1 error at a time?
if not then could just store a singe errorMsg

also, it'd be a product question, but i've been using Toast messages for these - and for Errors can just make them persistent if needed.

libs/react/shelter-operator/src/lib/components/beds/BedsView.tsx
```diff
@@ -67,146 +88,93 @@ export function BedsView({ shelterId }: { shelterId: string }) {
...
         });
-
-        if (result?.deleteBeds?.__typename === 'OperationInfo') {
-          const firstMessage = result.deleteBeds.messages?.[0]?.message;
-          setActionError(
-            firstMessage || 'Unable to delete bed(s). Please try again.'
-          );
+        const errorMessage = extractOperationInfoMessage(
+          response,
+          'deleteBeds'
```
tglaz: once #2209 merges (it's in queue), you'll be able to pull `deleteBeds` from generated file (generates a `bedMutations_meta.generated.ts` file) with

```typescript
export const deleteBedsOperationKey: keyof Omit<DeleteBedsMutation, '__typename'> = 'deleteBeds';
export const deleteBedsSuccessTypename: Extract<
  NonNullable<DeleteBedsMutation['deleteBeds']>,
  { __typename: 'BulkDeleteResult' }
>['__typename'] = 'BulkDeleteResult';

export const deleteBedsMeta = {
  operationKey: deleteBedsOperationKey,
  successTypename: deleteBedsSuccessTypename,
} as const;
```

libs/react/shelter-operator/src/lib/components/beds/BedsView.tsx
```diff
@@ -67,146 +88,93 @@ export function BedsView({ shelterId }: { shelterId: string }) {
...
       try {
-        const { data: result } = await updateBed({
+        const response = await updateBed({
           variables: {
-            id: rowObject.bedId,
-            data: {
-              lastCleaned: new Date().toISOString(),
-            },
+            id: rowObject.id,
+            data: { lastCleaned: new Date().toISOString() },
```
tglaz, line 134: is `lastCleaned` meant to reflect the bed actually being cleaned?
if so, seems like "ready" and "cleaned date" may need to be separate concerns - though perhaps for later

libs/react/shelter-operator/src/lib/components/beds/BedsView.tsx
```diff
@@ -270,7 +251,7 @@ export function BedsView({ shelterId }: { shelterId: string }) {
           label: 'Delete',
           onClick: () => {
             if (deleteConfirmation.bedIds.length > 0) {
-              handleDeleteBeds(deleteConfirmation.bedIds);
+              handleDelete(deleteConfirmation.bedIds);
```
tglaz: if this is false: ```deleteConfirmation.bedIds.length > 0```
then probably ConfirmationModal should never be rendered?

libs/react/shelter-operator/src/lib/components/BedTable.tsx
```diff
@@ -138,102 +96,30 @@ export function BedTable({
...
       {
         key: 'bedId',
         label: 'Bed',
         width: '1.1fr',
         cellClassName:
           'font-medium text-gray-900 overflow-hidden text-ellipsis whitespace-nowrap',
-        render: ({ bed }) => bed.name || bed.id,
-        sortValue: ({ bed }) => bed.name || bed.id,
-        filterValue: ({ bed }) => bed.name || bed.id,
+        render: (bed) => bed.name ?? bed.id,
```
tglaz: minor, but if bad has no name, then perhaps render ${bed #${bed.id}} or ${bed id-${bed.id}}?

libs/react/shelter-operator/src/lib/components/BedTable.tsx
```diff
@@ -138,102 +96,30 @@ export function BedTable({
...
+        sortValue: (bed) => bed.name ?? bed.id,
```
tglaz: you'll get some unexpected results if these can be mixed

libs/react/shelter-operator/src/lib/hooks/useCloneBed/useCloneBed.ts
```diff
@@ -0,0 +1,40 @@
...
+  const [cloneBed, { loading, error }] = useMutation<
+    CloneBedMutation,
+    CloneBedMutationVariables
+  >(CloneBedDocument, {
+    errorPolicy: 'all',
+    refetchQueries: refetch
+      ? (result) => {
+          const payload = result.data?.cloneBed;
+
+          if (payload?.__typename === 'BedType') {
```
tglaz: once #2209 merged (it's in queue) - then i'd suggest running codegen and use the generated key for `BedType` to be safer

we had any issue where typename changed and refetch failed because of magic string

tglaz (changes requested):

libs/react/shelter-operator/src/lib/components/beds/BedsView.tsx
tglaz: the reason I use Toast is 2-fold:
* we should have a single mechanism to deliver messages
* A Toast has position: fixed, so even if you scrolled down, it'll be visible
  * the `error` container at top of a form may not be visible after I click `submit` as it may be outside the viewport (long form - I scrolled down)

libs/react/shared/src/lib/utils/errors/index.ts
```diff
@@ -1,2 +1,4 @@
+export { extractOperationInfo } from './extractOperationInfo';
+export { extractOperationInfoMessage } from './extractOperationInfoMessage';
```
tglaz: these seem unused now?

tglaz (approved):

PR #2219, opened 2026-07-08
fix: ensure task count stays fresh and mutations update cache (BACS-93)

vecchp:
## Problem

Tasks were not appearing in the profile summary (ClientSummaryGeneral) even when they existed. Two root causes:

1. **Stale cache**: The `ClientSummaryGeneral` count query (`limit: 0`) shares the same Apollo cache key as the paginated `TasksTab` list query — because the cache key uses `['filters', 'ordering']` but excludes `pagination`. If the TasksTab loaded first, its cached data (with a non-zero limit) would be returned for the count query, potentially showing stale or missing counts.

2. **Missing cache updates**: The non-HMIS `createTask` and `updateTask` mutations didn't update the Apollo cache, so newly created/edited tasks wouldn't appear until the user navigated away and back.

## Changes

| File | Change |
|---|---|
| `ClientSummaryGeneral.tsx` | `fetchPolicy: 'network-only'` + `nextFetchPolicy: 'cache-first'` replaces `notifyOnNetworkStatusChange` |
| `TaskHeader.tsx` | Use `cache.modify()` for `updateTask` (surgical, zero network cost) + same non-null fix |

## Strategy

- **createTask** → `refetchQueries` (entity doesn't exist in cache yet; matches existing HMIS pattern)
- **updateTask** → `cache.modify()` (surgical update; all screens see changes instantly)
- **deleteTask** → already uses `cache.evict()` (no change needed)
- **count query** → `network-only` (always hits server for fresh `totalCount`)

Closes BACS-93


vecchp:
Closing — reopened as #2220 from a clean main-based branch

PR #2221, opened 2026-07-09
[BP] feat: add bed_status_metric selector with tests

Angus-Sun:
## Summary

Adds `bed_status_metric` — a reporting selector that returns daily bed status counts (available / occupied / reserved / out_of_service) for a shelter over a date range.

## Approach

Uses reservation events as the SSOT for bed status, with bed events for lifecycle tracking (create / delete / maintenance flag). For each call:

1. Fetch all `BedEvent`s for the shelter up to `end_date` to discover which beds ever existed
2. Optionally intersect with current `Bed.demographics` for demographic filtering
3. Fetch all `ReservationEvent`s for those beds
4. Merge both event streams chronologically and replay them day by day, snapshotting end-of-day state

When a bed's status changes multiple times in one day, only the final state at end-of-day is counted — this falls out naturally from chronological replay.

## Notes

- Demographic filtering uses current `Bed.demographics` only — pghistory does not track M2M through-tables, so historical demographic membership cannot be reconstructed
- All beds are assumed to have been created after pghistory tracking was installed

## Testing

- 38 unit tests covering status mapping, multi-day ranges, cumulative state, bed lifecycle, maintenance flag toggling, demographic filtering, boundary conditions, and year-long ranges
- Manually verified with dummy data in the Django shell

## Summary by Sourcery

Add a reporting selector that returns per-day bed status counts over a date range, including demographic filtering and support for long ranges.

New Features:
- Introduce bed_status_metric selector that reports daily available, occupied, reserved, and out-of-service bed counts for a shelter.
- Expose bed_status_metric through the shelters.selectors package for use by callers.

Tests:
- Add comprehensive unit tests for bed_status_metric covering status mapping, multi-day ranges, event boundaries, bed lifecycle, maintenance flag behavior, demographic filtering, and year-long ranges.


vecchp:
**Closing this PR — the functionality already exists in `main`.**

`report_bed_status_counts` (merged in #2113) already provides daily bed status counts — available, occupied, reserved, out_of_service — plus `in_turnaround`, using the same event-stream reconstruction approach against `BedEvent` and `ReservationEvent`. It also has a more performant implementation (cursor-based DB lookups rather than materializing every event into Python for replay).

Rather than maintain two parallel implementations, any future GraphQL resolver should call `report_bed_status_counts` and map its `DailyBedCounts` output to `DailyBedStatusMetricsType` in a trivial one-liner per day. The CSV export infrastructure for `DailyBedStatusMetricsType` is already in place and unchanged.

PR #2208, opened 2026-07-01
chore: upgrade Nx to 23.1.0-beta.5 and Expo 56 compatibility

vecchp:
## Summary

Upgrades Nx from 22.6.5 to 23.x and fixes Expo SDK 56 compatibility.

### @expo/cli resolution fix (Expo 56)
- Upgraded @nx/expo to 23.1.0-beta.5 which includes [nrwl/nx#35904](https://github.com/nrwl/nx/pull/35904)
- Expo 56's @expo/cli changed exports field, breaking require.resolve deep imports

### Version bumps
- @nx/*: 22.6.5 → 23.0.1 (expo → 23.1.0-beta.5)
- @nx-tools/*: 7.2.1 → 7.3.0
- @config-plugins/react-native-blob-util: 9.x → 14.x
- @config-plugins/react-native-pdf: 9.x → 14.x
- react-native-blob-util: 0.21.2 → 0.24.9
- vite: 7.x → 8.x, @vitejs/plugin-react: 4.x → 6.x

### Migrations
- vitest.workspace.ts → vitest.config.ts (Vitest 4 API)
- Nx 23 migrations: .gitignore, .prettierignore, metro config
- app.config.js: require → import (consistent ESM)

## Testing
- yarn ba starts successfully

## Summary by Sourcery

Upgrade Nx and related tooling to the 23.x line and align the React Native/Expo stack and test configuration with the latest ecosystem changes.

Bug Fixes:
- Update Nx Expo plugin and Metro integration to restore compatibility with Expo SDK 56 and its updated @expo/cli exports.

Enhancements:
- Switch betterangels app.config.js from CommonJS require to ESM import for dotenv to align with modern module usage.
- Replace the Vitest workspace configuration with a root vitest.config.ts using the Vitest 4 project discovery API.

Build:
- Bump Nx core and plugin packages to 23.x, including @nx/expo 23.1.0-beta.5 for updated Expo CLI compatibility.
- Upgrade @nx-tools container-related packages and @swc/cli to newer versions.
- Update Vite and @vitejs/plugin-react to the latest major versions and refresh related dev tooling versions in package.json.
- Add @expo/metro and remove direct metro-config/metro-resolver dependencies to match the newer Expo/Metro setup.

Tests:
- Introduce a centralized Vitest configuration file (vitest.config.ts) that discovers tests from Vite and Vitest config projects.

PR #2224, opened 2026-07-09
fix: replace @nx/expo:build with eas build command, upgrade to Expo SDK 57 and Nx 23.1.0

vecchp:
## What changed

- **Upgraded to Expo SDK 57** — React Native 0.86, no breaking changes
- **Upgraded to Nx 23.1.0** (stable, from 23.1.0-rc.1)
- **Replaced `@nx/expo:build` executor** with `command: "eas build"` target (deprecated in Nx 23.1)
- **Added `eas-build-pre-install` hook** — replicates the executor's `copyPackageJsonAndLock` behavior: copies root deps + lockfile to the app before EAS installs
- **Removed `eas-build-post-install`** — no longer needed
- **Simplified `metro.config.js`** — removed `withNxMetro` (SDK 52+ auto-configures Metro)
- **Added `expo-splash-screen`** to app deps (native-only, missed by `sync-deps`)
- **Fixed `--no-wait` flag** in `eas-deploy.ts` (was `--wait false`)
- **Installed `eas-cli` globally** in Dockerfile

## Why

`@nx/expo:build` is deprecated in Nx 23.1. The pre-install hook mirrors what the executor did: reads root `package.json` (single source of truth), replaces `*` with concrete versions, and copies the lockfile — so EAS installs from a consistent state.

## Key files

| File | Change |
|---|---|
| `tools/scripts/eas-build-pre-install.mjs` | New — mirrors `copyPackageJsonAndLock` |
| `apps/betterangels/package.json` | `*` deps, `eas-build-pre-install` script, splash-screen |
| `apps/betterangels/project.json` | `eas build` command target |
| `apps/betterangels/metro.config.js` | Simplified |
| `package.json` (root) | SDK 57, Nx 23.1.0, added missing deps |
| `yarn.lock` | SDK 57 + Nx 23.1.0 lockfile |
| `Dockerfile` | `eas-cli` globally |
| `tools/scripts/eas-deploy.ts` | `--no-wait` fix


tglaz:
I guess my main question is whether we should do this now? Or is it too late to ask lol

"Nx 23.1.0-rc.1 deprecates @nx/expo executors"
- It's an "rc.1" - so just curious about the stability, or why we upgrade this early.
-- Typically these types of core changes are stable in an rc so no real objections, but figured I'd check.  

Though now I see we were on beta earlier, so nvm I suppose. Think at first just was curious what triggered us updating (i'm assuming there was a real reason, but didn't know what)

tools/scripts/eas-resolve-star-deps.mjs
```diff
@@ -0,0 +1,23 @@
...
+const rootPkgPath = resolve(cwd, '../../package.json');
+const appPkgPath = resolve(cwd, 'package.json');
+
+const root = JSON.parse(readFileSync(rootPkgPath, 'utf-8'));
+const app = JSON.parse(readFileSync(appPkgPath, 'utf-8'));
+
+app.dependencies = root.dependencies || {};
+app.devDependencies = root.devDependencies || {};
+
+writeFileSync(appPkgPath, JSON.stringify(app, null, 2) + '\n');
```
tglaz: There are dependencies in `apps/betterangels/package.json` that I don't think I see in the root package.json, such as `apollo-link-rest` and perhaps others. 

This `eas-resolve-star-deps` script seems to overwrite appPkgPath with dependencies copied strictly from the root package.json. Is this an issue?

vecchp: Good questions. On the Nx front, the trigger for the update was the Expo 56 upgrade. To get to 56, we had to use a newer version of Nx (this came out of PR https://github.com/BetterAngelsLA/monorepo/pull/2208 and upstream Nx PR https://github.com/nrwl/nx/pull/35904).

As for the dependencies, the pushed fix addresses that exact concern. `apollo-link-rest` is actually an old dependency we no longer use, but it seems `sync-dep` doesn't automatically prune unused dependencies in certain cases. The update ensures the script only resolves dependencies strictly in the app's package.json and will explicitly flag any unresolved or unmatched dependencies.

mikefeldberg (approved):

PR #2227, opened 2026-07-14
[BP] Unify report date-range timezone handling

StockerMC:
## Description

Adds `report_date_range_to_utc`, one shared helper that converts an inclusive LA-local date range into a half-open `[start, end)` UTC window, so every date-range report selector defines day boundaries the same way. Timezone handling was fragmented: each selector rolled its own conversion and they disagreed (UTC-midnight vs LA-midnight).

## What changed

- `report_date_range_to_utc(start_date, end_date)` in `shelters/selectors/reports.py`: interprets the range as America/Los_Angeles calendar days and converts to UTC (where `pgh_created_at` is stored), inclusive of `end_date`, capped at one year.
- `reservation_status_change_counts` now goes through it, moving its day boundaries from UTC-midnight to LA-local. All BA shelters are in LA and the reporting UI sends calendar dates, so an evening LA event now lands on the correct day instead of rolling into the next.
- Unit tests for the helper; the reservation range-boundary test updated for LA-local semantics.

## Why LA-local

These are per-day / date-range metrics driven by a calendar date picker. Bucketing by UTC splits the day at ~4-5pm LA, so an evening check-in gets attributed to the next day. See the discussion on #2113 / #2114.

## Not in scope

- `report_bed_status_counts` is left untouched. It still reads the removed `BedEvent.status` column (dropped in migration 0008) so it does not run today; it belongs to the separate "daily bed status" rewrite.
- No GraphQL wiring, no migrations.

## Follow-ups that build on this

- The occupancy selectors (#2113 daily_occupancy, #2114 avg_days_to_occupancy) will stack on this branch and consume `report_date_range_to_utc` instead of carrying their own copies.

## Summary by Sourcery

Unify report date-range handling around a shared LA-local-to-UTC conversion helper and apply it to reservation status change reporting.

Enhancements:
- Introduce a shared helper to convert inclusive LA-local calendar date ranges into half-open UTC windows with validation and a one-year cap.
- Update reservation status change reporting to use the shared LA-local date range conversion so events are bucketed by local calendar days consistently.

Tests:
- Add unit tests for the shared date range conversion helper and adjust reservation status change range-boundary tests for LA-local semantics.

### DIFF
--- a/apps/betterangels-backend/shelters/selectors/reports.py
+++ b/apps/betterangels-backend/shelters/selectors/reports.py
@@ -4,6 +4,7 @@ import datetime
 from collections import Counter
 from itertools import takewhile
 from typing import TYPE_CHECKING, Any, TypedDict
+from zoneinfo import ZoneInfo
 
 import pghistory
 from django.db.models import Count, OuterRef, Q, Subquery, TextField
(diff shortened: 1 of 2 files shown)


StockerMC:
Regarding Sourcery comments:

Regarding the constants, the shared entry point is `report_date_range_to_utc` itself, and every selector goes through it rather than touching `LA_TZ`/`MAX_REPORT_RANGE_DAYS`, so there's no duplication to centralize. America/Los_Angeles is a domain fact, not env config, so settings would be the wrong home; if we ever go multi-region the fix is a per-shelter timezone field, not a global constant.

Regarding migrating remaining selectors, `report_bed_status_counts` is dead (reads the removed `BedEvent.status`) and #2221 replaces it with `bed_status_metric`, so I didn't annotate it. Added a line at the helper telling new selectors to use it instead.

apps/betterangels-backend/shelters/selectors/reports.py
```diff
@@ -15,6 +16,39 @@
...
+    Los Angeles, so the returned half-open interval covers all of ``end_date``
+    locally.
+
+    Raises:
+        ValueError: if ``end_date`` is before ``start_date`` or the inclusive
+            span exceeds :data:`MAX_REPORT_RANGE_DAYS`.
+    """
+    if end_date < start_date:
+        raise ValueError("end_date must be on or after start_date")
+    if (end_date - start_date).days + 1 > MAX_REPORT_RANGE_DAYS:
```
frank731: I don't think theres gonna be a cap on this we can prob remove

frank731 (changes requested):

frank731 (approved):

PR #2235, opened 2026-07-14
track select action in AddressAutocomplete for shelter profile (DEV-2462)

tglaz:
Map selection requires tapping address type
DEV-2462

## Summary by Sourcery

Enhancements:
- Add internal tracking in AddressAutocomplete to distinguish user typing from selection actions, avoiding redundant prediction requests after selecting an address.


shiozakiyuko (commented):
Hi @tglaz 

I tested it but I still had to click the address twice and the search result becomes zero even though I was just testing the address that I see in the default shelter list. 
Please take a look at the video if you need to.

https://github.com/user-attachments/assets/4c96c165-dc91-4b04-8512-3b7bc15ad629

Davit-BetterAngels (approved):

shiozakiyuko (approved):
Hi @tglaz 

Thanks for letting me know where I needed to look at.
Now, I checked on the SOS side, and now the address search works as expected. 
Will approve now. :)

PR #2239, opened 2026-07-15
shelter address autocomplete (DEV-2468)

tglaz:
### [Shelter Directory] Address Autocomplete
DEV-2468

* consolidate react `AddressAutocomplete` into single component
* consolidate to use single `usePlacesClient`
  * here I kept it as a local hook in AddressAutocomplete (need to reorg libs before giving it its own place and didn't want to blow up scope)


shiozakiyuko (commented):
Hi @tglaz 

Let me know if I'm understanding incorrectly.

I noticed two things.

1. So at the initial entering, I did confirm the address was visible only once. But when I go back to the search page, the address showed up below the search bar. I don't think I should see it twice when I go back to the page but let me know. 
2. The searched address didn't show up in the search result. But this search functionality is working on Production so I don't have to worry...? I remember I had a similar convo about this but I don't remember why this is happening on Dev side.

https://github.com/user-attachments/assets/13b7ab86-93dc-46f7-98c4-505847c981cc

Thanks!

libs/react/components/src/lib/AddressAutocomplete/hooks/useKeyboardListNav.ts
```diff
@@ -0,0 +1,65 @@
+import { useCallback, useRef, useState } from 'react';
```
Davit-BetterAngels: useEffect,

libs/react/components/src/lib/AddressAutocomplete/hooks/useKeyboardListNav.ts
```diff
@@ -0,0 +1,65 @@
...
+export function useKeyboardListNav<T>(
+  options: TUseKeyboardListNavOptions<T>,
+): TUseKeyboardListNavReturn {
+  const { items } = options;
+
+  const onSelectRef = useRef(options.onSelect);
+  onSelectRef.current = options.onSelect;
+
+  const [activeIndex, setActiveIndex] = useState(-1);
+
```
Davit-BetterAngels, line 40: useEffect(() => {
  setActiveIndex((prev) =>
    prev >= items.length ? items.length - 1 : prev,
  );
}, [items.length]);

libs/react/components/src/lib/AddressAutocomplete/hooks/useKeyboardListNav.ts
```diff
@@ -0,0 +1,65 @@
...
+          e.preventDefault();
+          setActiveIndex((prev) => (prev < items.length - 1 ? prev + 1 : 0));
+          break;
+
+        case 'ArrowUp':
+          e.preventDefault();
+          setActiveIndex((prev) => (prev > 0 ? prev - 1 : items.length - 1));
+          break;
+
+        case 'Enter':
```
Davit-BetterAngels: case 'Enter': {
  const item = items[activeIndex];

  if (activeIndex >= 0 && item !== undefined) {
    e.preventDefault();
    onSelectRef.current(item);
  }

  break;
}

Davit-BetterAngels (changes requested):

libs/react/components/src/lib/AddressAutocomplete/hooks/useKeyboardListNav.ts
tglaz: the current version handled this imperatively via `reset` - which the caller would need to implement.
but yeah, safer to do it via `useEffect` here.

i did it with just an `[items]` dep since you can have items change but the number of items remain the same

shiozakiyuko (approved):

Davit-BetterAngels (approved):

PR #2251, opened 2026-07-20
chore: move shelter management to sidebar (beds) (SDB-256)

mikefeldberg:
SDB-256

## Summary by Sourcery

Move bed management from the shelter management dashboard to sidebar with dedicated routing, navigation, and pages.

New Features:
- Introduce a dedicated shelter management route hierarchy, including beds listing and editing pages.
- Add a Shelter Management section to the sidebar with navigation to the beds view.

Enhancements:
- Refactor bed-related components and form logic into a ShelterManagement module with consolidated form schema and mapping utilities.


libs/react/shelter-operator/src/lib/components/ShelterOperations/segments/Beds/BedForm/sections/BedDetails.tsx
```diff
@@ -1,16 +1,13 @@
...
-} from '../constants/bedFormOptions';
+import { CheckboxGroup } from '../../../../../form/CheckboxGroup';
+import { FormSection } from '../../../../../form/FormSection';
+import { NumberField } from '../../../../../form/NumberField';
+import { RadioGroup } from '../../../../../form/RadioGroup';
+import { BOOLEAN_OPTIONS, MEDICAL_NEED_OPTIONS } from '../formSchema';
 import type { SectionProps } from '../types';
 
-export const BedDetailsSection = memo(function BedDetailsSection({
+export const BedDetails = memo(function BedDetails({
```
tglaz: don't think the `memo` pattern is necessary here. 
the props will change on most interactions which would re-render anyway - and it's a simple enough page that should be fine

libs/react/shelter-operator/src/lib/routing/routePaths.ts
```diff
@@ -33,13 +34,16 @@ export const shelterProfileSegments = {
   media: 'media',
 } as const;
 
+export const shelterOperationsSegments = {
+  beds: 'beds',
+  bedsCreate: 'beds/create',
+  bedsEdit: 'beds/:bedId/edit',
+} as const;
```
tglaz: these feel more like paths than segments, and we'd probably want to have a common convention throughout the file.

mikefeldberg: what do you think of something like this?

```ts
export const shelterOpsSegments = {
  beds: 'beds',
  occupants: 'occupants',
  summary: 'summary',
  rooms: 'rooms',
  reservations: 'reservations',
} as const;

export const shelterOpsPaths = {
  create: 'create',
  edit: ':id/edit',
} as const;

export function shelterOpsEditBedPath(
  shelterId: string,
  bedId: string,
): string {
  return generatePath(
    `${shelterOperationsRoute(shelterId)}/${shelterOpsSegments.beds}/${shelterOpsPaths.edit}`,
    { id: bedId },
  );
}
```

tglaz: lemme think about it - in theory could work, but we'd probably want to define a convention for what is a "path" vs "subpath" vs "segment" etc. 
as long as it works for now it's fine I think, since these are already centralized to a single file

libs/react/shelter-operator/src/lib/components/ShelterManagement/segments/Beds/Beds.tsx
```diff
@@ -159,28 +164,28 @@ export function BedsView({ shelterId }: { shelterId: string }) {
 
   const [readyConfirmation, setReadyConfirmation] = useState<{
     isOpen: boolean;
-    rowObject: BedRowObject | null;
-  }>({ isOpen: false, rowObject: null });
+    bedId: string | null;
+  }>({ isOpen: false, bedId: null });
```
tglaz: you could simplify and just store the ID?
const [readyBedId, setReadyBedId] = useState<string | null>(null);

libs/react/shelter-operator/src/lib/components/ShelterManagement/segments/Beds/Beds.tsx
```diff
@@ -159,28 +164,28 @@ export function BedsView({ shelterId }: { shelterId: string }) {
...
-        roomId: rowObject.bed.room?.id ?? null,
+        bedId: bed.id,
+        roomId: bed.room?.id ?? null,
       };
       navigate(shelterCreateResourceRoute(shelterId, 'reservation'), { state });
     },
     [navigate, shelterId],
   );
-  const readyRowObject = readyConfirmation.rowObject;
+  const readyBed = readyConfirmation.bedId;
```
tglaz: rename to `readyBedId`?
or just use readyConfirmation.bedId directly

libs/react/shelter-operator/src/lib/components/ShelterManagement/segments/Beds/Beds.tsx
```diff
@@ -130,13 +130,13 @@ export function Beds({ shelterId }: { shelterId: string }) {
   );
 
   const handleMarkReady = useCallback(
-    async (bed: Bed) => {
+    async (bedId: string) => {
       const errorMessage = 'Unable to update bed. Please try again.';
```
tglaz: I'd suggest not defining `errorMessage` here.
notice that you throw this message, but also use same variable in `showToast` - which gets confusing.
if there's only 1 message we show user via Toast - just define it there.

`fieldErrors` from `getFieldErrorsOrThrow` are meant to be recoverable, such as a phone number input validation that passes in FE but fails on server. 
* It this doesn't apply here, then doing `throw new Error(errorMessage)` is ok, but otherwise I'd look at how we throw `BaError` instead in other forms.
* if there are recoverable form field errors, then the modal shouldn't close on such error - right now it will

tglaz (changes requested):

libs/react/shelter-operator/src/lib/components/ShelterManagement/segments/Beds/Beds.tsx
```diff
@@ -57,26 +62,24 @@ export function BedsView({ shelterId }: { shelterId: string }) {
...
+        const response = await cloneBed({ variables: { id: bedId } });
 
         const fieldErrors = getFieldErrorsOrThrow({
           response,
           ...cloneBedMeta,
           fields: ['id'],
         });
         if (fieldErrors.length) {
-          throw new Error(errorMessage);
+          throw new Error('Field validation failed');
```
tglaz, line 75: minor, but this error will get logged in console, so would help include the actual `fieldErrors`

tglaz (approved):

PR #2254, opened 2026-07-21
chore: update shelter routing

mikefeldberg:
## Summary by Sourcery

Centralize shelter management and profile routing into shared configs and generic helpers for resource routes.

Enhancements:
- Introduce management resource metadata and CRUD route configs as single source of truth for shelter manage/profile paths and parameters.
- Refactor navigation, breadcrumbs, sidebar, and dashboard components to consume the new routing configs and generic helpers instead of hardcoded segments.
- Simplify reservation, room, and bed flows by using generic management resource route helpers for create/edit/navigation.

### DIFF
--- a/libs/react/shelter-operator/src/lib/components/NavBar/breadcrumbs.ts
+++ b/libs/react/shelter-operator/src/lib/components/NavBar/breadcrumbs.ts
@@ -7,7 +7,7 @@ import {
   useRoom,
   useShelterOperatorProfile,
 } from '../../hooks';
-import { manageRouteConfig, paths, profileRouteConfig } from '../../routing';
+import { mgmtRouteConfig, paths, profileRouteConfig } from '../../routing';
 
 // ─── Types ───────────────────────────────────────────────────────────────────
 
@@ -64,18 +64,18 @@ interface RoutePattern {
 const routePatterns: RoutePattern[] = [
   // ── Reservations ─────────────────────────────────────────────────────────
   {
-    pattern: `${manageRouteConfig.root}/${manageRouteConfig.children.reservations}/${manageRouteConfig.actions.edit}`,
+    pattern: `${mgmtRouteConfig.root}/${mgmtRouteConfig.children.reservations}/${mgmtRouteConfig.actions.edit}`,
     build: (params) => [
       {
         label: `__shelterId__:${params.shelterId}`,
-        path: buildPath(manageRouteConfig.root, params),
+        path: buildPath(mgmtRouteConfig.root, params),
       },
-      { label: 'Manage', path: buildPath(manageRouteConfig.root, params) },
+      { label: 'Manage', path: buildPath(mgmtRouteConfig.root, params) },
       {
         label: 'Reservations',
         path: buildPath(
-          `${manageRouteConfig.root}/${manageRouteConfig.children.reservations}`,
-          params
+          `${mgmtRouteConfig.root}/${mgmtRouteConfig.children.reservations}`,
+          params,
         ),
       },
       { label: `__reservationId__:${params.id}`, path: '#' },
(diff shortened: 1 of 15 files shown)


libs/react/shelter-operator/src/lib/components/beds/BedsView.tsx
```diff
@@ -82,14 +81,14 @@ export function BedsView({ shelterId }: { shelterId: string }) {
...
       }
     },
-    [cloneBed, showToast]
+    [cloneBed, showToast],
   );
 
   const handleEdit = useCallback(
     (rowObject: BedRowObject) => {
-      navigate(shelterEditBedRoute(shelterId, rowObject.id));
+      navigate(shelterEditResourceRoute(shelterId, 'bed', rowObject.id));
```
tglaz, line 89: nice. still have type safety

libs/react/shelter-operator/src/lib/components/NavBar/breadcrumbs.ts
```diff
@@ -7,7 +7,7 @@ import {
   useRoom,
   useShelterOperatorProfile,
 } from '../../hooks';
-import { manageSegments, paths, shelterProfileSegments } from '../../routing';
+import { manageRouteConfig, paths, profileRouteConfig } from '../../routing';
```
tglaz: not blocking, and don't have a great answer for this off top of my head.
but `manageRouteConfig` sounds like an action (a function) and not a config object

tglaz: but also, i'd seriously consider killing this file (at some point).

1. it's 400+ LOC for just a Header
- many other ways to tell the user where they're at
- do we really want to maintain this?
2. it can result in breadcrumbs like below
- super long and can be cut off
- in mobile/tablet almost unusable except for very shallow cases

Image

libs/react/shelter-operator/src/lib/pages/beds/EditBedPage.tsx
```diff
@@ -3,14 +3,14 @@ import { BedForm } from '../../components/beds/bed-form/BedForm';
...
-import { shelterManageBedsRoute } from '../../routing';
+import { shelterManageResourceRoute } from '../../routing';
 
 export function EditBedPage() {
   const navigate = useNavigate();
   const { shelterId, bedId } = useParams();
   const { bed, loading, error } = useBed(bedId ?? '');
 
-  const bedsPath = shelterManageBedsRoute(shelterId ?? '');
+  const bedsPath = shelterManageResourceRoute(shelterId ?? '', 'bed');
```
tglaz: this was already there - so just commenting...

should shelterManageResourceRoute require a shelterId?
- it's a shelter resource, so seems like we need it

if the idea is to have this also handle the "create" case, then either:
- have a BedForm component and separate create/edit components that use it
- or some other way where it's clear how we're build the route

otherwise do we allow editing a Bed resource purely by its own Id?
- we could, but not sure we do

same comment applies to other files

libs/react/shelter-operator/src/lib/components/NavBar/breadcrumbs.ts
mikefeldberg: > manageRouteConfig sounds like an action

agreed. `managmentRouteConfig`? `mgmtRouteConfig`?

> breadcrumbs

I don't see the point at all, tbh. Breadcrumbs are useful when you're drilling down a few levels but our site is basically flat. With the exception of the shelter name, all other info displayed in the breadcrumbs is also prominently displayed _on_ the page you're on. I think this header would make a lot more sense as: `<org>: <shelter>`.

libs/react/shelter-operator/src/lib/pages/beds/EditBedPage.tsx
mikefeldberg: > have a BedForm component and separate create/edit components that use it

good idea. Will add to upcoming refactor

libs/react/shelter-operator/src/lib/components/NavBar/breadcrumbs.ts
tglaz: `mgmtRouteConfig` works - but it's one of those that not sure - but guess no need to dwell right now

tglaz (approved):

PR #2257, opened 2026-07-22
schedule filter types (SDB-262)

Davit-BetterAngels:
SDB-262

## Summary by Sourcery

Support filtering shelters by multiple schedule types in the "open now" query and propagate the new schedule-type list through backend, GraphQL schema, and frontend filters.

New Features:
- Allow the open-now shelter filter to accept a list of schedule types and treat openness as the union of those schedule schedules.

Enhancements:
- Update shelter open-at query logic to handle multiple schedule types with per-type exception handling and support a no-op when the list is empty.
- Simplify shelter property filter handling on the frontend by removing the legacy boolean openNow flag and relying solely on openNowScheduleTypes.

Tests:
- Extend shelter open-now tests to cover multiple schedule types, per-type exceptions, and the empty-list no-op behavior, and adjust existing tests to the new API shape.


apps/betterangels-backend/shelters/open_at.py
```diff
@@ -87,32 +92,45 @@ def shelters_open_at(
...
-            shelter=OuterRef("pk"),
-            schedule_type=schedule_type,
-            is_exception=False,
-        ).filter(
-            covers,
-            Q(start_date=None) | Q(start_date__lte=date),
-            Q(end_date=None) | Q(end_date__gte=date),
+    # Deduplicate schedule_types while preserving order.  An empty iterable
+    # means "no filter" and returns the queryset unchanged.
+    unique_types: Sequence[ScheduleTypeChoices] = list(dict.fromkeys(schedule_types))
```
mikefeldberg: this could be `set(schedule_types)`, since order doesn't matter. Is it possible that `schedule_types` would contain duplicates?

apps/betterangels-backend/shelters/open_at.py
```diff
@@ -17,16 +18,20 @@ def shelters_open_at(
     queryset: "QuerySet[Shelter]",
     *,
     dt: datetime.datetime,
-    schedule_type: ScheduleTypeChoices = ScheduleTypeChoices.OPERATING,
+    schedule_types: Iterable[ScheduleTypeChoices] = (ScheduleTypeChoices.OPERATING,),
```
mikefeldberg: I don't think it makes sense to have a default value here

mikefeldberg (approved):

PR #2271, opened 2026-07-24
Fetch reporting metrics from the Reports overview

Anahat07:
- Add `shelterOccupancyMetrics` GraphQL query that assembles occupancy, bed status, reservation, and avg-days-to-occupancy metrics for a shelter + inclusive date range
- Wire the Reports overview to fetch those metrics from the current shelter and `dateRangeFilterAtom` (defaults to Last 30 Days)
- Log the returned payload in the browser console for verification


libs/react/shelter-operator/src/lib/components/reports/ReportsView.tsx
```diff
@@ -1,13 +1,43 @@
...
-export function ReportsView({ shelterId: _shelterId }: { shelterId?: string }) {
+export function ReportsView({ shelterId }: { shelterId?: string }) {
+  const { range } = useAtomValue(dateRangeFilterAtom);
+  const { metrics, loading, error } = useShelterOccupancyMetrics({
+    shelterId,
+    startDate: range.from,
+    endDate: range.to,
+  });
+
+  useEffect(() => {
```
frank731: likely can do this without a useEffect https://react.dev/learn/you-might-not-need-an-effect

frank731 (commented):
Look good! One comment about removing useEffect since you typically want to only use it due to external factors

frank731 (approved):

PR #2274, opened 2026-07-25
fix: re-apply Catalyst weak-link WidgetKit/ActivityKit flags

vecchp:
## Summary

Re-applies the `-weak_framework WidgetKit -weak_framework ActivityKit` linker flags that were originally added in #2269 and reverted in #2273.

## Why

On macOS < 14.2, when running the iOS app as "Designed for iPad" (Catalyst), the iOSSupport WidgetKit framework lacks the `activityBackgroundTint` symbol (pulled in transitively by Expo native deps). This causes a dyld crash at launch. Weak-linking tells dyld to treat missing symbols as NULL instead of aborting.

## What happened with #2273

The weak-link flags were incorrectly blamed for an iOS Simulator dyld crash. That crash was a separate local build environment issue — the security fix (#2238) did not change any Expo package versions in `yarn.lock`.

## Changes

| File | Change |
|---|---|
| `apps/betterangels/app.config.js` | Re-apply `extraIosProps` with weak framework flags; remove `LSMinimumSystemVersion` |
| Version | `1.2.5` → `1.2.6` |
| iOS buildNumber | `1.2.5` → `1.2.6` |
| Android versionCode | `79` → `80` |

Reverts the revert in #2273. Restores fix for BACS-97.

### DIFF
--- a/apps/betterangels/app.config.js
+++ b/apps/betterangels/app.config.js
@@ -21,7 +21,7 @@ export default {
     name: IS_PRODUCTION ? 'BetterAngels' : 'BetterAngels (Dev)',
     slug: 'betterangels',
     scheme: IS_PRODUCTION ? 'betterangels' : 'betterangels-dev',
-    version: '1.2.5',
+    version: '1.2.6',
     orientation: 'portrait',
     icon: IS_PRODUCTION
       ? './src/app/assets/images/icon.png'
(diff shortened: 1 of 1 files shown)


vecchp:
Closing — the iOS simulator crash was a local environment issue (stale prebuild + node_modules/lockfile mismatch), not caused by either the LSMinimumSystemVersion or weak-link approaches. Main is working correctly.

PR #2267, opened 2026-07-23
migrate eslint to v10 (DEV-2471)

tglaz:
### migrate eslint to v10
DEV-2471

* replace `.eslintrc.json` files with `eslint.config.mjs` 
* plugin updates
  * eslint-plugin-react → `@eslint-react/eslint-plugin`
  * add `eslint-plugin-react-native`
* new warnings
  * given the new plugins, this PR surfaces new warnings, while some rules are turned off
  * we will evaluate new warnings and either ignore or fix or promote to error separately

**Apps**
- [x] nx run shelter-web:lint
- [x] nx run betterangels-admin:lint
- [x] nx run betterangels:lint

Note: - `wildfires` app to be handled separately via DEV-2478

**Libs: react**
- [x] nx run react-shared:lint
- [x] nx run react-components:lint
- [x] nx run react-icons:lint
- [x] nx run react-shelter:lint
- [x] nx run react-shelter-operator:lint
- [x] nx run react-betterangels-admin:lint

**Libs: expo**
- [x] nx run expo-betterangels:lint
- [x] nx run expo-shared-ui-components:lint
- [x] nx run expo-shared-icons:lint
- [x] nx run expo-shared-images:lint
- [x] nx run expo-shared-static:lint
- [x] nx run expo-shared-utils:lint

**Libs: shared**
- [x] nx run shared-units:lint
- [x] nx run shared-places:lint

**Libs: ba-platform**
- [x] nx run ba-platform:lint
- [x] nx run ba-platform-permissions:lint
- [x] nx run ba-platform-web:lint
- [x] nx run ba-platform-expo:lint

**Libs: other**
- [x] nx run apollo:lint
- [x] nx run fetch:lint
- [x] nx run tailwind:lint

PR #2277, opened 2026-07-27
feat: type all GraphQL scalars — Date, DateTime, UUID, Time, PhoneNumber

vecchp:
> Stacked on #2357, which adds `@monorepo/shared/scalars`. Review that one first; this PR's base is set to it, so the diff here is only the codegen change and its fallout.

## Why

Only `NonBlankString` was mapped. The other eleven scalars fell through to `any`, so a date, a UUID and a phone number were indistinguishable from each other and from any other string.

## What this does

Maps all twelve scalars in `tools/codegen/scalars.ts` and pairs the map with `strictScalars`, so a scalar the backend adds later **fails the codegen run** rather than silently becoming `any` again.

Two placement decisions worth a look:

- **The map lives in `tools/`, not in a library.** It is build tooling, so this keeps `@graphql-codegen/cli` types out of a runtime library's source. `tools/` is also not an nx project, so the relative import each `codegen.ts` needs — jiti cannot resolve tsconfig path aliases — crosses no project boundary.
- **Scalars map to bare type names, with one static `import type` header per generated file.** Inlining `import('@monorepo/shared/scalars').DateString` at every usage reads worse, and makes nx classify the library as lazy-loaded, which then rejects every *static* import of it elsewhere in the same project.

## Four defects the typing exposed

Each was invisible while the scalar was `any`, because `Date` is assignable to `any`.

1. **The client profile form was typed two ways.** All nine step forms use `useFormContext<UpdateClientProfileInput>`, where dates are `DateString`, while the parent's `types.ts` said `Date`. Form state is now `DateString` throughout, which lets `toUpdateClienProfileInputs` drop both its `Record<string, unknown>` cast and its entire date-normalization block.
2. **The date pickers rendered the wrong day.** They round-tripped through `new Date(value)`, which reads a date-only string as *UTC* midnight — a stored `1990-05-05` displayed as May 4 anywhere west of UTC. Now `fromDateString`.
3. **`ReservationTable` had the same bug** via `new Date(startDate).toLocaleDateString()`, plus two optional fields assigned to `string | null`.
4. **Note titles and cards built `Date`s from possibly-null scalars**, producing an `Invalid Date` that `format` throws a `RangeError` on. They use the `from*` helpers now, which return `undefined`, and render nothing when there is no date.

## Testing

- `nx run-many -t typecheck,lint,test` clean across all eight affected projects.
- Regeneration is idempotent: `nx run-many -t generate-graphql-types --all --skip-nx-cache` then `git diff --exit-code`.
- The guard works: deleting the `Time` mapping fails codegen with `Unknown scalar type Time`.
- `tsc -p tsconfig.lib.json --listFiles` shows 0 codegen files in the `react-shelter-operator` and `ba-platform` library builds.
- Cross-brand assignment is rejected while plain `string` → `DateString` still compiles.
- `apps/betterangels-backend` untouched.

## Summary by Sourcery

Replace implicit GraphQL scalar `any` types with strict shared mappings and migrate client code to safe, consistent scalar handling.

New Features:
- Map all GraphQL custom scalars to explicit TypeScript types, including branded date, datetime, time, UUID, and phone-number strings.
- Add shared scalar utilities for parsing, formatting, serializing, and displaying date, time, and phone-number values.

Bug Fixes:
- Prevent date-only values from shifting across time zones and avoid rendering invalid or null dates.
- Correct phone-number display and dialing behavior, including extensions and unparsable values.
- Handle optional reservation and interaction dates safely throughout the affected applications.

Enhancements:
- Centralize scalar handling in the shared scalars library and migrate affected forms, cards, notes, schedules, reports, and shelter views to use it.
- Remove duplicated date and phone-number utilities from application-specific shared libraries.
- Use branded scalar types to prevent incompatible scalar assignments while preserving compatibility with plain strings.

Build:
- Share strict GraphQL scalar configuration across code generation projects and emit the required scalar type imports.
- Enable codegen configuration files in the relevant TypeScript project configurations and allow their tooling imports through ESLint.

Tests:
- Add coverage for date, time, and phone-number parsing, formatting, serialization, invalid inputs, extensions, and timezone behavior.

Chores:
- Regenerate GraphQL operation types across the affected applications and libraries.


.eslintrc.json
```diff
@@ -12,6 +12,10 @@
           {
             "enforceBuildableLibDependency": true,
             "allow": [".*libs/tailwind.*"],
+            // NX misreads TypeScript import('module').Type expressions in
+            // generated files as JavaScript dynamic import() calls, marking
+            // ba-platform as "lazy-loaded". This suppresses the false positive.
+            "checkDynamicDependenciesExceptions": ["@monorepo/ba-platform"],
```
tglaz: FYI #2267 will replace this `.eslintrc.json` so will just need to remember to port this addition (depending on which PR goes in first)

libs/expo/betterangels/src/lib/screens/ClientProfileForms/ClientProfileForm/PersonalInfoForm/PersonalInfoForm.tsx
```diff
@@ -170,9 +171,9 @@ export function PersonalInfoForm() {
             endDate: new Date(),
             startDate: new Date('1900-01-01'),
           }}
-          value={dateOfBirth || undefined}
+          value={fromDateString(dateOfBirth)}
```
tglaz: the more I look at `fromDateString` usage, makes me think if i may ever remember this util exists.
can't really use `toDate`, as it'd be overloaded, but maybe `parseDateString` or `dateStringToDate`?
probably same would apply to `fromDateTimeString`

tglaz (approved):

libs/expo/betterangels/src/lib/screens/ClientProfileForms/ClientProfileForm/PersonalInfoForm/PersonalInfoForm.tsx
vecchp: Renamed in 8d645b73 — `parseDateString`, `parseDateTimeString`, `parseTimeString`. You were
right that it does not read as a parser: each of these already opened its docstring with
"Parse a `Date` scalar into…", so the name was the only part not saying so.

Went with `parse*` over `dateStringToDate` for two reasons. It spells out a return type the
signature already shows, and it does not survive the third case — `parseTimeString` returns
minutes since midnight, not a `Date`, so `timeStringToDate` would be wrong.

The `to*` serializers keep their names, so the pairs are now deliberately asymmetric:
`toDateString` / `parseDateString`. That is the point rather than an oversight — you no longer
have to know which side of the codec you are on to read the call site.

Eight call sites plus the library; no signature changes.

.eslintrc.json
vecchp: Confirming this one is handled — #2267 did land first, and `.eslintrc.json` no longer exists on
`main`. The addition was ported rather than lost: the codegen hatch now lives in
`eslint.config.mjs`, in the `@nx/enforce-module-boundaries` `allow` list, and it is in this PR's
diff.

```js
allow: [
  '.*libs/tailwind.*',
  'libs/vite-utils/.*',
  // Each codegen.ts imports the shared scalar map from tools/ by relative
  // path. graphql-codegen loads these configs through jiti, which does not
  // resolve tsconfig path aliases, so a scoped import fails at runtime.
  'tools/codegen/.*',
],
```

Worth noting it briefly went wrong on the way: a merge recorded `.eslintrc.json` →
`.eslintrc.json.bak` as a rename and silently carried the addition into the dead file. Restored
from `main` and re-applied to the live config.

PR #2279, opened 2026-07-28
[BP] - create json exporter for report

c4bae:
## Summary by Sourcery

Add JSON export support for shelter occupancy reporting metrics and corresponding validation tests.

New Features:
- Introduce a JSON exporter that serializes selected shelter occupancy metrics into a date-keyed report file.

Tests:
- Add tests covering JSON metric export structure, option filtering, and validation for empty or unknown export options.

### DIFF
--- a/apps/betterangels-backend/reports/export_to_json.py
+++ b/apps/betterangels-backend/reports/export_to_json.py
@@ -1,0 +1,70 @@
+"""JSON exporter for shelter reporting metrics."""
+
+import json
+
+from shelters.types.reporting import ShelterOccupancyMetricsType
+
+from .export_options import MetricsExportOptions
+
+
+def metrics_to_json(metrics: ShelterOccupancyMetricsType, options: list[MetricsExportOptions]) -> tuple[str, str]:
+    """Serialize selected shelter metrics into one date-keyed report file."""
+    selected_options = set(options)
+    invalid_options = selected_options - set(MetricsExportOptions)
+    if invalid_options:
+        raise ValueError(f"Unknown metric export options: {', '.join(sorted(map(str, invalid_options)))}")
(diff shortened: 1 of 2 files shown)


apps/betterangels-backend/reports/export_to_json.py
```diff
@@ -0,0 +1,70 @@
...
+    shelter_id = str(metrics.shelter_id)
+    start_date = metrics.start_date
+    end_date = metrics.end_date
+    date_range = f"{start_date.isoformat()}_{end_date.isoformat()}"
+    report: dict[str, object] = {}
+
+    if MetricsExportOptions.DAILY_OCCUPANCY_METRICS in selected_options:
+        report[MetricsExportOptions.DAILY_OCCUPANCY_METRICS] = {
+            metric.date.isoformat(): {
+                "shelter_id": shelter_id,
```
frank731: does shelter id need to be in every exported metric, in every day, could it just be at the top level

frank731 (commented):
One nit otherwise should be solid

apps/betterangels-backend/reports/tests/test_exporters.py
```diff
@@ -402,18 +402,16 @@ def test_metrics_to_json_exports_selected_metrics_as_date_keyed_report(self) ->
 
         assert filename == "20260601_20260630_shelter_report.json"
         assert json.loads(json_content) == {
-            "report": {
+            "shelter-1_report": {
```
frank731: this feels like it would cleaner if shelter id was still a field under report, feel like it would be hard to ingest if the schema is diff for each report

frank731 (changes requested):

frank731 (approved):

PR #2295, opened 2026-08-03
chore/op portal clean up

mikefeldberg:
just a few touchups
- remove unwired buttons
- remove unused tabs
- replace hardcoded shelter tags with address
- clean up room & bed forms

## Summary by Sourcery

Clean up the shelter operator portal UI and forms to reduce unused elements and improve display of shelter information.

New Features:
- Show shelter address as a dedicated column in the shelter table.
- Provide default empty-state messages for room and bed tables when no entries exist.

Enhancements:
- Render bed and room property checkbox groups only when corresponding property options are available, reducing clutter in the forms.
- Rename the shelter management sidebar section label to "Shelter Operation" for clearer navigation.

Chores:
- Remove unused or unwired UI elements, including the shelter tags column, the dashboard sort button, the sidebar operations link, and the reports index route in the management dashboard.

### DIFF
--- a/libs/react/shelter-operator/src/lib/components/ShelterManagement/segments/Beds/BedForm/sections/BedDetails.tsx
+++ b/libs/react/shelter-operator/src/lib/components/ShelterManagement/segments/Beds/BedForm/sections/BedDetails.tsx
@@ -13,34 +13,38 @@ export function BedDetails({
 }: SectionProps) {
   return (
     <FormSection title="Bed Details">
-      <Controller
-        name="demographics"
-        control={control}
-        render={({ field }) => (
-          <CheckboxGroup
-            name="demographics"
-            label="Demographics"
-            options={filteredPropertyOptions?.demographics ?? []}
-            values={field.value}
-            onChange={field.onChange}
-            error={errors.demographics?.message}
-          />
-        )}
-      />
-      <Controller
-        name="accessibility"
-        control={control}
-        render={({ field }) => (
-          <CheckboxGroup
-            name="accessibility"
-            label="Accessibility"
-            options={filteredPropertyOptions?.accessibility ?? []}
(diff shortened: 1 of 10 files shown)


libs/react/shelter-operator/src/lib/components/ShelterManagement/segments/Beds/BedForm/sections/BedDetails.tsx
```diff
@@ -13,34 +13,38 @@ export function BedDetails({
...
-      />
+      {!!filteredPropertyOptions?.demographics.length && (
+        <Controller
+          name="demographics"
+          control={control}
+          render={({ field }) => (
+            <CheckboxGroup
+              name="demographics"
+              label="Demographics"
+              options={filteredPropertyOptions?.demographics ?? []}
```
tglaz, line 24: minor - given the filter above can just do `filteredPropertyOptions.demographics ?? []`

is it possible to have this state?
- filteredPropertyOptions?.demographics === undefined
- field.value exists

if so, then we should address it - though can just do in separate ticket

libs/react/shelter-operator/src/lib/components/ShelterManagement/segments/Beds/BedTable.tsx
```diff
@@ -80,7 +80,11 @@ export function BedTable({
   onReserve,
   loading,
   loadingState,
-  emptyState,
+  emptyState = (
+    <div className="px-6 py-8 text-center text-sm text-gray-500">
```
tglaz: also minor:
it's static, so better define the EmptyState functional component at top of file and then pass:
emptyState={emptyState || EmptyState}

tglaz (approved):

PR #2299, opened 2026-08-06
[BP] added max-bar bucketing and fixed chart UI

Angus-Sun:
## Summary

Improves the `BarChart` component used in shelter reporting with two new features and several layout/style fixes.

### `BarChart` — max-bar bucketing (`maxBars` prop)

When a chart has more unique x-axis values than `maxBars` (default **40**), consecutive points are automatically grouped into evenly-sized buckets and their y-values are **averaged**. This prevents overcrowding for wide date ranges.

- Single-series charts: average the y-value across all points in the bucket
- Stacked/grouped charts: average per series (colorField) within each bucket
- Bucket x-axis label: `"first – last"` (e.g. `"Jul 7 – Jul 9"`)
- No bucketing occurs when data is within the limit (pass-through)

### `BarChart` — fixed canvas height (`chartHeight` prop)

Adds a `chartHeight` (px) prop that pins the plot canvas to a fixed height. The card then sizes itself around `header + canvas` instead of dividing a fixed card height between them.

- When the Bed Status legend wraps to extra rows the **card grows** rather than the chart shrinking
- Both charts always have the same title-to-plot distance regardless of legend height
- Backward-compatible: omitting `chartHeight` keeps the existing `flex-1` behaviour

### Layout & style fixes

- Legend placeholder div is always rendered (`min-h-[20px]`) so charts without a legend (Daily Occupancy) have the same vertical spacing from title to bars as charts with one (Bed Status)
- Chart card height increased to give G2's canvas enough room for rotated x-axis labels without clipping against the card boundary
- `clsx` → `mergeCss` from `@monorepo/react/shared` per codebase styleguide
- `bucketData` internals replaced with remeda `chunk` / `groupBy` / `meanBy` / `unique`

### Storybook

Two new stories: `BucketedStackedBarChart` and `BucketedVerticalBarChart` — 90-day datasets that exercise the bucketing path.
Screenshot 2026-08-05 211107
Screenshot 2026-08-05 211122

## Testing

- View the **Reports** tab for a shelter with seeded data to verify both charts render correctly at various date ranges
- Run Storybook (`yarn nx run storybook-react:storybook`) → **Charts → BarChart** to see the bucketed stories
- TypeScript: `npx tsc --noEmit -p libs/react/shelter-operator/tsconfig.lib.json`

## Summary by Sourcery

Improve shelter reporting bar charts with wide-range data aggregation and stable, consistent chart layouts.

New Features:
- Add automatic bucketing and averaging of wide-range bar chart data through a configurable maxBars prop, including per-series aggregation and range labels.
- Add an optional chartHeight prop for fixed plot sizing and use it to keep report chart layouts consistent.
- Add Storybook examples covering bucketed stacked and single-series charts.

Bug Fixes:
- Prevent chart legends and rotated axis labels from causing inconsistent spacing or clipping in shelter reports.
- Ensure report cards expand for wrapped legends instead of shrinking the plot area.

Enhancements:
- Align chart styling utilities with the shared mergeCss convention and memoize report chart data transformations.

Tests:
- Add unit coverage for pass-through data, bucket sizing and labels, averaging, multi-series aggregation, empty input, and insertion-order preservation.


libs/react/shelter-operator/src/lib/components/BarChart/BarChart.tsx
```diff
@@ -1,10 +1,81 @@
...
+ * averaged per series (colorField) within each bucket.
+ *
+ * The x-axis label for a bucket with more than one point becomes
+ * "firstLabel - lastLabel".
+ *
+ * **Ordering**: bucket boundaries are derived from the insertion order of
+ * unique x-values in `data`. Callers must pass data in the intended display
+ * order (e.g. chronological for date series) so bucket ranges are meaningful.
+ */
+function bucketData(
```
frank731: perhaps should have unit tests for this

libs/react/shelter-operator/src/lib/components/BarChart/BarChart.tsx
```diff
@@ -133,13 +210,25 @@ export function BarChart({
...
+  const xField = cfg['xField'] as string | undefined;
+  const yField = cfg['yField'] as string | undefined;
+  const normalizedMaxBars = Math.max(1, Math.floor(maxBars));
+  const displayData =
+    ((displayConfig as Record<string, unknown>)['data'] as
+      | Record<string, unknown>[]
+      | undefined) ?? [];
+  const finalData =
+    xField && yField
+      ? bucketData(displayData, xField, yField, colorField, normalizedMaxBars)
```
frank731: i think this data should be memiozed with useMemo or smth like that, running this here will recalc on any rerender

frank731 (dismissed):
Mainly just the memoization issue

vecchp (approved):

PR #2301, opened 2026-08-07
[BP] - add export modal

c4bae:
Add the modal for exporting a report that allows the user to select the format as well as select the desired metrics.

## Summary by Sourcery

Enable users to export shelter report metrics through a configurable download flow in the report interface.

New Features:
- Add a shelter report export API supporting CSV, XLSX, and JSON downloads with selectable metrics and date ranges.
- Add a report export modal that lets users choose the file format and metrics before downloading the report.

Enhancements:
- Centralize shelter reporting date-window resolution and standardize metric export generation, filenames, content types, and validation.
- Improve modal behavior by preventing background scrolling while open and allowing custom backdrop styling.

Build:
- Configure mypy to ignore missing type information for tablib.

Tests:
- Add service and API coverage for metric export formats, selections, validation, permissions, date ranges, and response metadata.
- Add frontend coverage for export requests, metric selection, download handling, and error notifications.

Chores:
- Move shelter export functionality from the reports package into the shelters API and service structure.


frank731 (dismissed):
/workspace/libs/react/shelter-operator/src/lib/components/ShelterProfile/segments/Schedules/ScheduleForm/ScheduleForm.tsx
    82:18  warning  Compilation Skipped: Use of incompatible library

lint failing

vecchp:
The modal looks good on its own terms. The thing I'd want settled before it merges is what it's
going to call, because as far as I can tell the backend export path isn't reachable yet.

**The exporters have no caller outside their own tests.** On `main`, the only files referencing
`metrics_to_zip`, `metrics_to_xlsx` or `metrics_to_json` are the three modules that define them
plus `reports/tests/test_exporters.py`. No resolver, mutation, query or view calls them, so
there's nothing for this modal to hit for CSV, XLSX or JSON. PDF is the exception — that one
runs client-side through `useExportPdf.ts`.

**`MetricsExportOptions` can't reach the frontend as it stands.** It's a plain `StrEnum` in
`reports/export_options.py`, not a `@strawberry.enum`, so it never enters the schema and never
appears in the generated TS types. Making it a Strawberry enum would let the dataset choices —
`DAILY_OCCUPANCY_METRICS`, `DAILY_BED_STATUS_METRICS`, `RESERVATION_METRICS`,
`AVG_DAYS_TO_OCCUPANCY` — come from codegen rather than being retyped here, and would keep them
from drifting apart.

That's also the shape of my one comment on the modal itself:

```ts
type ExportFileType = 'PDF' | 'CSV' | 'XLSX' | 'JSON';
const FILE_TYPES: ExportFileType[] = ['PDF', 'CSV', 'XLSX', 'JSON'];
```

These four are a local invention, and three of them name backend capabilities that aren't
exposed. If the export mutation ends up taking a format argument, this list should come from its
generated enum for the same reason.

None of this is a criticism of the modal — it's that the PR it depends on doesn't exist yet.
Worth deciding whether this waits for the export endpoint, or ships PDF-only with the other
three disabled behind a flag.

vecchp (approved):

PR #2306, opened 2026-08-12
Background-safe client document uploads: inline progress bar, per-file retry, and crash recovery (DEV-2199)

vecchp:
Replaces the upload progress drawer with an in-flow progress bar and a
dedicated upload screen, then hardens the upload pipeline behind it.

## What the user sees

- Uploads start the moment files are picked; the picker closes.
- A slim progress bar sits between header and content while uploads run.
  It lives in `Header` now, so every screen gets it — including the HMIS
  screens, which previously had no global upload surface at all.
- Tapping it opens a detail screen with per-file progress, Cancel, Retry,
  "Retry all N failed files", and "Dismiss failed".
- Terminal state is a snackbar when the detail screen isn't open.

## Correctness fixes

Found while reviewing the original branch:

- **Partial failures silently lost files.** With `failFast: true` one file's
  failure rejected the batch before the save step, so files already in S3
  were never recorded — and the UI showed them green with no retry. Item
  status now distinguishes `uploaded` (bytes in S3) from `done` (persisted),
  and docs uploads run with `failFast: false` so successes are saved.
- **A failed upload could never be dismissed.** Failed sessions are never
  auto-pruned, and the row's Cancel was gated to pending/uploading — so a
  repeatedly-failing file pinned the red bar to every screen permanently.
  Failed rows now offer Dismiss.
- **Concurrent uploads were hidden.** The detail screen scoped itself to the
  first resumed session's group id, so with two uploads running the bar
  counted files the screen never showed.
- **A missing client id left a zombie session** stuck at "Uploading 0 of N".
- **VoiceOver couldn't reach Cancel/Retry** — `accessible` on the row
  container collapsed the subtree on iOS.

## Reliability

- **Concurrency capped at 3** with exponential backoff + jitter on transient
  failures (network drops, 408/429/5xx); 4xx never retries. Previously every
  file uploaded at once via `Promise.all` and any failure was terminal —
  the worst combination for the field connections this app runs on.
- **Interrupted uploads resume on launch.** The save step is a GraphQL call
  that can't run without a JS runtime, so a process death between S3
  accepting a file and the save landing orphaned the object silently. A
  manifest is now written before anything is sent and projected from session
  state; on launch, files already in S3 with credentials inside the ~4 min
  reuse window are saved directly, everything else re-uploads.
- **Sign-out cancels in-flight uploads**, which would otherwise write to a
  client record the session no longer owns.

## Structure

- Session state moved from React context to module-scoped Jotai atoms, so
  uploads survive the screen that started them.
- Runners (abort controllers, retry) live in a registry, not on the session
  — session state is plain serializable data, which is what makes
  persistence possible. There's a test asserting it survives a JSON round trip.
- Progress counts derive from items instead of being hand-maintained in five
  places.
- Retry is in-place: one user action stays one session, and retrying several
  files costs one transport run rather than one per file.
- Blocking flows (HMIS documents, profile photos) no longer use the
  background store; they were putting phantom rows in the global bar.

## Testing

150 tests in `expo-betterangels`, 38 in `expo-shared-services`, 56 in
`expo-shared-ui-components`; typecheck and lint clean. No new import cycles
(verified with madge).

**Not yet exercised on a device.** The persistence, resume-on-launch,
concurrency and backoff paths are covered by unit tests only. Worth a real
run before merge, particularly: an upload interrupted by killing the app,
and the `Header` change, which now renders the bar on more screens than
before.

## Known gaps

- Android has no background upload continuation (`sessionType` is iOS-only).
  The stage copy says "while the app is open" rather than "in the
  background" because of this; persistence makes a suspended-and-killed
  Android upload recoverable on next launch, which is the mitigation, not a
  fix. Real Android background upload needs a foreground service.
- `uploadStageVisibleAtom` is UI state read by the upload runner to decide
  whether to show a snackbar; it belongs in a root-level observer.
- "Manifest" means two things in this code: the pipeline's refId↔file
  correlation and the persisted recovery record.


vecchp:
Addressed the review + Slack feedback in c89e832:

- **Upload stage screen** per the thread design: Ready (review files) → Uploading (per-file progress + cancel) → Done (per-file ✓/✗ with Retry, min 1.5s visible, auto-close). Dismissing mid-upload continues in the background.
- **Global WhatsApp-style top progress bar** instead of inline tree rows — visible on every screen, tap to reopen the stage, auto-hides when idle.
- **Progress UI decoupled from FileCard**: new standalone `UploadItemRow`; FileCard's upload mode deleted (kills the status-union drift and the `summary` role issue).
- **Store migrated to Jotai** (existing codebase convention); provider removed, cleanup timer map now prunes entries when they fire.
- **Completion confirmation fixed**: nested SnackbarProvider in modal-screen removed, so terminal snackbars render at the root and survive navigation.

typecheck + lint + full test suites for expo-betterangels, expo-shared-ui-components, and the app are all green. @tom @mikefeldberg whenever you get a chance.

vecchp:
Adjusted per feedback in 252047c: the confirmation step is gone — picking files starts the upload immediately and closes the picker, and the thin top bar is now the primary progress surface (it was there before, but hidden while the upload screen was open, so you'd never see it on the main path). Tapping the bar opens the per-file detail screen; completion still surfaces a root-level snackbar.

tglaz (approved):

PR #2307, opened 2026-08-12
fix(accounts): 500 on login-code for deactivated users; reactivate on member re-add (DEV-2516)

vecchp:
## Problem

Production incident: an existing user (deactivated by a bulk script) could not log in, and the backend returned HTTP 500.

- Login uses django-allauth headless with `ACCOUNT_PREVENT_ENUMERATION=True` and `LOGIN_BY_CODE_ENABLED=True`.
- With enumeration prevention, allauth's `RequestLoginCodeForm.clean_email` only resolves *active* users, so an existing-but-inactive account yields `self.input._user = None`.

### Bug 1 — `accounts/headless_views.py`
When `self.input._user` was `None`, the view unconditionally called `User.objects.create_user(...)`. For an existing inactive account this raised `UniqueViolation` on `accounts_user_email_key` → 500.

### Bug 2 — `accounts/services.py` (`member_add`)
`get_or_create(email=email, ...)` never reactivated existing users, so someone re-added to an org after bulk deactivation stayed inactive and then hit Bug 1. The lookup was also exact-case while `OrgInvitationInput` doesn't normalize email — mixed-case emails could miss the existing row and 500 on the unique constraint.

## Fix

Centralized user provisioning in `accounts/services.py`:

- `get_or_create_user_by_email(email, *, defaults=None)` — **pure find-or-create with no side effects on existing users**:
  - normalizes email (`strip().lower()`) before lookup
  - creates new users active with an unusable password and a uuid username (codebase convention)
  - accepts `defaults` (mirrors `get_or_create`) so `member_add` seeds names in the single INSERT — no extra save, no query-count change
  - returns existing-but-deactivated accounts **unchanged**
- `reactivate_user(user)` — explicit helper; called **only by authorized flows**.

Call sites:

- `headless_views.AutoCreateRequestLoginCodeView` — auto-creates **brand-new** emails only. An existing-but-inactive account is left untouched: allauth's enumeration-safe path returns the same generic success and sends nothing (`ACCOUNT_EMAIL_UNKNOWN_ACCOUNTS=False`) — no 500, no email, no state change. Anonymous requests cannot undo a deactivation.
- `services.member_add` → called by the `addOrganizationMember` mutation — admin-authorized re-add calls `reactivate_user(user)`.
- `backends.invite_by_email` — same latent bug pattern fixed; re-invitation calls `reactivate_user(user)`.

Making reactivation an explicit step (rather than a `reactivate=True/False` parameter) means new call sites **fail safe**: forgetting to reactivate is harmless, whereas a default-on flag could let a future anonymous flow silently reactivate accounts.

Transaction ownership stays at the call sites (the service is a plain unit of work) to avoid nested savepoints.

## Policy decision

Anonymous login-code requests **do not** reactivate deactivated accounts. Deactivation is strictly a Better Angels (platform) admin action — org admins only manage membership, and removing an org member never deactivates the `User`. Reactivation may come from BA-admin or org-admin flows: users deactivated by bulk scripts can be re-enabled by re-inviting them (`addOrganizationMember` / `invite_by_email`), which now reactivates automatically. `User.is_active` is global (not org-scoped), so there is no cross-org concern — the only rule is that no single org's request deactivates a user.

## Behavior notes

- Auto-created login-code users now get a uuid `username` instead of `username=email` (matches `member_add`/`invite_by_email`/`create_client`; also avoids the 150-char username limit).
- Email normalization (`strip().lower()`) lives only in `get_or_create_user_by_email`; call sites pass input through unchanged.

## Tests

- Login code for an existing inactive account: generic success, **no email sent, no state change, no duplicate row, no 500**.
- Brand-new email: still auto-creates an active user with an unusable password and an unverified `EmailAddress`.
- Brand-new mixed-case email: normalized end-to-end (user and `EmailAddress` rows stored lowercase).
- `member_add` reactivates existing inactive users; mixed-case email finds the existing user (no duplicate).
- `invite_by_email` reactivates a deactivated user re-invited with mixed-case email.
- `add_organization_member` with mixed-case email: no duplicate, no `IntegrityError`, user reactivated.
- Unit tests: provisioning service leaves inactive users unchanged; `reactivate_user` reactivates and no-ops on active users.

## Jira

- [DEV-2516](https://betterangels.atlassian.net/browse/DEV-2516) — this work
- [BACS-103](https://betterangels.atlassian.net/browse/BACS-103) — related customer ticket ("Not receiving one time code")

## Verification

- `pytest accounts/tests/` → **116 passed** (36 warnings)
- `ruff check` + `ruff format --check` → clean
- `mypy` on changed modules → clean

[DEV-2516]: https://betterangels.atlassian.net/browse/DEV-2516?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/betterangels-backend/accounts/headless_views.py
+++ b/apps/betterangels-backend/accounts/headless_views.py
@@ -29,26 +29,27 @@ class AutoCreateRequestLoginCodeView(RequestLoginCodeView):
         if not self.input._user:  # type: ignore[union-attr]
             email = self.input.cleaned_data.get("email")  # type: ignore[union-attr]
             if email:
-                email = email.strip().lower()
+                # get_or_create_user_by_email normalizes (strips + lowercases)
+                # the email before lookup and storage.
                 with transaction.atomic():
                     user, created = get_or_create_user_by_email(email)
                     if created:
                         # Brand-new email: auto-provision so allauth can issue a
-                        # real code.  Reuse any existing EmailAddress row
-                        # (case-insensitive) so mixed-case input can't create a
-                        # duplicate; new rows are primary and unverified.
-                        email_address = EmailAddress.objects.filter(user=user, email__iexact=email).first()
-                        if email_address is None:
-                            EmailAddress.objects.create(
-                                user=user,
-                                email=email,
-                                primary=True,
-                                verified=False,
-                            )
+                        # real code.  The user row is brand-new, so it can't own
+                        # an EmailAddress yet — create a primary, unverified one
+                        # with the normalized address.
+                        assert user.email is not None
+                        EmailAddress.objects.create(
+                            user=user,
+                            email=user.email,
+                            primary=True,
+                            verified=False,
+                        )
                         self.input._user = user  # type: ignore[union-attr]
                     elif user.is_active:
-                        # Defensive: an active account allauth didn't resolve
-                        # still gets a real code.
+                        # Defensive: allauth's form already resolves active
+                        # users case-insensitively, but if it somehow didn't,
+                        # still issue a real code.
                         self.input._user = user  # type: ignore[union-attr]
                     # else: existing-but-inactive — leave ``_user`` unset so
                     # super().post() fakes success without sending anything
--- a/apps/betterangels-backend/accounts/headless_views.py
+++ b/apps/betterangels-backend/accounts/headless_views.py
@@ -17,13 +17,12 @@ class AutoCreateRequestLoginCodeView(RequestLoginCodeView):
     calls ``initiate(user=None, ...)`` which silently fakes success without
     sending a real code.
 
-    We override ``post()`` to provision the account and set ``self.input._user``
-    so the verification process finds it and issues a real code.  Provisioning
-    is delegated to :func:`accounts.services.get_or_create_user_by_email`, which
-    reuses an existing account — reactivating it if it was deactivated — instead
-    of creating a duplicate (which would raise a unique-constraint violation).
-    This follows allauth's subclassing pattern: ``self.input._user`` is the
-    documented protocol variable that ``RequestLoginCodeView.post()`` reads.
+    We override ``post()`` to provision brand-new emails and set
+    ``self.input._user`` so the verification process issues a real code.
+    An email that already belongs to an existing-but-deactivated account is
+    deliberately left untouched — no reactivation, no email — so an anonymous
+    request can't undo an admin's deactivation; ``super().post()`` then fakes
+    success exactly as it does for unknown accounts.
     """
 
     def post(self, request: HttpRequest, *args: Any, **kwargs: Any) -> HttpResponse:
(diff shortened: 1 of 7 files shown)


apps/betterangels-backend/accounts/tests/test_services.py
```diff
@@ -109,6 +110,36 @@ def test_create_org_atomic() -> None:
...
+
+@pytest.mark.django_db
+def test_get_or_create_user_by_email_new_user() -> None:
+    """Brand-new email yields an active user with an unusable password and a
+    normalized (lowercased, stripped) email."""
+    user, created = get_or_create_user_by_email(" NewUser@Example.com ")
+
+    assert created
+    assert user.email == "newuser@example.com"
+    assert user.is_active
```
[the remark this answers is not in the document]
mikefeldberg: this is unnecessary as it would effectively be testing django's get_or_create. If a record is found, it's retrieved without mutation.

apps/betterangels-backend/accounts/tests/test_headless_views.py
```diff
@@ -0,0 +1,89 @@
+"""Tests for ``accounts.headless_views.AutoCreateRequestLoginCodeView``.
+
+Login-by-code self-signup: when ``ACCOUNT_PREVENT_ENUMERATION=True`` the
+headless form only resolves *active* users, so an existing-but-deactivated
+account (e.g. disabled by a bulk script) leaves ``self.input._user = None``.
+Brand-new emails are auto-provisioned; existing-but-inactive accounts are left
+untouched (no 500, no duplicate, no email) so anonymous requests can't undo an
+admin's deactivation.
```
mikefeldberg: "admin" is ambiguous here. It's referring to BA admin, but could be interpreted as referring to an org admin. Maybe change to "anonymous requests can't reactivate a deactivated user"

apps/betterangels-backend/accounts/headless_views.py
```diff
@@ -18,25 +17,41 @@ class AutoCreateRequestLoginCodeView(RequestLoginCodeView):
...
 
-    We override ``post()`` to create the user first and set ``self.input._user``
-    so the verification process finds the account and issues a real code.  This
-    follows allauth's subclassing pattern — ``self.input._user`` is the
-    documented protocol variable that ``RequestLoginCodeView.post()`` reads.
+    We override ``post()`` to provision brand-new emails and set
+    ``self.input._user`` so the verification process issues a real code.
+    An email that already belongs to an existing-but-deactivated account is
+    deliberately left untouched — no reactivation, no email — so an anonymous
+    request can't undo an admin's deactivation; ``super().post()`` then fakes
```
mikefeldberg: "admin" is ambiguous here. It's referring to BA admin, but could be interpreted as referring to an org admin. Maybe change to "anonymous requests can't reactivate a deactivated user"

mikefeldberg (approved):

apps/betterangels-backend/accounts/headless_views.py
vecchp: Updated in a3a5122 — rewrote to "anonymous requests can't reactivate a deactivated user", and the PR description now spells out that deactivation is strictly a BA platform-admin action while reactivation is BA or org admin.

apps/betterangels-backend/accounts/tests/test_headless_views.py
vecchp: Same fix applied here — rewrote to "anonymous requests can't reactivate a deactivated user".

PR #2310, opened 2026-08-13
fix(notes,tasks)!: org-scope every note, task and referral write

vecchp:
**Breaking:** every note, task and referral write now resolves through the active organization.

Replayed onto `main` after #2340 squash-merged and deleted this branch's base. The net contribution is unchanged; #2351's `full_clean()` work, #2340's membership check in the `teams` query, and `tasks/migrations/0004` are all preserved intact — verified by diffing against `main` for reversals.

## The hole this closes

Guardian grants object permissions to the permission group that *created* a record, and that grant says nothing about which organization the caller is currently acting as. A multi-org user who legitimately reached a record through org A could therefore act on it while their header said org B, and the write was accepted and attributed to the wrong organization.

Four of the call sites — `generateNoteFileUploads`, `resolveNoteFileUploads`, `deleteServiceRequest`, `updateReferral` — had no compensating `.filter(organization_id=...)` in the resolver body at all.

## What changes

`PermissionedQuerySet` gains a **required** `organization_field`, so each of the 14 call sites states in the decorator whether the endpoint is org-owned (a field path) or deliberately platform-shared (`None`). `createNote`, `createTask`, `importNote` and `createNoteServiceRequest` move to `@hasOrgPerm`; `deleteNote` and `deleteTask` gain `@permissionedQuerySet`. Reads stay platform-shared by design — organizations coordinate on shared clients and see each other's interactions.

Making the argument required is what makes that reviewable: adding a mutation now forces an answer, and mypy rejects a call site that omits it (`error: Missing named argument "organization_field"`).

## Compatibility with app builds predating #2330

The `X-Organization-ID` interceptor landed in #2330. Older builds send no header, and without a fallback they would lose note and task creation entirely — the outreach app's primary function.

`allow_implicit_org=True` adopts the caller's **only** organization when the header is absent, and refuses when there is not exactly one: nothing to adopt, or no way to choose between two. It never guesses first-match, which is the behaviour this stack exists to remove. Adoption writes back to `request.organization_id`, because resolver bodies re-read it — `update_note` derives a permission group from it and would otherwise reach the ORM with `"None"`.

Removed once every build sends the header. `grep -rn allow_implicit_org` is the removal list.

## Also here

- `deleteNote` keeps its `NoteType` return shape for client compatibility; only `id` is meaningful on the result (#2338).
- The org-owned vs platform-shared model goes into `apps/betterangels-backend/docs/permissions.md`, where the rest of the permission architecture lives.
- The `docs/teams_org_scoping.md` runbook that earlier revisions of this branch carried is **dropped**. Its live content had all shipped — the `old_team` and `Team.slug` drops in #2331, the `HasOrgPerm` group check in #2327, clearable `teamId` in #2331 — and the one part still outstanding, making a cross-org team reference unstorable, is #2312's composite foreign key.

## Verification

`1448 passed, 31 skipped` (`main` is at 1394); mypy clean across 387 files; `ruff check` and `ruff format --check` clean; `export_schema` reproduces `schema.graphql` exactly.

Mutation-tested — each mutation fails a distinct set, measured on the replayed branch:

| Mutation | Failures |
| --- | --- |
| `PermissionedQuerySet` ignores its declared `organization_field` | **20** — the org-scoping cases across notes, tasks, referrals, attachments and the extension's own tests |
| sole-organization adoption weakened to first match | **4** — the multi-org denial cases |
| adoption applied even when a header *was* sent | **32** |
| `permissioned_queryset`'s organization filter dropped | **7** — accounts, reports, shelters membership cases |

## Turned up while reviewing this

Fixes are in flight rather than tickets: #2350 (tests shared one Redis cache), #2355 (clients could send a headerless org-scoped query), #2343 (non-numeric `teamId` 500s), #2348, #2338, #2344, #2349.


vecchp:
## Reconciliation list from #2331

#2331 has grown to absorb the legacy-enum removal, so this branch will need a few deliberate resolutions rather than blind conflict-fixing. Recording them here so they aren't rediscovered as surprises.

### 1. `common/team_shim.py` is gone

Deleted in #2331. Its two helpers moved to `common/graphql/utils.py`, using **your implementations verbatim** so the files converge. `teams/schema.py` no longer imports from it either — #2329 replaced that with the idiom.

### 2. `apply_maybe` should be deleted, not merged

Its only two call sites here — `notes/schema.py` and `tasks/schema.py`, both `team_id` — don't need it. `strawberry.asdict` already resolves all three `Maybe` states:

| client sends | `asdict` produces |
| --- | --- |
| `teamId: "7"` | `{'team_id': '7'}` — unwrapped |
| `teamId: null` | `{'team_id': None}` — clears |
| omits `teamId` | key absent — leaves alone |

The bug both PRs fix was never a missing helper: it was that the mutations *overrode* asdict's correct output with an unconditional assignment. #2331 fixes it by deleting the override, keeping only the str-to-int narrowing. `common/tests/test_graphql_utils.py` there pins asdict's behaviour so this can't silently regress.

### 3. `maybe_value` is gone; `maybe_int_value` stays

`maybe_value` had no production callers after #2329 and was inlined into `maybe_int_value`. If this branch still imports it, switch to the idiom: `data.name.value if data.name else None`.

### 4. `teams/0002` collides

`main` now has `teams/migrations/0002_team_is_active.py` (from #2329) and #2331 renumbered its slug-drop migration to `0003`. This branch will hit the same collision on rebase — it merges "cleanly" as text and then breaks the migration graph, so check `makemigrations --check` explicitly.

### 5. What #2331 deliberately left for this branch

- the composite `(team_id, organization_id)` FKs on notes and tasks, and the `(id, organization)` constraint they target
- `teams/validators.py` and the `Model.clean()` methods that use it

Also worth reading: the comment on #2312 about those `clean()` methods not running on updates, because `task_update` / `note_update` don't call `full_clean()`. Tracked in #2335.

vecchp:
Follow-up to the reconciliation list above — #2331 now also creates `teams/validators.py`, so this branch will merge into an existing file rather than adding a new one.

**One change is required when you rebase:** `validate_team_in_org` imports `Team` at module level:

```python
from .models import Team
```

That becomes a circular import. `Team.name` on #2331 declares a field validator from this module:

```python
name = models.CharField(max_length=255, validators=[validate_has_alphanumeric])
```

so `teams.models` imports `teams.validators`. A top-level `from .models import Team` in the other direction makes `import teams.models` fail outright — it's a loud `ImportError` at startup, not a subtle one, but it will look like this branch broke something when the cause is the merge.

The fix is to move the import into the function body:

```python
def validate_team_in_org(*, team_id: int | None, organization_id: int | None) -> None:
    from .models import Team
    ...
```

This also happens to be what your module docstring is already reaching for — keeping validators callable from `notes.models` / `tasks.models` without a models-depend-on-services inversion.

**Why the file exists on #2331 at all:** the alphanumeric name rule moved out of `team_create`/`team_update` and onto the model as a field validator. It was service-only, so `TeamAdmin`'s ModelForm accepted a team named `"---"` — the same editable-name path the `teams.0003` dedupe step cleans up after. `_validate_name` now holds only the uniqueness check, and its docstring records that the check is there to name the team in the message rather than to catch something `full_clean()` would otherwise miss.

**One behaviour change to be aware of** if you have tests asserting on it: a whitespace-only name now reports `"This field cannot be blank."` instead of the alphanumeric message, because the service strips it to `""` first and `full_clean()` skips validators for empty values.

vecchp:
Rebuilt as #2392, stacked on #2390 — **no longer breaking**, and 12 files instead of 29.

Rebuilt rather than rebased, the same way #2312 was: this branch carries 36 commits that are largely earlier forms of work since merged, so `git rebase --onto` conflicts on commit 1 of 36. Comparing trees instead of histories showed what was actually left.

What dropped out:

- **`allow_implicit_org` — 30 mentions.** The whole compatibility shim. Pointless now that no pre-header builds are live and #2390 made the last headerless fallback strict. This PR's own body called `grep -rn allow_implicit_org` "the removal list" — so it never needed to ship. Dropping it is what makes #2392 non-breaking.
- `organization_get_sole_for_member` and `active_organization` — both existed only to serve that fallback.
- `permission_group_for_request` — would have shipped unused; it belongs with the de-guessing follow-up.

What survived intact: `organization_field` as a required argument on `PermissionedQuerySet`, `require_organization_id`, the nine call-site declarations, `docs/permissions.md`, and the new machinery tests. 1507 passing.

Two things #2392 deliberately leaves out, both flagged there: the nine sites that still resolve the *attributed* org by first match (its `docs/permissions.md` now names all nine — this PR's version listed three, assuming its own notes/tasks fix had landed), and the `HasOrgPerm` moves, which change the denial from a structured `OperationInfo` to a raw GraphQL error.

Closing in favour of #2392.

PR #2311, opened 2026-08-13
fix(accounts): make the permission sync fail loudly instead of adding a command

vecchp:
## The question this answers

The sync already runs on `post_migrate` ([accounts/apps.py:35](apps/betterangels-backend/accounts/apps.py#L35)), and every deploy runs migrations — so it always happens. A management command that calls the same function adds no safety: **if the automatic run failed, a manual run fails identically.**

The original version of this PR added `sync_org_permission_groups` with a `--check` mode. That was tooling around a defect rather than fixing it.

## The actual defect

```python
try:
    for org in Organization.objects.all():
        reconcile(org)
except Exception:
    return          # <- also skips _sync_template_permissions() below
```

Any error — not just the "tables not ready" case the comment described — aborted the handler and returned, skipping the template-permission sync too. **A deploy could leave every organization on stale permissions and still report success.** And the `--check` command couldn't reveal it, because checking took the same path.

## The fix

- Only `ProgrammingError` / `OperationalError` are tolerated — the genuine incomplete-schema case — and they're **logged**, not passed over silently. Everything else propagates and fails the migrate.
- Local-dev role seeding still never fails a migrate, but logs instead of `except Exception: pass`.
- `sync_org_permission_groups` and its tests are **deleted**.

## Why the tolerance isn't load-bearing

I checked rather than assumed: with the `except` removed **entirely**, a full fresh-database migrate completes fine. Django applies every migration before emitting `post_migrate`, so the tables always exist by then. Only a targeted `migrate <app>` can reach that branch — not a deploy path.

## Tests

Coverage moves from the command to the handler: drift repair in both directions (missing and surplus permissions), idempotency, an unexpected error propagating, and an incomplete schema being tolerated and logged.

The propagation test **fails if the blanket `except Exception` is restored** — verified.

## Deploy impact

The runbook's "run `manage.py sync_org_permission_groups`" step goes away; the sync is automatic and now fails loudly if it can't complete. That runbook edit lands in #2310, which owns the file.

## Verification

`1317 passed, 31 skipped` on a fresh database. `ruff` and `mypy` clean with the cache cleared.

Independent of the teams stack — based on `main`.


vecchp:
Closing — this PR's reason to exist is gone.

It originally carried **two** commands:

- `audit_team_org_scoping` — reported cross-org team references. This was the real content, and it was retired deliberately: cross-org teams are now unstorable, because notes and tasks carry a composite foreign key on `(team_id, organization_id)` referencing `teams_team(id, organization_id)`. There is nothing left to audit.
- `sync_org_permission_groups` — turned out to duplicate work that already happens. The sync is wired to `post_migrate` (`accounts/apps.py`), so every deploy performs it; a command calling the same function could only fail the same way an automatic run did.

Nothing in the teams stack depends on this, and the sync works today.

**What is not shipping as a result:** the `except Exception` narrowing in `accounts/signals.py`. That handler still catches broadly, so a failed permission sync is quiet rather than fatal. It is a pre-existing, latent issue rather than a regression, and it is recorded as a known gap in the permission-sync step of `docs/teams_org_scoping.md` (shipping in #2310) so that missing `teams.*` permissions after a deploy points an operator at the migrate logs instead of at the feature.

The branch `chore/permission-sync-tooling` is left on the remote — the signals fix and its tests are one `gh pr create` away if we want them later.

PR #2312, opened 2026-08-13
fix(teams): stop the admin repointing a team's organization

vecchp:
Rebased on `main` at `8a4ee562`.

## What this does

A note or task must not reference a team belonging to a different organization. `teams.validators.validate_team_in_org` is the single definition of that rule, enforced from `Note.clean()` and `Task.clean()` since #2339.

This PR started by adding a composite foreign key on `(team_id, organization_id)` to make the rule structural. Enumerating the writers that can actually reach the field showed it defended nobody, so it is gone. What is left is the one hole that was real, plus the one writer that had no guard.

### Every writer that can reach `Note.team` / `Task.team`

| Writer | Reachable | Guard |
| --- | --- | --- |
| `createNote`, `updateNote`, `createTask`, `updateTask` — `team_id` is a **client-supplied** `Maybe[ID \| None]` input | yes | `note_create` / `note_update` / `task_create` / `task_update` all `full_clean()` before `save()` |
| `NoteAdmin`, `TaskAdmin` — both offer `team` **and** `organization` as editable fields | yes | ModelForm `_post_clean()` → `full_clean()` |
| `TeamAdmin` — repointing a team's organization strands every record already holding it | yes | **the `disabled` form field in this PR** |
| `updateTeam` | **no** — `UpdateTeamInput` is `id`, `name`, `is_active` | — |
| `queryset.update()` / `bulk_create` / `bulk_update` / raw SQL | **none exist** — nothing under `notes/`, `tasks/`, `teams/` or `common/` writes `team` that way | — |
| `Note.revert_action` | yes | **`clean()`, added here** |

`Model.clean()` runs unconditionally inside `full_clean()` — `exclude` only narrows `clean_fields`, `validate_unique` and `validate_constraints` — so a field a form drops cannot take `validate_team_in_org` down with it.

## The admin could move a team between organizations

`TeamAdmin` was a bare `ModelAdmin` and `Team` has no `clean()`, so its change form offered `organization` as an editable dropdown:

```
Team  TeamAdmin  editable -> ['organization']          readonly=()   <- before
Note  NoteAdmin  editable -> ['organization', 'team']  readonly=('created_by', …)
Task  TaskAdmin  editable -> ['organization', 'team']  readonly=('created_at', …)
```

Moving a team leaves every note and task holding it in a state its own `clean()` rejects: unsavable, failing validation on a field the next editor never touched, with nothing on the page to say why. GraphQL never reached this — `UpdateTeamInput` carries only `name` and `is_active` — so this form was the only way in.

`organization` is now **`disabled` on the change form**, not listed in `readonly_fields`. That distinction is load-bearing. A read-only field is dropped from the form, which puts it in the ModelForm's validation exclusions — and `UniqueConstraint.validate` skips any constraint whose expressions reference an excluded field. Using `readonly_fields` took `unique_team_name_per_org` down with it: renaming a team to a name another team in the organization already held stopped being a form error and became an unhandled `IntegrityError`. A disabled field stays on the form, so the name constraint still validates, and ignores submitted data in favour of its initial value, so the team still cannot move. It also keeps the field in its declared position instead of appended after the read-only ones.

The rule is deliberately not duplicated onto `Team.clean()`. Rejecting the change needs the previous organization and a count of referencing rows — two queries and data the caller did not supply, which `docs/styleguides/python.md` and HackSoft both place in the service layer rather than `clean()`.

Creating a team still offers the dropdown.

## Reverting a note validated nothing

`Note.revert_action` replays one pghistory event's diff field by field and called `save()` directly, so a restored `team` could be paired with an `organization` from a different point in the note's history. It now calls `clean()` first.

`clean()` and not `full_clean()`: every value being restored was persisted once already, so re-checking that each foreign key still exists would make a revert fail on unrelated rows deleted since — turning working reverts into errors. It also costs 4 queries per reverted event, which is why `full_clean()` here moved `test_revert_note_mutation_restores_note_details` from 15 queries to 19. `clean()` costs one query, and none at all when the note has no team.

`NoteReverter` also reverts the note through pghistory's own `Events.revert()`, which restores a whole row snapshot via `update_or_create` — consistent by construction, so it needs nothing here.

## Why no database constraint

Postgres cannot express this as a same-table constraint, and Django blocks the attempt twice:

```
CheckConstraint(condition=Q(team__organization=F("organization")))
UniqueConstraint(fields=["team"], condition=Q(team__organization=F("organization")))

  models.E041  'constraints' refers to the joined field 'team__organization'
  FieldError:  Joined field references are not permitted in this query
```

A conditional `UniqueConstraint` compiles its condition into the `WHERE` clause of a partial unique index on one table, so it can only reference that table's own columns; index predicates cannot contain subqueries either. The fact needed — *this team's* `organization_id` — lives in another row of another table, which leaves a foreign key, a trigger, or application code. Django ships no foreign-key constraint class (`ForeignObject` is a virtual relation that emits no DDL), and `django-multitenant` is the packaged form of the same composite-FK mechanism but is unadoptable here: its backend subclasses plain `postgresql` while ours is PostGIS plus AWS IAM token auth, and it requires `TenantModel` / `TenantForeignKey` across every model.

Given that application code already covers every reachable writer, the guarantee is not worth a bespoke `BaseConstraint` subclass. Removing it deletes `common/constraints.py`, its tests, both `AddConstraint` migrations, and `Team.unique_team_id_per_org`, which existed only as the composite key's referenced unique constraint.

## Migrations

- `teams/0004_alter_team_unique_team_name_per_org` — `AlterConstraint` carrying `violation_error_message` alone, which is what gives the admin its readable duplicate-name error. Metadata only: no DDL, and no `ACCESS EXCLUSIVE` window on deploy.
- `notes/0005_null_cross_org_teams`, `tasks/0006_null_cross_org_teams` — `RunPython` detaching any pre-existing cross-org row. Kept: such a row cannot be saved again, because `clean()` rejects the pair on every later edit, so it fails validation on a field the editor did not touch. It logs the pks it detaches, since the step cannot be undone and a silent count of zero is indistinguishable from a silent count of many.

Everything here is what `makemigrations` wrote. There is no `RunSQL` in this PR.

## Tests

`teams/tests/test_admin.py` is new — the app had no admin tests:

- `organization` is offered on create
- `organization` is disabled on change
- posting the real change form with a different organization does not move the team
- renaming a team to a name already taken in the organization is still reported as a form error rather than reaching Postgres

`notes/tests/test_models.py` adds `NoteRevertActionTestCase`: a revert restores a same-org team, and refuses a team from another organization while leaving the stored value untouched.

The five existing `clean()` tests in `notes/tests/test_models.py` and `tasks/tests/test_models.py` are unchanged.

**Mutation-tested.** Dropping the `clean()` call from `revert_action` fails the cross-org revert test and nothing else. Putting `get_readonly_fields` back in place of the disabled form field fails both the disabled-field test and the duplicate-rename test, so the guard is pinned at the surface it is reachable from.

## Verification

- **1524 passed, 31 skipped, 0 failures** on a **freshly created** database, so the migrations actually run
- `makemigrations --check --dry-run` → no changes detected, twice
- mypy level with the merge base: 33 errors over 394 files, all pre-existing and none in the files this PR touches. `ruff check` and `ruff format --check` clean
- `schema.graphql` untouched — no API surface here

Net: **9 files, +259 / −2.**


vecchp:
Heads up from #2331 — this PR adds `Model.clean()` to `Note` and `Task` to reject a cross-org team, but **the validator will not run on a normal update.**

`docs/styleguides/python.md` requires services to "call `full_clean()` before `save()`", and two of the three don't:

| service | line | `full_clean()`? |
| --- | --- | --- |
| `teams/services.py:42,60` | `team.save()` | **yes** — with a comment citing the styleguide by name |
| `tasks/services.py` | `task.save()` | no |
| `notes/services.py` | `note.save()` | no |

Django only runs `clean()` from `full_clean()`, so as things stand the new rule fires through the admin's `ModelForm` — which is what the `clean()` docstrings here describe — and not through `updateNote` / `updateTask`, the paths that actually assign teams.

The composite `(team_id, organization_id)` FK in `notes/0004` and `tasks/0004` still backstops it at the database level, so this isn't a hole in the enforcement. It does mean the `clean()` methods are close to dead code on the write path, and the error a user sees comes from an `IntegrityError` rather than the `ValidationError` the validator was written to raise.

Worth deciding which you want before merge:

1. Add `full_clean()` to `task_update` / `note_update` here, and accept that it will surface any existing fixture rows that don't validate.
2. Leave it, and drop or re-scope the `clean()` methods so nobody assumes they guard the API.

Tracked with the wider `Maybe`/services cleanup in #2335 either way.

Context: #2331 has absorbed the enum removal, the `old_team` drop and the `Team.slug` drop from this branch, so this PR reduces to the composite-FK work. `apply_maybe` is also no longer needed there — `strawberry.asdict` already resolves all three `Maybe` states, so its two call sites here can be deleted rather than merged.

PR #2313, opened 2026-08-13
docs(permissions): encode the shared-layer vs org-owned permission model

vecchp:
## Goal

Make the shared-layer vs org-owned permission model explicit in code and tests.

Consolidates #2314 (cross-org client data policy), whose product decisions are now answered — see below. Both PRs reduced to the same job: write the layer down and test it.

## Decided model

| Resource | Layer | Notes |
| --- | --- | --- |
| Client profiles — read **and write** | **Platform-shared** | Clients are shared across orgs; there is no concept of client ownership |
| Client profile **delete** | **Platform-shared** (accepted) | See the accepted risk below |
| Client documents — read | **Platform-shared** | `Attachment.perms` is ADD + VIEW, so no cross-org delete |
| Note/task **reads** | **Platform-shared** | Orgs coordinate on shared clients and see each other's interactions |
| Note/task **writes** | **Org-owned** | Header-driven via `HasOrgPerm`; object perms granted per org at creation |
| Note private details | **Org-owned** | |
| Teams, members, reports, shelter ops | **Org-owned** | |

The permission templates already implement this, so **no behavior change is needed** — this PR is encoding and tests:

- `notes/groups.py` `CASEWORKER` holds `ClientProfile` ADD/CHANGE/DELETE/VIEW → shared client writes.
- `Task.perms` is ADD + VIEW only, CHANGE/DELETE granted per object at creation → tasks write-owned by the creating org.
- `Attachment.perms` is ADD + VIEW only → shared-read documents, no cross-org delete.

## Accepted risk: cross-org delete of a client profile

Decided to leave as-is — clients are shared and there is no ownership concept, and we do not want to change app behavior now.

Recorded so it is a known, deliberate risk rather than a surprise: `delete_client_profile` filters on `ClientProfile.perms.DELETE`, which `CASEWORKER` holds globally, and `Note.client_profile` is `on_delete=CASCADE`. So any caseworker in any org can delete a shared client profile, and doing so **also deletes other orgs' notes** for that client — note rows are org-owned, so this destroys data outside the deleting org's own layer. (`Task.client_profile` is `SET_NULL`, so tasks survive with no client.)

If that is ever revisited, the cheapest mitigations that preserve today's delete rights are soft-deleting the profile, or `PROTECT`ing the cascade when another org holds notes.

## Cross-org task visibility

Tasks stay org-owned for writes. Reads are currently shared: `tasks_for_user` does not filter by organization, and `Task.perms.VIEW` resolves through the global fallback — so one org already sees another org's tasks on a shared client, with no indication of who owns them.

Decision: show them as read-only history rather than hiding them, provided the UI marks the owning org and does not offer edit/complete/reassign affordances on another org's task. Hiding cross-org tasks is the fallback if labelling turns out to be impractical. **Frontend work — not in this PR.**

## Next steps

1. Write the layer table above into `docs/permissions.md` (or extend `docs/teams_org_scoping.md`).
2. Annotate `notes/groups.py` and `shelters/groups.py` marking which template perms belong to the shared layer and why.
3. Rename/annotate the ~35 codified cross-org permission tests from accidental ("CM in different org should succeed") to deliberate ("shared client data — cross-org visible"), and add explicit tests for the org-private parts (private details, note/task writes, teams).
4. Add a test pinning the accepted delete behavior, so a future change to it is a conscious one.

## Note

The optional step-4 architecture work from the original scope — replacing guardian's global-permission fallback with an explicit platform-shared mechanism — is what would let the defensive org filters come out of the note/task update resolvers. Still worth doing, still larger than this PR.

## Status

Model decided; this is now a codify-and-test PR. No behavior change intended.

---

## Step 4 (guardian's global fallback) — now the main architectural item

Since #2310 made `organization_field` a **required** argument on
`PermissionedQuerySet`, the defensive filters are no longer scattered
per-resolver: every call site declares its layer in the decorator, and the
filter is applied in one place. That changes the shape of this work.

**What is left to remove.** `filter_for_user` OR's in
`Exists(user.user_permissions ...)` and the group equivalent, so *any*
model-level grant matches every row in that table — object permissions stop
being a row filter at all. Nothing detects that; the `organization_field`
filter merely happens to narrow it back down.

Concretely:

1. Audit which permission templates grant model-level (not per-object)
   permissions, and for each decide whether the model is shared or org-owned.
2. Replace the fallback with an explicit shared-layer mechanism, so
   `filter_for_user` cannot match rows the caller holds no object grant on.
3. Only then can `organization_field` be reconsidered — and the honest answer
   is probably that it should stay, since it also confines the **multi-org
   caller** (object perms follow the record's creating group and ignore the
   `X-Organization-ID` header entirely). That is a separate concern from the
   fallback and does not go away with it.

The required argument is what makes step 1 mechanical: every mutation's
intended layer is now written down in its decorator rather than inferable
only from whether someone remembered a filter.

See `docs/teams_org_scoping.md` → "Why the org filter is load-bearing, and
what it is not".

## Summary by Sourcery

Add an operational command and tests for keeping organization permission groups synchronized with their templates.

New Features:
- Add a management command to reconcile organization permission groups with their templates, including a read-only check mode.

Enhancements:
- Provide operational verification and repair for permission-group drift across organizations.

Tests:
- Add coverage for detecting missing and extra permissions, non-mutating checks, synchronization, and post-sync consistency.


vecchp:
Closing: the diff here is a stray `sync_org_permission_groups` management command plus tests and a no-op file move, none of which is what the title describes. The body was the useful part, and it now lives in `apps/betterangels-backend/docs/permissions.md` via #2310 — the org-owned vs platform-shared layer, what retiring guardian's global-permission fallback actually requires, and the accepted cross-org client-profile delete risk with its two candidate mitigations.

#2315 had a byte-identical diff and is closed for the same reason.

If the permission-sync command is wanted, it should ship as its own PR judged on its own merits rather than riding along here.

PR #2314, opened 2026-08-13
product: cross-org client data write/sharing policy (decisions + implementation)

vecchp:
## Goal

Product decisions for the shared client layer, then encode them in the permission model and tests.

## Decisions (answered)

| # | Question | Decision | Matches today? |
| --- | --- | --- | --- |
| 1 | Can org B **edit** org A's client profile? | **Yes — client profiles are global/shared** | ✅ yes |
| 2 | Can org B **complete/reassign** a task on a shared client? | **No — tasks are org-specific** | ✅ yes |
| 3 | Are client **documents** shared-read? | **Yes — shared read** | ✅ yes |

**All three decisions confirm current behavior, so this PR needs no behavior change** — only encoding and tests. The permission templates already implement the matrix:

- `notes/groups.py` `CASEWORKER` grants `ClientProfile` ADD/CHANGE/DELETE/VIEW (global CRUD → shared edits ✔).
- `Task.perms` is **ADD + VIEW only**, with CHANGE/DELETE granted per-object at creation → a task is write-owned by the org that created it ✔.
- `Attachment.perms` is **ADD + VIEW only** → documents are shared-read, and no cross-org delete ✔.

## One question still open: cross-org **delete** of a client profile

Decision 1 covers *edit*; deletion was not addressed, and it is the case that matters most.

Today `delete_client_profile` filters on `ClientProfile.perms.DELETE`, which `CASEWORKER` holds globally — so **any caseworker in any org can delete any client profile**. And `Note.client_profile` is `on_delete=CASCADE`:

```python
# notes/models.py
client_profile = models.ForeignKey("clients.ClientProfile", on_delete=models.CASCADE, ...)
```

So one org deleting a shared client profile **silently destroys every other org's notes for that client**. (`Task.client_profile` is `SET_NULL`, so tasks survive orphaned instead.)

Options:

1. **Restrict delete to org admins** (the original proposal) — edits stay shared, deletion needs elevated rights.
2. **Keep delete global** but make the cascade non-destructive (soft-delete the profile, or `PROTECT` when other orgs hold notes).
3. **Keep as-is** — accept that any caseworker can remove a shared client and its cross-org history.

Given every other write in this model is org-owned, option 3 looks inconsistent with the rest of the matrix.

## Next steps

1. Decide the delete question above.
2. Encode the decided matrix: annotate `notes/groups.py` / `shelters/groups.py` marking which template perms are the shared layer and why (overlaps #2313).
3. Rename/annotate the ~35 codified cross-org permission tests from accidental ("CM in different org should succeed") to deliberate ("shared client data — cross-org visible"), and add explicit tests for the org-private parts (note private details, note/task writes, teams).

## Status

Decisions 1–3 answered; delete question open. No behavior change expected for 1–3 — this is now a codify-and-test PR, and could reasonably merge with #2313.


vecchp:
Consolidated into #2313 — the decisions here are answered, and both PRs reduced to the same work (write the layer model down, annotate the templates, rename/extend the cross-org tests).

**Decisions carried over:**

1. Client profiles are **global/shared** for read and write. Cross-org **delete** stays as-is too — clients are shared and there is no ownership concept, and we are not changing app behavior now. The cascade consequence (deleting a shared profile also deletes other orgs' notes, since `Note.client_profile` is `CASCADE`) is recorded on #2313 as an accepted, known risk.
2. Tasks are **org-specific** for writes — another org cannot complete or reassign them. Reads stay shared as read-only history, provided the UI marks the owning org and hides edit affordances; hiding cross-org tasks is the fallback.
3. Client documents are **shared-read**.

Closing in favour of #2313.

PR #2315, opened 2026-08-13
refactor(backend): header-driven org for remaining first-match permission resolution

vecchp:
## Goal

Finish migrating mutations off first-match org resolution (`resolve_permission_group(user, template=...)` without an org id), which can pick the wrong org for multi-org users.

## Remaining call sites

- `clients/schema.py` — `create_client_document`
- `referrals/schema.py` — referral mutations
- `hmis/schema.py` — `create_hmis_note_service_request`
- `teams/schema.py` — `teams` query fallback when the `X-Organization-ID` header is missing

## Pattern (from notes/tasks in #2310)

`HasOrgPerm` + `get_current_organization(info)` + `permission_group_for_user(user, org_id=..., template_name=...)`.

For the `teams` query fallback: remove it once the mobile app reliably sends the org header on every request.

## Extra detail on the `teams` query fallback

Found while verifying org-scoped team management for #2310:

- The fallback calls `resolve_permission_group(user, template=CASEWORKER)`, so it only works for users who hold a **Caseworker** group. An **Org Admin who is not also a caseworker** — the exact persona the admin app's Teams page serves — gets `PermissionError: User does not hold a 'Caseworker' permission group in any organization` instead of their teams, whenever the org header is missing.
- The header is only injected when an active org is stored (`libs/fetch/src/lib/fetchInterceptors.ts`), so "no header" is reachable on a fresh session, not hypothetical.
- Until the fallback goes, "orgs can only see their own teams" is not *enforced* for multi-org users: first-match resolution picks whichever org the database returns first.

The mutations (`createTeam`/`updateTeam`/`deleteTeam`) are already header-driven and org-scoped, with tests in `teams/tests/test_mutations.py`; the query fallback is the last piece.

## Next steps

1. Port each mutation to the header-driven pattern.
2. Update/extend multi-org regression tests per app.
3. Drop the `teams` query fallback and assert the header is required.

## Status

Not started.

## Summary by Sourcery

Add an operational command to audit and synchronize organization permission groups with their templates.

New Features:
- Add a management command to reconcile organization permission groups with their permission templates, with a read-only check mode.

Enhancements:
- Provide operational verification and repair for permission-group drift across organizations.

Tests:
- Add coverage for detecting permission mismatches, preserving check-mode behavior, repairing drift, and confirming synchronization.


vecchp:
Closing as a duplicate of #2313 — the two diffs were byte-identical (the same stray `sync_org_permission_groups` command, tests, and no-op file move), and neither matched its title.

The content that mattered is preserved in `apps/betterangels-backend/docs/permissions.md` via #2310, as an explicit table of the call sites still resolving the organization by first match and what each one costs: client-document ownership on upload, `create_referral`, and `create_hmis_note_service_request`.

The two items this PR listed that are now done: the `teams` query fallback is fixed in #2345 — including the org-admin-who-is-not-a-caseworker case called out here — and note/task writes moved to the header-driven pattern in #2310.

PR #2309, opened 2026-08-13
feat(teams): per-org legacy team backfill migration

vecchp:
## What

- `teams.0002_backfill_org_teams`: idempotent data migration that creates per-org `Team` rows from legacy `old_team` values on notes/tasks and backfills the FKs. Data-driven only — no default teams are seeded for new orgs.
- `docs/teams_org_scoping.md`: deploy runbook (audit → migrate → sync perms → audit) + permission model + follow-ups.

## Deploy notes

- Run `manage.py audit_team_org_scoping` before/after deploying (command ships in a later layer).
- Migration is safe to re-run and reverses to a no-op.
- New orgs start with zero teams; admins create them in the UI (Teams page). `load_report_test_data` still seeds SELAH teams for local fixtures.

## Stack

Layer 2 of 5.

PR #2151, opened 2026-06-19
feat(backend): dynamic teams with per-org scoping (slug + name)

vecchp:
## Backend: Dynamic Teams

Based on `sdb-178-pr4-self-signup-complete`.

### Changes
- **Teams model** with slug/name/isActive, migrations, admin, permissions
- **GraphQL schema**: Offset-paginated teams query, CRUD mutations
- **team_shim** helper, TeamPermissions enum registered as strawberry enum
- **Notes/Tasks** updated with team FK, migrations, schema, types
- **Accounts**: team permissions groups, signals, types
- **Settings**: x-organization-id CORS header
- **Reports**: test data seeding for teams
- Removed `active_only` filter — always returns all teams
- Regenerated schema.graphql

### Files Changed
40 files under `apps/betterangels-backend/`

PR #2317, opened 2026-08-13
feat(teams): per-org legacy team backfill migration

vecchp:
## What

- `teams.0002_backfill_org_teams`: idempotent data migration that creates per-org `Team` rows from legacy `old_team` values on notes/tasks and backfills the FKs. Data-driven only — no default teams are seeded for new orgs.
- `docs/teams_org_scoping.md`: deploy runbook (audit → migrate → sync perms → audit) + permission model + follow-ups.

## Deploy notes

- Run `manage.py audit_team_org_scoping` before/after deploying (command ships in a later layer).
- Migration is safe to re-run and reverses to a no-op.
- New orgs start with zero teams; admins create them in the UI (Teams page). `load_report_test_data` still seeds SELAH teams for local fixtures.

## Stack

Layer 1 of 4; supersedes closed PR #2309.

## Summary by Sourcery

Backfill legacy team assignments into organization-scoped Team records while documenting the rollout and permission model.

Bug Fixes:
- Backfill missing organization-scoped Team records and associate legacy note/task team references with them.

Enhancements:
- Document the deployment runbook, permission model, and remaining follow-ups for organization-scoped teams.
- Keep test fixtures self-contained by creating legacy-slug teams for each test organization instead of relying on production enum definitions.

Documentation:
- Add user-facing deployment and operational guidance for auditing, migrating, synchronizing permissions, and validating organization-scoped teams.


vecchp:
Closing — the migration this PR existed for turned out to be a no-op, and its remaining content has moved into #2310.

## Why the migration went

`teams.0002_backfill_org_teams` created per-org `Team` rows from `old_team` values. Production shows it would do nothing:

- `teams_team` already holds **46 rows** — the 15 legacy enum teams for each of **three** organizations, created by #2151 in June (three timestamp clusters seconds apart), plus one team an admin created through the Teams UI on 26 June.
- Every row this migration would `get_or_create` already exists.

It also could not become necessary later: **nothing in application code writes `old_team`**, so no row can acquire one and the candidate set has been frozen since June. Keeping it would mean a permanent migration in the graph that duplicates finished work, carries a hardcoded map of one customer's team names, and would seed those names into any org it ever fired for.

## What moved to #2310

The runbook (`docs/teams_org_scoping.md`) and the test-fixture change — the latter is required, since it removes the `SelahTeamEnum` import that #2312 deletes. #2310 is now based on `main`, so the stack is three layers instead of four and one fewer approval.

## What the runbook keeps

The pre-deploy checks, which matter **more** without a backfill, because the `old_team` drops are irreversible:

```sql
SELECT count(*) FROM notes_note  WHERE old_team IS NOT NULL AND team_id IS NULL;
SELECT count(*) FROM tasks_task  WHERE old_team IS NOT NULL AND team_id IS NULL;
SELECT count(*) FROM notes_noteevent WHERE old_team IS NOT NULL AND team_id IS NULL;
SELECT count(*) FROM tasks_taskevent WHERE old_team IS NOT NULL AND team_id IS NULL;
```

The event-table variants matter because June's migration updated live rows only — pre-June history may carry `old_team` with no `team_id`, and dropping the column removes the only team attribution on it.

PR #2321, opened 2026-08-14
feat(shelter-operator): link the printable report from the Reports tab [3/3]

julianchen24:
﻿**Stack 3 of 3** â€” builds on #2320.

## Problem

The printable report exists but nothing links to it, so it's reachable only by typing a URL. For operators, that means it doesn't exist.

## Solution

An entry point on the Reports tab, which is the live reporting surface today.

The original plan put this on the shelter dashboard, but that page is no longer routed on `main` â€” the manage root redirects to beds and nothing renders it. If reviving it is intended work, this PR deliberately leaves it alone.

## How to test

Open a shelter â†’ Manage â†’ Reports â†’ **Printable Report**, and confirm it lands on that shelter's report.

## Summary by Sourcery

Link the shelter Reports tab to its printable report for the current shelter.

New Features:
- Add a Printable Report entry point to the shelter Reports tab.

Tests:
- Add coverage for printable report navigation, shelter ID propagation, and button semantics.
- Add coverage ensuring printable report route generation stays aligned with the registered route.

### DIFF
--- a/libs/react/shelter-operator/src/lib/pages/shelterManagement/ReportsPage.test.tsx
+++ b/libs/react/shelter-operator/src/lib/pages/shelterManagement/ReportsPage.test.tsx
@@ -1,0 +1,82 @@
+import { fireEvent, render, screen } from '@testing-library/react';
+import {
+  MemoryRouter,
+  Route,
+  Routes,
+  useLocation,
+  useParams,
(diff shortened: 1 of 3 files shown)


Anahat07 (approved):
Looks good to me!

PR #2328, opened 2026-08-17
fix(permissions): require the org header instead of returning "None"

vecchp:
## The problem

`get_current_organization` returned `str(request.organization_id)`, so a missing `X-Organization-ID` header produced the **literal string `"None"`** — never a valid organization id. The signature promised an org id and delivered a sentinel:

```python
def get_current_organization(info: Info) -> str:
    return str(info.context.request.organization_id)   # -> "None"
```

Every consumer feeds that value straight into a queryset filter, so the request **already always failed**. It just failed downstream, with:

```
Field 'id' expected a number but got 'None'
```

which says nothing about a missing header.

## Scope — deliberately narrow

This does **not** change whether anything fails, only where and with what message. Two groups of callers:

- **Behind `@HasOrgPerm`** — never reach this. The extension rejects a headerless request first, with the same message. No change.
- **Not gated** — the type-level `get_queryset` hooks in `shelters/types/outputs.py` (`ShelterType`, `BedType`, `RoomType`, `ReservationType`) and the `_room_beds_prefetch` / `_reservation_clients_prefetch` helpers. These are where the cryptic error actually surfaces today, and where this change is visible.

So the practical effect is **better diagnostics on shelters read paths when the active-org header is absent**.

## Why `PermissionError`

Matches how the rest of the codebase reports permission failures in query paths — `clients/schema.py`, `accounts/schema.py`, and `accounts/selectors.py` all raise the builtin. There is no error masking configured (`Schema(extensions=[DjangoOptimizerExtension])`), so the message reaches the client rather than being swallowed.

Mutations use Django's `PermissionDenied` instead, because strawberry-django converts that into a typed `OperationInfo` the client can render. That distinction is intentional and left alone here.

## Tests

`common/tests/test_org_context.py` — returns the id as a string for int and str inputs, raises naming the header when absent, and never returns `"None"`.

## Verification

`schema.graphql` unchanged, `ruff` clean, `mypy` clean on 367 files with the cache cleared.

## Note for review

Worth a look from someone who knows the shelters read paths: if any of those ungated `get_queryset` hooks is currently reached without the header in a flow that *should* keep working, this converts a confusing error into an explicit one — which is still a failure, just an honest one. I did not find such a flow, but I know teams better than shelters.

## Summary by Sourcery

Improve missing-organization diagnostics while preserving public shelter queries that do not require an active organization.

Bug Fixes:
- Require the X-Organization-ID header when resolving the active organization instead of returning the invalid string "None".
- Preserve unauthenticated, headerless access to public shelter directory queries.

Enhancements:
- Add focused coverage for organization ID conversion and missing-header errors.

Tests:
- Add regression tests confirming public shelter list and detail queries work without authentication or an organization header.
- Add unit tests for active organization resolution with numeric, string, and absent header values.

### DIFF
--- a/apps/betterangels-backend/common/permissions/utils.py
+++ b/apps/betterangels-backend/common/permissions/utils.py
@@ -6,6 +6,7 @@ from typing import Any, Sequence, Tuple, Type, TypeVar
 
 import strawberry
 from django.contrib.auth.models import AbstractBaseUser, Group
+from django.core.exceptions import PermissionDenied
 from django.db.models import Exists, Model, OuterRef, Q, QuerySet, TextChoices
 from django.utils.encoding import force_str
 from guardian.shortcuts import assign_perm
@@ -232,12 +233,24 @@ def get_current_organization(info: Info) -> str:
     method needs the active organization for selector/service calls.
 
     Raises:
-        ``AttributeError`` if the request does not have ``organization_id``
-        set.  This only happens when a field/mutation uses ``@HasOrgPerm``
-        without the middleware being installed — which is a configuration
-        error.
+        ``PermissionDenied`` if the header is absent.  Returning ``str(None)``
+        — the literal ``"None"`` — used to push the failure downstream into a
+        queryset filter, where it surfaced as ``Field 'id' expected a number
+        but got 'None'``.  Fields behind ``@HasOrgPerm`` never get here (that
+        extension rejects the request first), but type-level ``get_queryset``
+        hooks are not gated and can.  Django's ``PermissionDenied`` rather than
+        the builtin ``PermissionError`` because it is one of the three
+        exceptions strawberry-django turns into ``OperationInfo``, so a
+        mutation reports it as a message rather than an unstructured error.
+        ``AttributeError`` if the middleware is not installed, which is a
+        configuration error.
     """
-    return str(info.context.request.organization_id)
+    org_id = info.context.request.organization_id
+
+    if org_id is None:
+        raise PermissionDenied("Organization ID (X-Organization-ID header) is required.")
+
+    return str(org_id)
 
 
 _T = TypeVar("_T", bound=Model)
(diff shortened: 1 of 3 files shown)


vecchp:
Reviewing this alongside #2310, which touches the same twenty lines of `common/permissions/utils.py`. Two things.

## `PermissionError` is the Python builtin, not Django's

```python
raise PermissionError("Organization ID (X-Organization-ID header) is required.")
```

That's `builtins.PermissionError`, an `OSError` subclass — not `django.core.exceptions.PermissionDenied`. strawberry-django only converts three exception types into `OperationInfo` ([`strawberry_django/mutations/fields.py:100`](https://github.com/strawberry-graphql/strawberry-django/blob/main/strawberry_django/mutations/fields.py)):

```python
if isinstance(error, (ValidationError, PermissionDenied, ObjectDoesNotExist)):
```

and the builtin is not one of them:

```
>>> issubclass(PermissionError, django.core.exceptions.PermissionDenied)
False
>>> PermissionError.__mro__
(PermissionError, OSError, Exception, BaseException, object)
```

So a request that hits this raises an unhandled error and surfaces as a 500 — the same outcome the PR is removing, with a better message attached. `from django.core.exceptions import PermissionDenied` and raising that gets the typed denial.

Worth saying why the tests don't catch it: `test_org_context.py` calls `get_current_organization` directly and asserts `assertRaises(PermissionError)`, which passes regardless, because nothing exercises the GraphQL error path. An end-to-end case on a resolver that reaches `get_current_organization` without `@HasOrgPerm` in front of it — the `get_queryset` hooks your docstring names — would.

## Sequencing with #2310

These conflict, and I think this one should land first.

#2310 added `require_organization_id` immediately below `get_current_organization`, for use *inside* permission extensions, where the denial has to be `DjangoNoPermission` to be caught by the framework. It also validates that the header is a usable primary key, not just non-`None` — a malformed header (`X-Organization-ID: abc`) otherwise reaches the ORM and raises `ValueError: Field 'id' expected a number but got 'abc'` from inside the query.

Once this merges, the two are complementary: `require_organization_id` for extensions, `get_current_organization` for resolver bodies, both refusing rather than passing a bad value downstream. #2310 currently carries a docstring caveat pointing at this PR, which I would then delete. It would also be worth giving `get_current_organization` the same primary-key validation, so the two helpers differ only in exception type rather than in how much they check.

Happy to do either half if it's easier.

mikefeldberg (approved):

PR #2329, opened 2026-08-17
deactivate team be (DEV-2503)

mikefeldberg:
DEV-2503

## Summary by Sourcery

Support team lifecycle management by adding an active status that can be queried, filtered, and updated.

New Features:
- Add support for deactivating teams and filtering teams by active status through the GraphQL API.

Enhancements:
- Expose team active status and support partial updates to team names and activation state.
- Make team listing paginated and organization-scoped with configurable filters.

Tests:
- Add GraphQL coverage for team creation, updates, deletion, organization scoping, and active-status filtering.


apps/betterangels-backend/teams/services.py
```diff
@@ -34,19 +36,24 @@ def team_create(
...
+
+        slug = slugify(name)
         if Team.objects.filter(slug=slug, organization=team.organization).exclude(pk=team.pk).exists():
             raise ValidationError(f'A team with slug "{slug}" already exists in this organization.')
+
         team.name = name
         team.slug = slug
 
+    if data["is_active"] is not None:
+        team.is_active = data["is_active"]
```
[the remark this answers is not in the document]
mikefeldberg: this is handled by the caller with asdict

apps/betterangels-backend/teams/tests/test_queries.py
```diff
@@ -0,0 +1,55 @@
...
+    def test_teams_query(self) -> None:
+
+        expected_query_count = 4
+        with self.assertNumQueriesWithoutCache(expected_query_count):
+            response = self.execute_graphql(self.get_teams_query())
+
+        self.assertIsNone(response.get("errors"))
+        self.assertEqual(response["data"]["teams"]["totalCount"], 1)
+
+        team = response["data"]["teams"]["results"][0]
```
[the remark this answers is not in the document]
mikefeldberg: don't need to test a simple bool filter

apps/betterangels-backend/teams/types.py
```diff
@@ -23,3 +31,4 @@ class CreateTeamInput:
 class UpdateTeamInput:
     id: ID
     name: Maybe[str]
+    is_active: Maybe[bool] = UNSET
```
vecchp: Drop the `= UNSET`. Strawberry already injects the absence default for a `Maybe` field, and setting one by hand overrides it — `strawberry/types/object_type.py:115`:

```python
def _inject_default_for_maybe_annotations(cls, annotations):
    """Inject `= None` for fields with `Maybe` annotations and no default value."""
```

The result is that the two fields in this input report absence differently. Paste this into a REPL to see it — no Django needed:

```python
import dataclasses
import strawberry
from strawberry import ID, UNSET, Maybe
from strawberry.types.maybe import Some


@strawberry.input
class UpdateTeamInput:
    id: ID
    name: Maybe[str]              # no explicit default
    is_active: Maybe[bool] = UNSET


print("defaults Strawberry ended up with:")
for f in dataclasses.fields(UpdateTeamInput):
    default = "<required>" if f.default is dataclasses.MISSING else repr(f.default)
    print(f"  {f.name:10} default={default}")

omitted = UpdateTeamInput(id="1")
print(f"\nclient omits both fields:  name={omitted.name!r}  is_active={omitted.is_active!r}")

rows = [("absent (name)", None), ("absent (is_active)", UNSET), ("Some(False)", Some(False)),
        ("Some('')", Some("")), ("Some('Night Team')", Some("Night Team"))]
print(f"\n{'state':<22}{'is not None':>12}{'is not UNSET':>14}{'bool(f)':>9}{'bool(f.value)':>15}")
for label, f in rows:
    inner = str(bool(f.value)) if isinstance(f, Some) else "—"
    print(f"{label:<22}{str(f is not None):>12}{str(f is not UNSET):>14}{str(bool(f)):>9}{inner:>15}")
```

```
defaults Strawberry ended up with:
  id         default=<required>
  name       default=None
  is_active  default=UNSET

client omits both fields:  name=None  is_active=UNSET

state                  is not None  is not UNSET  bool(f)  bool(f.value)
absent (name)                False          True    False              —
absent (is_active)            True         False    False              —
Some(False)                   True          True     True          False
Some('')                      True          True     True          False
Some('Night Team')            True          True     True           True
```

Row 2 is the hazard: `data.is_active is not None` evaluates to **True** for an omitted field, so the obvious presence check reads it as provided and assigns `UNSET` into a `BooleanField`. The current `asdict` path hides this — both spellings get filtered out — so it surfaces the first time anyone reads the input directly.

The table also shows the read idiom to use: check the field, then unwrap. `Some.__bool__` always returns `True`, so `bool(f)` is the only column that means "the client sent this field" in every row. `is not None` breaks on `is_active`, `is not UNSET` breaks on `name`, and truthiness on the unwrapped **value** breaks on `Some(False)` — the payload this PR ships.

```python
if data.is_active:                 # the client sent this field
    value = data.is_active.value   # the value itself
```

One related gotcha: a bare `Maybe[T]` rejects an explicit `null` at validation — `"Field 'name' of type 'Maybe[str]' cannot be explicitly set to null. Use 'Maybe[str | None]'"`. `Maybe[T]` and `Maybe[T | None]` emit identical nullable SDL, so clients can't distinguish them by introspection. That's correct for these two non-nullable columns, but `null` won't arrive as `Some(None)` unless the annotation says `| None`. `shelters/types/inputs.py` draws the line the same way — `email: Maybe[Optional[str]]` and `city_id: Maybe[ID | None]` for nullable columns and FKs, `is_private: Maybe[bool]` for a flag. Those fields have the same `= UNSET` override and could use the same cleanup in a follow-up.

apps/betterangels-backend/teams/schema.py
```diff
@@ -60,12 +59,11 @@ def update_team(self, info: Info, data: UpdateTeamInput) -> TeamType:
         if team is None:
             raise ValueError(f"Team with id {data.id} not found.")
 
+        clean = strawberry.asdict(data)
```
vecchp: **Blocking.** `updateTeam` raises `KeyError` on every single-field update, including deactivation. Against this branch:

```
{ id, isActive: false }   → {'data': None, 'errors': [{'message': "'name'"}]}
{ id, name: "new name" }  → {'data': None, 'errors': [{'message': "'is_active'"}]}
```

`asdict` doesn't null out absent `Maybe` fields, it deletes the keys — `_prepare()` maps an absent `Maybe` to `UNSET`, then filters `UNSET` out of the dict:

```
only isActive sent → {'id': '1', 'is_active': False}    → services.py:42 data['name'] KeyError('name')
only name sent     → {'id': '1', 'name': 'new name'}    → services.py:54 data['is_active'] KeyError('is_active')
both sent          → {'id': '1', 'name': 'new name', 'is_active': False}
```

Because `team_update` takes `data: dict[str, Any]`, mypy can't flag the unsafe access, so both a full test run and `mypy teams/` pass clean. The only request shape that works is sending both fields, which is what `test_update_team_mutation` does.

`strawberry.asdict` is also on the way out of this codebase rather than in. Suggested shape — unwrap in the resolver, pass the service typed kwargs:

```python
# schema.py
return cast(TeamType, team_update(
    team=team,
    name=data.name.value if data.name else None,
    is_active=data.is_active.value if data.is_active else None,
))

# services.py
def team_update(*, team: Team, name: str | None = None, is_active: bool | None = None) -> Team:
```

I ran all five shapes through that signature: `only isActive → is_active=False`, `only name → name and slug updated`, `both`, `neither → no-op`, `punctuation-only → ValidationError`. Switching the two lookups to `data.get(...)` also stops the crash, but keeps mypy blind to the boundary.

`team_create` in this module already takes explicit kwargs, so the two services currently disagree on their calling convention. `ShelterPhotoResolveItem` in `shelters/services/shelter_photo.py:32-44` documents the same reasoning — a dict boundary means "losing type safety and risking `KeyError`".

For context on severity: this mutation is already broken on `main` in a different way — `maybe_value` returns the `Some` wrapper instead of the value, so rename fails with `'Some' object has no attribute 'strip'`. This PR fixes that and introduces the `KeyError`, so `updateTeam` has never worked for a single-field update. A test per field in isolation would have caught both.

On why `asdict` keeps coming back: `Maybe` reads are verbose, `asdict` flattens them in one line, and `shelter_create(*, data: Dict[str, Any])` reads as a template with a docstring that endorses the dict. A Ruff `flake8-tidy-imports` `banned-api` entry on `strawberry.asdict`, with per-file ignores for the ~20 existing call sites, would keep the count moving one direction. I can open that as a separate ticket.

apps/betterangels-backend/teams/services.py
```diff
@@ -34,19 +36,24 @@ def team_create(
...
-    name: str | None = None,
+    data: dict[str, Any],
 ) -> Team:
-    """Update a Team's name. Slug is auto-generated from name."""
-    if name is not None:
-        name = name.strip()
-        slug = slugify(name)
-        if not slug:
+    """Update a Team's name and/or active flag. Slug is auto-generated from name."""
+    if data["name"] is not None:
```
vecchp: These two lookups (`data["name"]` here, `data["is_active"]` at :54) are what raise — `asdict` deletes absent `Maybe` keys rather than nulling them. Details and suggested fix in the comment on `schema.py:62`.

apps/betterangels-backend/teams/services.py
```diff
@@ -34,19 +36,24 @@ def team_create(
...
+        name = data["name"].strip()
+        if not name:
             raise ValidationError("Team name must contain at least one alphanumeric character.")
+
+        slug = slugify(name)
```
vecchp: **Blocking.** Reordering the validation dropped the empty-slug guard, so punctuation-only names now save `slug=""`. The check moved from slugify's output to the stripped name, and `slugify("---") == ""`. On this branch:

```
rename "alpha" → "---"   ⇒ succeeds; row saved with name='---' slug=''
rename "beta"  → "!!!"   ⇒ VALIDATION: 'A team with slug "" already exists in this organization.'
```

Two consequences: the first team silently gets an empty slug, which is still the migration-shim lookup key per the `Team` docstring; and the second rename fails against `unique_team_slug_per_org` with a message that won't mean anything to the user.

`team_create` still guards on slugify's output, so create and update now disagree on what a valid name is. Guarding the slug also keeps the error message accurate:

```python
name = name.strip()
slug = slugify(name)
if not slug:
    raise ValidationError("Team name must contain at least one alphanumeric character.")
```

apps/betterangels-backend/teams/tests/test_queries.py
```diff
@@ -0,0 +1,55 @@
...
+        self.assertEqual(team["id"], str(self.team.pk))
+        self.assertEqual(team["name"], self.team.name)
+        self.assertTrue(team["isActive"])
+
+    def test_teams_query_is_scoped_to_org(self) -> None:
```
vecchp, line 34: Three cases worth adding, each of which fails on this branch: update with only `isActive`, update with only `name`, and a punctuation-only name.

The PR description also lists filtering under coverage, but there's no test for `isActive` filtering. The filter works — `teams(filters: { isActive: { exact: true } })` returned 1 of 2 teams when I ran it — but it's new public API and currently untested.

vecchp (commented):
Reviewed the backend changes and ran the suite locally against this branch (`1313 passed`, schema in sync, `mypy teams/` clean). Two blocking issues that the current tests don't reach, both left inline.

Summary:

- **`updateTeam` raises `KeyError` on any single-field update**, so deactivating a team — the point of the PR — fails. Only sending `name` *and* `isActive` together works, which is what the new test does. Inline on `teams/schema.py:62`.
- **Punctuation-only renames now save `slug=""`** because the empty-slug guard moved off slugify's output. Inline on `teams/services.py:44`.
- `= UNSET` on `is_active` overrides the default Strawberry injects for `Maybe` fields and breaks the `is not None` presence check. Inline on `teams/types.py:34`.
- Missing coverage for the exact gaps that hid both bugs, plus the `isActive` filter the description lists as covered. Inline on `teams/tests/test_queries.py`.

Two minor points not worth their own threads:

- `common/team_shim.maybe_value` is now used only inside its own module, and its docstring says it extracts the value from a `Maybe` when it actually returns the `Some` wrapper. Worth making private or dropping with the shim.
- `team_list` orders by `slug` while `Meta.ordering = ["name"]`, and returns inactive teams — so `useOrgTeams` pickers will show deactivated teams until the client passes the filter. Assuming that's the frontend ticket, just flagging it's tracked somewhere.

vecchp (approved):

PR #2330, opened 2026-08-17
fix(platform): hold the active org outside React so the request header can't lag

vecchp:
## The problem

The `X-Organization-ID` header is attached by a **fetch interceptor** — not a component, so it cannot read React state.

The active org was React state, mirrored into storage by an effect, and the interceptor read *that storage*. So it read a copy that lagged the UI:

- **React Native** — lagged by an `AsyncStorage` round trip.
- **Web** — lagged by one commit. `localStorage` is synchronous, but the write was in an effect, and React runs effects **child-before-parent**, so a child's query fires first.

Either way, requests went out header-less while the UI already showed an organization.

## The fix

The id lives in **one place, outside React**: a module store, written synchronously, read synchronously by both the interceptor and React via `useSyncExternalStore` — the API React documents for integrating non-React code, and already used elsewhere in this codebase. Storage becomes persistence, not a channel between two halves of the app.

Two details make it airtight rather than merely better:

**The reconcile happens during render, not in an effect.** An effect cannot work: child effects run first, so a child would query before the provider had chosen an org. It is safe as a render-phase write because it is idempotent and derived purely from props — a discarded render can only publish what a committed render would.

**MMKV backs it on React Native.** The value must be readable synchronously — during `getSnapshot` and on every request — which `AsyncStorage` cannot do. MMKV is not a new dependency: it has been in `package.json` since #1777 (January) and already backs `userPreferencesState`. `package.json` is untouched here.

## Who installs the platform's storage

The store lives in the platform-agnostic package and cannot reach `localStorage` or MMKV itself, so the implementation has to be handed in. **The platform fetch clients do it** — in the same file where they already import their own CSRF reader:

```ts
// libs/ba-platform/web/src/lib/fetchClient.ts
configureActiveOrgStorage(webActiveOrgStorage);
```

`configureActiveOrgStorage` does no I/O: it records the backend and marks the seed stale, and the first read pulls. That makes the call order-independent, so there is no "install before anything issues a request" contract for a caller to get wrong — and re-entering it on an environment switch cannot resurrect an org that was just cleared.

**Neither app entry point changes.** `apps/betterangels-admin/src/main.tsx` and `apps/betterangels/src/init.ts` are identical to `main`.

## What this deletes

The `storage` / `defaultStorage` prop chain that existed only to feed this, threaded from four apps through `createUserProvider` into `ActiveOrgProvider`, along with the `UserProvider` runtime check that threw when neither was supplied. Also `StorageReader` in `libs/fetch` and `localStorageAdapter` in `libs/react/shared` — both left without consumers.

## Bugs fixed on the way

**Restoring your last organization never worked on mobile.** `useActiveOrgState` called `storage.getItem()` synchronously behind an `as string | null` cast. On Expo that returns a **Promise**, so the restore branch never matched and mobile users always landed on their first org. Synchronous storage makes the cast unnecessary.

**Nothing cleared the org on logout or environment switch.** A stale org survived a user switch — the org-list check caught it, but only *after* a header had been sent — and an id from one API environment was sent against another. All three sign-out paths and the environment switcher now clear explicitly. shelter-web was the easiest to miss: it builds a web fetch client and `OperatorApp` renders `ActiveOrgProvider`, so it writes an id like the others, and its next user would have inherited it.

## Behaviour change to be aware of

Two browser tabs now keep **independent** active organizations rather than following each other. That is deliberate — a tab's UI and its request headers always agree now, where previously switching in one tab changed the other's headers without changing what it displayed. Nobody asked for this; say so if you would rather have cross-tab following, which is a `storage` event listener in the store.

## Migration note

Values previously in `AsyncStorage` under `betterangels_active_org_id` are not migrated. Mobile users land on their first org once, and the choice sticks from then on.

`DEFAULT_ORG_STORAGE_KEY` is renamed `ACTIVE_ORG_STORAGE_KEY` — the prefix implied a per-app override that never existed. The stored key string is unchanged.

## Test infra

`libs/ba-platform/{web,expo}` vitest configs gain explicit monorepo aliases. `tsconfigPaths` resolves `@monorepo/*` from source but **not** from test files there, so these tests could not import a sibling lib. The repo already ships `monorepoTsconfigAliases` for this exact gap.

## Verification

`nx affected -t lint typecheck test --base=origin/main` — **19 projects, all green.**

New coverage: the store itself; the **render-ordering property** — a child sees the org during its own render, and this test fails if the reconcile moves back into an effect; restoring a remembered org; discarding one the user no longer belongs to; keeping the remembered org through the window where `UserProvider` renders children with an empty org list; the interceptor re-reading per request rather than once at construction; and that the MMKV handle is created on first use rather than at import.

`useOrgTeams.test.tsx` gains tests for a hook this PR does not otherwise touch. That is deliberate: it stands in for any org-scoped query, and those are the cross-project tests that exercise the store, provider and interceptor together — the in-project suite resets the store per test and so cannot catch install-time regressions.

The previous tests asserted the race as expected behaviour, so they are replaced.

**Not yet done:** a device smoke test. Every test mocks `createMMKV`, so the real native path has not executed.

Independent of the backend stack — based on `main`.


libs/expo/betterangels/src/lib/hooks/useOrgTeams/useOrgTeams.test.tsx
```diff
@@ -0,0 +1,118 @@
...
+  data: {
+    teams: {
+      __typename: 'TeamTypeOffsetPaginated',
+      totalCount: 1,
+      results: [{ __typename: 'TeamType', id: 'team-1', name: 'WDI On-site' }],
+    },
+  },
+};
+
+function createSyncStorage(
```
mikefeldberg, line 42: `useOrgTeams.test.tsx` does not call `resetActiveOrgStoreForTests` between tests.

The three tests each call `configureActiveOrgStorage(...)`, which marks the store as unseeded, effectively lazy-resetting the `current` value on the next read. This works by coincidence of the lazy-seed design. A future test author could write a test that reads `current` before the lazy seed fires and observe leakage. Adding `resetActiveOrgStoreForTests` to a `beforeEach` here would make the isolation explicit.

mikefeldberg (approved):

libs/expo/betterangels/src/lib/hooks/useOrgTeams/useOrgTeams.test.tsx
vecchp: Good catch — you're right that isolation held only because each test happens to call `configureActiveOrgStorage` as its first statement. Fixed in `a0fd5548`:

```ts
describe('useOrgTeams', () => {
  // So a test that doesn't install its own storage still starts empty.
  beforeEach(() => configureActiveOrgStorage(createSyncStorage()));
```

That's the whole change — four lines in this file, no production change.

I went down two longer routes first and backed out of both, in case you were considering them. `resetActiveOrgStoreForTests` is deliberately out of the package barrel, so reaching it from this library would have meant publishing a test-only entry point — permanent public API that application code could import just as easily, which defeats the point of keeping it out. And making `configureActiveOrgStorage` null out `current` in addition to the seed flag turned out to be unobservable: every read goes through `getActiveOrgId()`, which re-seeds when `seeded` is false, so no caller can ever see the stale value and no test can be written that fails without it.

So the trap you described — a future test reading the store before the first seed — is closed by the `beforeEach` alone, and the store keeps its current shape.

libs/react/betterangels-admin/src/lib/pages/teams/TeamsPage.tsx
```diff
@@ -48,7 +48,7 @@ const COLUMNS: {
 
 export function TeamsPage(props: IProps) {
   const { className } = props;
-  const { hasPermission } = useActiveOrg();
+  const { hasPermission, isOrgHeaderReady } = useActiveOrg();
   const { showDrawer } = useAppDrawer();
   const { showAlert } = useAlert();
   const [search, setSearch] = useState('');
```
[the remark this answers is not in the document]
vecchp: This review was generated against an early commit and none of it describes the current diff. `isOrgHeaderReady`, `persistedOrgId`, the `skip` on the teams query and the `awaitingOrg` flag were all part of an approach that was abandoned — none of those symbols exist on this branch:

```
$ grep -rn "isOrgHeaderReady" --include=*.ts --include=*.tsx apps libs
(no matches)
```

`libs/react/betterangels-admin/src/lib/pages/teams/TeamsPage.tsx` isn't in the diff at all any more, so there's no readiness flag to align and no permanent spinner — the page is untouched by this PR.

What the branch actually does is hold the active org in a module-level store outside React, seeded lazily on first read, so the interceptor can read it synchronously when a request goes out. There is no readiness signal because there is nothing to wait for.

@sourcery-ai review

PR #2331, opened 2026-08-18
feat(api)!: remove the legacy team enum, add team alongside currentTeam, make teamId clearable

vecchp:
## What

Removes the legacy team enum and everything that existed to bridge it to the `teams.Team` model, adds `team` as the name for the FK-backed output field now that the name is free, and makes `teamId` clearable. **This is the breaking change in the teams stack.**

Gone from the GraphQL surface:

- `team: SelahTeamEnum` inputs on `CreateNoteInput`, `UpdateNoteInput`, `CreateNoteTaskInput`, `ImportNoteDataInput`, `CreateTaskInput`, `UpdateTaskInput`
- `team: SelahTeamEnum @deprecated` output fields on `NoteType` and `TaskType`
- the `teams` filter on `NoteFilter` and `TaskFilter`
- the `SelahTeamEnum` type
- `slug` on `TeamType`

Added on the GraphQL surface:

- `team: TeamType` on `NoteType` and `TaskType`. `currentTeam` remains as a deprecated alias resolving the same FK — see below.

Gone from the backend:

- `SelahTeamEnum` itself, from `common/enums.py`
- the `old_team` columns on `Note` and `Task` (`notes/0003`, `tasks/0003`)
- `Team.slug` (`teams/0003`), with uniqueness moving to a case-insensitive constraint on `(lower(name), organization)`
- `common/team_shim.py` — its two `resolve_team_id*` functions existed only to accept the enum, and its `maybe_value` had no caller once they were gone

`common/graphql/utils.py` gains one helper in their place, `maybe_int_value`, for the `ID`-to-`int` narrowing that GraphQL string IDs need.

`Team.name` is the only identifier now. `slug` existed solely as the key the enum values mapped onto, so it goes with them.

## Client impact

Removing the enum needs no client work: its replacements already shipped, and no frontend document selected `team` or `teams`.

`TeamType.slug` was a different story, and an earlier revision of this description got it wrong. **Every `currentTeam` selection baked into released native builds is `currentTeam { id slug name }`** — 12 of them across 10 documents in `libs/expo/betterangels`, with not one omitting `slug`. GraphQL rejects the whole document on an unknown field, so dropping `slug` broke exactly the requests the `currentTeam` alias was added to protect. `slug` therefore stays too, as a deprecated resolver returning `null` — the `Team.slug` **column** is still dropped in `teams/0003`. Nothing ever read the value and the field was already nullable, so old builds get a `null` their generated types accept.

Both compatibility fields come out together in **#2342** (draft until builds roll over). One test per app now selects the shipped block verbatim, so removing either field fails both.

`team` is additive, **not** a replacement. Shipped app builds select `currentTeam`, so it stays as a deprecated alias resolving the same FK; removing it would break every client in the field. New code uses `team` — the 21 hand-written frontend files here (10 `.graphql` documents, 11 components) now select it — and `currentTeam` is deleted in a later release once nothing deployed selects it. `test_note_query_deprecated_current_team_still_resolves` and its task counterpart pin the alias so it cannot go by accident.

`libs/ba-platform`, `shelter-web` and `betterangels-admin` referenced neither name — this is Expo-only.

## Migration notes

`teams/0003` dedupes team names before adding the new constraint. It has to: the old constraint was on `slug`, and the admin exposed `slug` as editable, so an organization could hold both "Hollywood Outreach" and "hollywood outreach". Blank names get a `Team {pk}` placeholder, then case-insensitive duplicates get a numeric suffix, oldest keeping the name. The step is idempotent and cannot fail on real data.

The `old_team` drops are irreversible — they reverse to empty columns, not to the values. Nothing in application code has written `old_team` since the June migration, so the set of rows carrying one has been frozen since then.

## Bug fixed: an update that omitted `teamId` cleared the team

`updateNote` and `updateTask` silently wiped a record's team whenever the client didn't mention `teamId`. Reproduced before fixing — create a task with a team, update only its summary, and `team_id` came back `None`.

Both mutations assigned the resolved team unconditionally, which turned "not mentioned" into "set to null". They now assign only when the field is actually present:

```python
clean = asdict(data)
if data.team_id:
    clean["team_id"] = maybe_int_value(data.team_id)
```

The presence check reads the typed `Maybe[ID]` attribute rather than a value pulled back out of the dict, so the narrowing needs no `Any` annotation. It is also what makes #2316 a one-line change: annotating `teamId` as `Maybe[ID | None]` makes `Some(None)` reachable, and `maybe_int_value` already maps that to a cleared team.

These two resolvers still build their update dict with `strawberry.asdict`, which is the pattern the backend is moving away from (#2335). Replacing it here needs an `UNCHANGED` sentinel — `Task.description` and `Task.team` are nullable, so a plain `None` default cannot separate "absent" from "set to null" — which is more machinery than belongs in this PR. The surrounding comments no longer present `asdict` as the mechanism to reach for.

Both directions are covered by regression tests, mutation-checked: removing the presence check fails exactly those two.

## Team name validation moved onto the model

The alphanumeric rule lived in `team_create` / `team_update`, so it covered the GraphQL path only — `TeamAdmin`'s ModelForm accepted a team named `"---"`, which is the same editable-name path `teams/0003`'s dedupe step exists to clean up after.

It is now a field validator in `teams/validators.py`, declared on `Team.name`, so `full_clean()` runs it for the admin, management commands and the shell as well as the mutations. That is the split the styleguide asks for: content rules on the model, services calling `full_clean()` before `save()`.

`_validate_name` keeps only the uniqueness check. `full_clean()` validates `unique_team_name_per_org` on its own, so that query is there to name the team in the message rather than to catch something that would otherwise escape.

One behaviour change: a whitespace-only name now reports `"This field cannot be blank."` instead of the alphanumeric message, because the service strips it to `""` first and `full_clean()` skips validators for an empty value. The mutation test asserts a message per input rather than one message for both.

## Test coverage for the team field

Deleting the deprecated field initially took its assertions with it — the shared fixtures dropped the enum field and added nothing, so no test selected the FK-backed team anywhere, and the task mutation tests filtered `teamId` out of their expected responses. Both fixtures now select `team` and the create, update and read tests assert it. Measured by breaking `team_id` resolution: **3 tests failed before, 9 fail now.**

`common/tests/test_graphql_utils.py` is new. `maybe_int_value` had no direct coverage, and nothing pinned the `asdict` tri-state behaviour these mutations sit on top of. Writing it surfaced that a bare `Maybe[ID]` cannot hold an explicit null at all, so that case is exercised on a `Maybe[ID | None]` field and the input records why `teamId` cannot currently be cleared (#2316).

## Scope: what's deliberately not here

This branch absorbed the enum removal from #2312 so it lands as one unit. It stops short of that PR's org-enforcement work, which belongs with #2310:

- the composite `(team_id, organization_id)` FKs on notes and tasks, and the `(id, organization)` constraint they target
- the `Note.clean()` / `Task.clean()` methods and `validate_team_in_org`

`teams/validators.py` now exists here, so #2310 adds a function to it rather than creating it. That file must not import `teams.models` at module level any more — `Team` imports it for the field validator, so the reverse import is circular. Flagged on #2310.

Also left for follow-ups:

- **#2334** — enable ruff's import sorting (`I`). Configured but never selected, so import order is unenforced; 130 files, all auto-fixable, so it wants its own PR.
- **#2335** — the wider `Maybe` rework: one absence spelling (drop ~50 `= UNSET`), rename `maybe_int_value` to say what it collapses, convert `task_update` / `note_update` to typed keyword arguments, and add the `full_clean()` the styleguide requires. Noted on #2312, because the `clean()` methods it adds won't run on updates without it.

## From review

- **`team` added alongside `currentTeam`.** `currentTeam` existed only because `team` was occupied by the enum. With the enum gone, `team` matches the model field it resolves and is what new code should select; `currentTeam` keeps working, deprecated, for builds already in the field.
- **Dropped the six `# FK-based team field` comments.** Each contrasted the FK field with the enum field declared directly above it, so they described the migration rather than the code. `UpdateTaskInput` gains the class docstring `UpdateNoteInput` already had, stating the `Maybe` semantics once per input instead of once per field.
- **The shared test fixture uses `baker.make(Team, organization=…)`.** Nothing in the suite asserts a team name literally — only `.name` read off the fixture object — so generated names say as much as invented ones and claim less. Literals are kept only in `teams/tests/test_services.py`, where the name *is* what is under test (creating `Morning Outreach` and rejecting `MORNING OUTREACH` is how the case-insensitive constraint gets exercised).
- **Dropped the SELAH team names from test data.** `SEED_TEAMS` in `load_report_test_data` was the enum's 15 labels verbatim, the last of its lineage in the tree, and the shared GraphQL fixture created `"WDI On-site"` / `"SLCC On-site"`. Both now use fixture-local names. The fixture also binds its teams to `org_1_team_1` / `org_1_team_2`, which let **eleven** `Team.objects.get(name=…)` lookups across six test files be replaced by direct references, and lets the parametrized `teamIds` filter case parametrize over fixture labels instead of name strings.

### Where validation belongs — now written down

The styleguide names HackSoft as its primary reference but is silent on the three things
this stack had to decide, so `docs/styleguides/python.md` gains a short **Validation**
subsection. The layering it records:

| Rule shape | Home |
| --- | --- |
| Single-field content or format | field validator in `<app>/validators.py` |
| Consistency of what is already on the instance | `clean()` |
| Needs a query, or decides on data the caller did not supply | the service |
| Must hold for every writer | `Meta.constraints` |

Two points worth calling out. HackSoft's rule of thumb is "multiple **non-relational**
fields" for `clean()`, but taken literally that reclassifies `Reservation.clean()`, which
compares `self.bed.shelter_id` to `self.room.shelter_id` and is good code — so the line
drawn here is *consistency of what is on the instance* versus *fetching external data to
decide*, which keeps both existing `clean()` implementations valid.

And the last bullet is the load-bearing one: **services must call `full_clean()` before
`save()`**, because without it none of the model-side rules are reachable from the API at
all — only from the admin's ModelForm. `notes` and `tasks` do not call it, which is why
#2335 matters more than it looks; `teams` does.

### Also from review

- **`team_attrs` → `team_labels`** in the parametrized `teamIds` filter test. The file
  already calls a string naming a fixture attribute a *label* (`expected_note_labels`, in
  the same signature); `attrs` named the mechanism and invented a third word for it.
- **`_validate_name` → `_validate_name_is_unique`**, naming the rule the way
  `validate_has_alphanumeric` does. Its docstring loses the paragraph that existed only to
  disclaim what the old name implied.
- **Comments that restated their own code are gone**: a duplicated `get_or_create`
  rationale in `load_report_test_data`, the paragraph in `maybe_int_value`'s docstring
  documenting the alternative that was rejected rather than the function (20 docstring lines
  for a 5-line body), and "Narrow the team ID from GraphQL's str to the FK's int", which
  restated the helper's name in three places.
- **`create_task` loses a no-op guard.** `if data.team_id:` does nothing on a create path —
  `asdict` has already omitted an absent key — so it read as a precaution with no stated
  reason. The two create paths now agree.
- **A docstring that was wrong**: `test_graphql_utils` claimed `maybe_int_value` has "two
  callers" and told them not to build update dicts with it. There are five, three of which
  do exactly that, and `common/graphql/utils.py` documents how.
- **`teams/0003`'s docstring narrowed.** "Cannot fail on real data" is true of the
  *constraint*; the `validate_has_alphanumeric` `AlterField` in the same migration is
  state-only, so a legacy `"---"` name survives and is reported on the next write. No
  normalisation added — nothing shows such a row exists, and that error is the correct
  outcome if one does.
- **Two unused fixture attributes dropped** (`org_2_team_1`, `org_2_team_2`) — assigned and
  never read on this branch.
- **Two test gaps closed**: a nested task carrying `teamId` (that path goes through `asdict`,
  so `task_create` receives the raw string rather than the narrowed int) and the `currentTeam`
  alias on `TaskType`, whose test no longer leans on the unrelated `teamIds` filter. Both
  mutation-checked.

## Bug fixed: the team picker's "none" option failed

The team picker sets `allowSelectNone` and maps an empty selection to `null` (`Team.tsx`), which `buildNotePayload` forwards as `teamId: null`. Every `teamId` input was `Maybe[ID]`, which Strawberry **rejects** rather than reading as "clear":

> Field of type 'Maybe[ID]' cannot be explicitly set to null. Use 'Maybe[ID | None]' if you need to allow null values.

It is raised in `strawberry/types/arguments.py` during argument conversion, so the mutation failed outright. I verified it for **variables**, not just inline literals, since that is how Apollo sends the input.

All six `teamId` inputs are now `Maybe[ID | None]`. Nothing else needed changing: the FK is `null=True, on_delete=SET_NULL`, `maybe_int_value` already collapses `Some(None)` to `None`, and the resolvers' `if data.team_id:` guard passes for `Some(None)` because `Some.__bool__` is always true. Both spellings emit identical SDL (`teamId: ID`), so there is no schema or codegen change and no Inspector entry.

Closes #2316.

## Verification

- **1348 passed, 31 skipped** on a fresh database (`--create-db`, since migrations changed) — eight fewer than the previous revision, which deleted `common/tests/test_graphql_utils.py`
- `mypy` clean over 377 files with `.mypy_cache` deleted
- `ruff check` and `ruff format --check` clean
- `makemigrations --check` clean; migration graph contiguous, and it builds from scratch
- `export_schema` reproduces the committed `schema.graphql` byte-for-byte
- the affected frontend projects pass lint, typecheck and test — codegen-only churn

One unrelated deletion rode along and I kept it: `libs/react/shared/src/lib/apollo/graphql/__generated__/types.ts` (2808 lines, ~87% of this PR's deletions). It is orphaned — that library has no codegen config and `"targets": {}`, its barrel does not export the path, and nothing imports it — and it still referenced `SelahTeamEnum`. #2312 deletes it too, so keeping it deleted avoids re-conflicting. It is the largest file in the diff and has nothing else to do with this change.

## Merge order

**This merges before #2310**, which then carries only the org-enforcement work.

## Summary by Sourcery

Replace the legacy team enum with the dynamic team model across the API and clients while preserving compatibility and correcting team update semantics.

New Features:
- Expose the FK-backed team object as `team` on notes and tasks while retaining `currentTeam` as a deprecated compatibility alias.
- Allow note and task team assignments to be explicitly cleared with `teamId: null`.

Bug Fixes:
- Prevent partial note and task updates that omit `teamId` from unintentionally clearing the existing team.
- Ensure team-picker requests to remove a team are accepted by GraphQL mutations.

Enhancements:
- Remove the legacy `SelahTeamEnum`, enum bridging code, legacy team columns, and team-based enum filters.
- Make `Team.name` the sole identifier with case-insensitive organization-scoped uniqueness and model-level validation.
- Update Expo GraphQL queries and components to use the new `team` field.
- Document backend validation-layer conventions and add coverage for team behavior and compatibility aliases.

Documentation:
- Document validation placement and the requirement for services to run model validation before saving.

Tests:
- Add regression and service tests covering team preservation, explicit clearing, validation, uniqueness, nested assignments, and deprecated alias resolution.

Chores:
- Remove the obsolete generated GraphQL types artifact and update local report/test fixtures to use dynamic team names.


apps/betterangels-backend/notes/types.py
```diff
@@ -174,17 +170,6 @@ class NoteType:
     requested_services: List[ServiceRequestType]
     tasks: list[TaskType]
 
-    # TEMPORARY — @deprecated, use currentTeam instead.
-    # Remove this resolver and the `team` field after the deprecation window.
-    @strawberry_django.field(field_name="team", deprecation_reason="Use currentTeam instead")
-    def team(self, root: models.Note) -> Optional[SelahTeamEnum]:
```
mikefeldberg, line 180: maybe leave team and have it mirror current_team. i.e., `team: Optional[TeamType]`. That way the fe can go back to using `team` instead of `currentTeam`. Unless we want to rename the field in the fe, but I don't think we do. Same comment for `TaskType`

apps/betterangels-backend/teams/validators.py
```diff
@@ -0,0 +1,25 @@
...
+    """Raise unless *value* contains at least one alphanumeric character.
+
+    A team called "---" is indistinguishable from a blank one to anyone reading
+    a list, and *name* is the only identifier a team has.
+
+    Note that ``full_clean()`` skips validators for an empty value, so a name
+    that is blank after stripping is rejected by ``blank=False`` with Django's
+    own message rather than by this rule.
+    """
+    if not any(character.isalnum() for character in value):
```
mikefeldberg, line 24: TIL `isalnum`

apps/betterangels-backend/tasks/types.py
```diff
@@ -140,8 +125,7 @@ class CreateTaskInput:
     note: Optional[ID]
     hmis_note: Optional[ID]
     summary: str
-    team: Optional[SelahTeamEnum]  # TEMPORARY — @deprecated, use teamId. Remove after deprecation window.
-    team_id: Maybe[ID]  # new FK-based field
+    team_id: Maybe[ID]  # FK-based team field
```
mikefeldberg: comment seems unnecessary

apps/betterangels-backend/tasks/types.py
```diff
@@ -150,6 +134,5 @@ class UpdateTaskInput:
     id: ID
     description: auto
     summary: auto
-    team: Optional[SelahTeamEnum]  # TEMPORARY — @deprecated, use teamId. Remove after deprecation window.
-    team_id: Maybe[ID]  # new FK-based field
+    team_id: Maybe[ID]  # FK-based team field; absent = leave unchanged
```
mikefeldberg: comment seems unnecessary

apps/betterangels-backend/notes/types.py
```diff
@@ -247,22 +232,21 @@ class CreateNoteTaskInput:
     summary: str
     description: Optional[str] = None
     status: Optional[int] = None  # Task.Status int choices (0=TO_DO, 1=IN_PROGRESS, 2=COMPLETED)
-    team: Optional[SelahTeamEnum] = None  # TEMPORARY — @deprecated, use teamId. Remove after deprecation window.
-    team_id: Maybe[ID]  # new FK-based field
+    team_id: Maybe[ID]  # FK-based team field
```
mikefeldberg: remove comment (a few more of these below)

apps/betterangels-backend/reports/management/commands/load_report_test_data.py
```diff
@@ -150,25 +150,24 @@
...
-    ("los_feliz_outreach", "Los Feliz Outreach"),
-    ("northeast_hollywood_outreach", "Northeast Hollywood Outreach"),
-    ("selah_staff", "SELAH Staff"),
-    ("silver_lake_outreach", "Silver Lake Outreach"),
-    ("slcc_on_site", "SLCC On-site"),
-    ("sunday_social_atwater_on_site", "Sunday Social / Atwater On-site"),
-    ("sunday_social_atwater_outreach", "Sunday Social / Atwater Outreach"),
-    ("wdi_on_site", "WDI On-site"),
-    ("wdi_outreach", "WDI Outreach"),
+    "Bowtie & Riverside Outreach",
```
mikefeldberg: Do we want to keep using selah team names in test data?

mikefeldberg (approved):

apps/betterangels-backend/notes/types.py
vecchp: Done — went with the straight rename rather than mirroring. There are no deployed clients pinned to `currentTeam` (the documents ship in the Expo JS bundle, which updates with the deploy), so there is no need to carry both names through a deprecation window.

`NoteType` and `TaskType` now declare `team: Optional[TeamType]`, and the `strawberry_django.field(field_name="team")` bridge is gone — the API field name matches the model field again. 21 hand-written frontend files now select `team`; nothing outside `libs/expo/betterangels` referenced it.

apps/betterangels-backend/tasks/types.py
vecchp: Removed — all six of them. Each contrasted the FK field with the `team: Optional[SelahTeamEnum]` line declared directly above it, which this PR deletes, so they were describing the migration rather than the code.

apps/betterangels-backend/tasks/types.py
vecchp: Removed. The `absent = leave unchanged` half was worth keeping, so `UpdateTaskInput` now carries the class docstring `UpdateNoteInput` already had — the `Maybe` semantics stated once per input instead of once per field.

apps/betterangels-backend/notes/types.py
vecchp: All six removed — four here and two in `tasks/types.py`.

apps/betterangels-backend/reports/management/commands/load_report_test_data.py
vecchp: No, and good catch — that list was `SelahTeamEnum`s 15 labels verbatim, the last of its lineage left in the tree. Replaced with six generic names.

The same applied to the shared GraphQL fixture in `common/tests/utils.py`, which was creating `WDI On-site` / `SLCC On-site`. It now creates fixture-local names and binds them to `self.org_1_team_1` … `self.org_2_team_2`, which let eleven `Team.objects.get(name=...)` lookups across six test files become direct references.

apps/betterangels-backend/notes/types.py
vecchp: Correction to my last reply — your original suggestion was right and I went the wrong way first.

I claimed no deployed client was pinned to `currentTeam`. That is not true: `currentTeam` is what every shipped document selects, and it was the legacy `team: SelahTeamEnum` field that had no consumers. Deleting `currentTeam` would have broken every build in the field.

So it is now the mirror you described. `NoteType`/`TaskType` declare `team: Optional[TeamType]`, and `currentTeam` stays as a deprecated alias resolving the same FK:

```python
team: Optional[TeamType]
current_team: Optional[TeamType] = strawberry_django.field(field_name="team", deprecation_reason="Use team instead")
```

Frontend code selects `team` going forward; `currentTeam` is removed in a later release once nothing deployed selects it. `test_note_query_deprecated_current_team_still_resolves` and its task counterpart pin the alias so it cannot go by accident — both fail if the field is deleted.

vecchp:
Two corrections for anyone reading back through this PR.

**Sourcery's reviewer guide is wrong about org scoping.** Its sequence diagram shows `resolve_team_id` doing `Team.objects.get(id=team_id, organization_id=...)`. The deleted function did not do that — when `team_id` was supplied it returned it unchecked, and `organization_id` gated only the enum-slug lookup:

```python
if team_id is not None:
    return team_id
if team_slug is None:
    return None
return Team.objects.get(slug=team_slug, organization_id=organization_id).pk
```

So **this PR removes no org guard**. A caseworker could already set a team from another organization by passing its `teamId`, on `main` and before. That gap is closed by #2339, the next PR up the stack — please do not cite the diagram as evidence #2331 introduced it.

**Where validation belongs is now written down.** `docs/styleguides/python.md` named HackSoft as its primary reference but was silent on field validators, rules needing a query, and rules spanning models — all three of which this stack had to decide. It now has a short `Validation` subsection. The bullet worth knowing about: services must call `full_clean()` before `save()`, because without it no model-side rule is reachable from the API at all, only from the admin's ModelForm. `teams` calls it; `notes` and `tasks` do not, which is #2335.

PR #2333, opened 2026-08-18
deactivate team fe (DEV-2503)

mikefeldberg:
DEV-2503

## Summary by Sourcery

Support deactivating and managing teams consistently across the admin and mobile experiences.

New Features:
- Enable administrators to edit team names and active status from the teams page.
- Add inactive-team visibility controls and inactive labels across team listings and selection fields.

Enhancements:
- Keep inactive teams available when editing existing tasks or notes while excluding them from new team selections by default.
- Centralize team display-name formatting and support active-status filtering in organization team queries.

Chores:
- Add a reusable shared Dropdown component for form controls.


libs/expo/betterangels/src/lib/screens/NoteForm/Team.tsx
```diff
@@ -11,14 +12,20 @@ interface ITeamProps {
 export default function Team(props: ITeamProps) {
   const { teamId, onTeamChange } = props;
   const { teams } = useOrgTeams();
+  const [initialTeamId] = useState(teamId);
```
tglaz: any reason to `useState` if we're not changing state anywhere?
state triggers renders, so should avoid unless necessary.

libs/expo/betterangels/src/lib/ui-components/Filters/FilterTeams/FilterTeamsOptions.tsx
```diff
@@ -21,7 +21,8 @@ export function FilterTeamsOptions(props: TProps) {
   const [localSelected, setLocalSelected] =
     useState<TFilterOption[]>(initialSelected);
 
-  const { teams, error } = useOrgTeams();
+  // Fetch all teams, including inactive ones
+  const { teams, error } = useOrgTeams({ isActive: null });
```
tglaz: unless just want to be explicit, probably better to not require setting `isActive: null`.
does `useOrgTeams()` not do same thing?

libs/react/betterangels-admin/src/lib/pages/teams/TeamFormDrawer.tsx
```diff
@@ -88,10 +125,10 @@ export function AddTeamDrawer(props: TProps) {
         <Button
           size="2xl"
           variant="accent"
-          onClick={handleCreate}
+          onClick={() => void handleSubmit()}
           disabled={disabled || !name.trim()}
         >
-          Create
+          {isEditing ? 'Save' : 'Create'}
```
tglaz: minor, and don't really mind, but can probably use `Save` for both

libs/react/betterangels-admin/src/lib/pages/teams/TeamsPage.tsx
```diff
@@ -27,12 +27,15 @@ type IProps = {
 
 type SortField = 'name' | 'createdAt';
 
+const teamDisplayName = (t: TeamType) =>
```
tglaz: can probably move it to some team related util as we do this in multiple places

libs/react/betterangels-admin/src/lib/pages/teams/TeamsPage.tsx
```diff
@@ -67,7 +71,10 @@ export function TeamsPage(props: IProps) {
 
   const { data, loading, previousData, refetch } = useQuery(
     AdminTeamsDocument,
-    { fetchPolicy: 'cache-and-network' },
+    {
+      variables: { filters: { isActive: showInactive ? null : true } },
```
tglaz: we can probably fetch all - the banwidth saved is probably not worth the server-side filtering + the extra query
then filter in FE

tglaz (changes requested):

libs/expo/betterangels/src/lib/ui-components/Filters/FilterTeams/FilterTeamsOptions.tsx
mikefeldberg: I made `useOrgTeams` default `true`, since most consumers only need active teams. This is the only place that might need both.

libs/expo/betterangels/src/lib/screens/NoteForm/Team.tsx
mikefeldberg: updated to `useMemo`

libs/expo/betterangels/src/lib/ui-components/Filters/FilterTeams/FilterTeamsOptions.tsx
mikefeldberg: I'd initially set the isActive default to `true` but that doesn't make sense. Updated

apps/betterangels-backend/teams/migrations/0002_team_is_active.py
tglaz, line 15: for future-proofing purposes, today we have concept or active/inactive, which works for the current use case.

however, in future we may have active/inactive/archived (or some other)
- `archived` may be a status we store just for posterity reasons, but those teams would just never get used (never displayed in UI - basically considered finalized Trash state). 
- as an example, let's say SELAH renames all of their teams 2-3 times over the course of a year, then at some point they may not want to see all of the `inactive` options in the UI. they may just mark some as `archived` - we don't need to do anything with this today, but storing `status` vs `isActive` makes this a simpler change in the future

libs/expo/betterangels/src/lib/screens/NoteForm/Team.tsx
```diff
@@ -11,14 +13,20 @@ interface ITeamProps {
 export default function Team(props: ITeamProps) {
   const { teamId, onTeamChange } = props;
   const { teams } = useOrgTeams();
+  const initialTeamId = useMemo(() => teamId, []);
```
tglaz, line 16: `useRef` may be better, it's the standard way to `keep the value from the first render`

mikefeldberg: I had `useRef` initially but got `Error: Cannot access refs during render`. Claude suggested changing it to `useState`, we compromised with `useMemo`

tglaz: ok.
that's one of the new warnings I think, though not necessarily a good one

tglaz (approved):

Issue #2335, opened 2026-08-18
Rework Maybe handling: one absence spelling, and services that take typed keyword arguments


vecchp:
## Background

Strawberry's `Maybe[T]` is tri-state — absent, `Some(value)`, and `Some(None)` for explicit null — but **absence has two different spellings** depending on how the field is declared. Strawberry injects `= None` for a bare `Maybe[T]`; writing `= UNSET` by hand overrides that:

| state | `is not None` | `is not UNSET` | `bool(f)` |
| --- | --- | --- | --- |
| absent, bare `Maybe[T]` | False | **True** | False |
| absent, `Maybe[T] = UNSET` | **True** | False | False |
| `Some(False)` / `Some('')` / `Some(None)` | True | True | True |

So `bool(f)` is the only presence check that is correct in every row. `is not None` breaks on `= UNSET` fields, `is not UNSET` breaks on bare ones, and truthiness on the unwrapped value breaks on `Some(False)`.

This already caused a real bug: `updateNote` / `updateTask` silently cleared a record's team on any update that did not mention `teamId`. Fixed in #2331 by relying on `strawberry.asdict`, which resolves the tri-state correctly on its own.

### 1. One absence spelling

Drop every `= UNSET` on a `Maybe` field so absence is always `None`, matching what Strawberry injects. `shelters/types/inputs.py` has roughly 50; the note/task/team inputs are already bare.

### 2. Rename the one remaining collapsing helper

**Partly done in #2331.** `maybe_value` had no production callers left once #2329 switched `teams/schema.py` to the idiom, so it was inlined and removed. Only `maybe_int_value` survives, with two callers in `notes/schema.py` that pass `team_id` on as a keyword argument.

What is left: it still folds absent and explicit null into `None`, and the name does not say so. Suggest `maybe_or_none_int`. Its docstring in #2331 now carries the guidance — prefer the idiom for a plain read, never use it to build an update dict — so start there.

The idiom to standardise on, which #2329 arrived at independently:

```python
name = data.name.value if data.name else None
```

Correct for both absence spellings, and needs no helper.

Do **not** reintroduce an `apply_maybe`-style helper for update dicts: `asdict` already omits absent keys, so the correct code is to not override it.

### 3. Services that take typed keyword arguments

`task_update` and `note_update` take `data: Dict[str, Any]` and `setattr` whatever they find. `teams/services.py::team_update` shows the better pattern:

```python
[...]

PR #2339, opened 2026-08-19
feat(notes,tasks): reject a team from another organization

vecchp:
Split out of #2310 so the rule can be reviewed on its own. **11 files, ~290 lines.**

## The gap

A note or task could reference a team belonging to a different organization. Nothing stopped it — the FK points at `teams_team` with no constraint tying it to the record's own organization, and no code checked.

## The rule

`teams.validators.validate_team_in_org` is the single definition, called from two places:

- **The note and task services**, so a GraphQL write gets a validation message rather than an integrity error.
- **`Note.clean()` / `Task.clean()`**, so the Django admin — whose `team` field is an unfiltered dropdown over every organization — cannot bypass it.

It lives in `teams.validators` rather than `teams.services` so `notes.models` and `tasks.models` can call it from `clean()` without models depending on services.

`clean()` covers **ModelForm writers**, which is the writer this design needs. It is not a general backstop: `objects.create()`, `bulk_create()`, `queryset.update()`, `loaddata`, data migrations and the shell all skip it, because nothing on the note/task path calls `full_clean()` yet. #2335 adds that, at which point the explicit service calls become the redundant ones — recorded in both `clean()` docstrings, per the layering `docs/styleguides/python.md` sets out on #2331.

## Tasks with no organization

`Task.organization` is nullable and `SET_NULL`, so deleting an organization leaves its tasks without one. Those tasks are still validated. `validate_team_in_org` allows a null team, so a legacy row stays editable in the admin and its organization can be repaired there; what is refused — in the admin and through GraphQL alike — is *attaching* a team to a record that has no organization.

Two things make the permissive alternative wrong. Deleting an organization cascades to its teams and nulls `team_id` in the same operation, so no org-less task ever carries a team that needed preserving. And #2312's composite FK is MATCH SIMPLE, satisfied whenever either column is null, so Postgres accepts any team on an org-less row — skipping the check in `clean()` would leave the admin as the one writer that could still introduce a cross-org reference.

## No storage-level backstop yet

This rule is enforced in Python only. #2312, two PRs up this stack, adds a composite foreign key on `(team_id, organization_id)` that makes a cross-org row unstorable.

## Verification

- **1367 passed**, 31 skipped
- mypy clean over 378 files; `ruff check` and `ruff format --check` clean
- No schema change — `schema.graphql` is byte-identical to #2331's
- Mutation-tested: neutering the validator fails **14** tests; dropping the `note_update` call fails **2**; restoring the org-less skip in `Task.clean()` fails **1**

Three query-count assertions rise by one — the existence check the validator performs.

## Deliberately not here

- A non-numeric `teamId` returns a top-level error rather than a validation message, on both the top-level and nested-task paths. It comes from `maybe_int_value`, so the fix belongs with it and covers every caller at once — #2343.
- A task nested in `createNote` takes its organization from the caller's permission group rather than from the parent note, so a note and its tasks are two separate decisions — #2344.

## Stack

Merges after #2331. #2312 and the rest follow via #2310.

## Summary by Sourcery

Enforce organization ownership when assigning teams to notes and tasks.

Bug Fixes:
- Prevent notes and tasks from referencing teams belonging to a different organization.
- Reject assigning teams to tasks that no longer have an organization while preserving organization-less tasks without teams.

Enhancements:
- Centralize note and task team-organization validation and apply it to GraphQL services and Django model validation so admin edits receive field-level validation errors.

Tests:
- Add validator, model, and GraphQL mutation coverage for valid, cross-organization, unknown-team, and organization-less team assignments.

### DIFF
--- a/apps/betterangels-backend/tasks/tests/test_models.py
+++ b/apps/betterangels-backend/tasks/tests/test_models.py
@@ -1,0 +1,51 @@
+from accounts.tests.baker_recipes import organization_recipe
+from django.core.exceptions import ValidationError
+from django.test import TestCase
+from model_bakery import baker
+from tasks.models import Task
+from teams.models import Team
+
+
+class TaskTeamOrgValidationTestCase(TestCase):
+    """A task's team must belong to the task's organization."""
+
+    def setUp(self) -> None:
+        self.org = organization_recipe.make()
+        self.other_org = organization_recipe.make()
+        self.own_team = baker.make(Team, organization=self.org)
+        self.other_team = baker.make(Team, organization=self.other_org)
+
+    def test_clean_allows_a_team_from_the_same_org(self) -> None:
+        task = baker.make(Task, organization=self.org, team=self.own_team)
+
+        task.clean()
+
+    def test_clean_allows_no_team(self) -> None:
+        task = baker.make(Task, organization=self.org, team=None)
+
+        task.clean()
+
+    def test_clean_rejects_a_team_from_another_org(self) -> None:
+        # Unsaved: #2312 adds a composite FK that makes the row unstorable.
+        task = Task(organization=self.org, team=self.other_team)
+
+        with self.assertRaises(ValidationError) as ctx:
+            task.clean()
+
+        self.assertIn("team", ctx.exception.message_dict)
+
+    def test_clean_allows_an_org_less_task_without_a_team(self) -> None:
+        # Deleting an organization nulls both columns, so this is the only
+        # org-less shape that exists; it has to stay editable in the admin.
+        task = baker.make(Task, organization=None, team=None)
+
+        task.clean()
+
+    def test_clean_rejects_a_team_on_an_org_less_task(self) -> None:
+        task = baker.make(Task, organization=None, team=None)
+        task.team = self.other_team
+
+        with self.assertRaises(ValidationError) as ctx:
+            task.clean()
+
+        self.assertIn("team", ctx.exception.message_dict)
--- a/apps/betterangels-backend/tasks/tests/test_mutations.py
+++ b/apps/betterangels-backend/tasks/tests/test_mutations.py
@@ -26,7 +26,7 @@ class TaskMutationTestCase(GraphQLBaseTestCase, TaskGraphQLUtilsMixin):
         client_profile = baker.make(ClientProfile)
         assert self.org
 
-        expected_query_count = 22
+        expected_query_count = 23
         with self.assertNumQueriesWithoutCache(expected_query_count):
             variables = {
                 "clientProfile": str(client_profile.pk),
@@ -81,7 +81,7 @@ class TaskMutationTestCase(GraphQLBaseTestCase, TaskGraphQLUtilsMixin):
             "teamId": str(self.org_1_team_1.pk),
         }
 
-        expected_query_count = 7
+        expected_query_count = 8
         with self.assertNumQueriesWithoutCache(expected_query_count):
             response = self.update_task_fixture(variables)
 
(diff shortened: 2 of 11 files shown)


vecchp:
Two things to reconcile from #2331, now that it has moved.

**1. `clean()` covers ModelForm writers, not "any writer that bypasses the services".** The commit message and both service docstrings here say the `clean()` calls cover writers that never touch the services. They cover the Django admin, which is the writer this design needs — but `Task.objects.create()`, `bulk_create()`, `queryset.update()`, `loaddata`, data migrations and the shell all skip `clean()` entirely, because nothing calls `full_clean()` on the note/task path. Worth narrowing the wording so the next reader does not over-trust it. There is no database-level backstop for this rule either; expressing it in Postgres needs a `UniqueConstraint(fields=["id", "organization"])` on `Team` plus a composite FK, which Django cannot generate.

`docs/styleguides/python.md` on #2331 now records the layering this follows, including that services must call `full_clean()` before `save()` for any model-side rule to be reachable from the API. Adding it to the note/task services is #2335.

**2. The shared test fixture changed shape.** `common/tests/utils.py` now builds its teams with `baker.make` and binds them to `self.org_1_team_1` / `self.org_1_team_2`. Two consequences for this branch:

- Every `Team.objects.get(name="WDI On-site", organization=...)` lookup here will raise `DoesNotExist` — no team carries that name any more. Use the fixture attributes.
- There is no longer an `org_2` team in the shared fixture (both were unused on #2331 and were dropped). The cross-org tests here need one, so add `self.org_2_team_1 = baker.make(Team, organization=self.org_2)` to the fixture as part of this PR, or keep the local `baker.make` these tests already use.

I will handle both in the forward-merge when #2331 lands unless you would rather I push them now.

apps/betterangels-backend/tasks/tests/test_mutations.py
```diff
@@ -187,3 +187,33 @@ def test_create_task_fails_when_linking_both_note_types(self) -> None:
...
+
+    def setUp(self) -> None:
+        super().setUp()
+        self.graphql_client.force_login(self.org_1_case_manager_1)
+        self.task_id = self.create_task_fixture(
+            {
+                "summary": "Org 1 task to amend",
+                "teamId": str(self.org_1_team_1.pk),
+            }
+        )["data"]["createTask"]["id"]
```
mikefeldberg: this should be in the update test, not `setUp`

apps/betterangels-backend/tasks/tests/test_models.py
```diff
@@ -0,0 +1,51 @@
...
+    def test_clean_allows_no_team(self) -> None:
+        task = baker.make(Task, organization=self.org, team=None)
+
+        task.clean()
+
+    def test_clean_rejects_a_team_from_another_org(self) -> None:
+        # Unsaved: #2312 adds a composite FK that makes the row unstorable.
+        task = Task(organization=self.org, team=self.other_team)
+
+        with self.assertRaises(ValidationError) as ctx:
```
mikefeldberg, line 32: could also do
```py
with self.assertRaisesRegex(ValidationError, "team"):
    task.clean()
```

mikefeldberg (approved):

apps/betterangels-backend/tasks/tests/test_mutations.py
vecchp: Done in b57b0369 — it builds the task inside the update test now.

Worth noting the other half of why it mattered: the fixture was calling `createTask`, which is the mutation the sibling test exercises. So a regression in create surfaced as a `TypeError` on `["data"]["createTask"]["id"]` in `setUp`, failing *both* tests, instead of failing the create test that is actually asserting on it.

apps/betterangels-backend/tasks/tests/test_models.py
vecchp: I looked at this one and would rather keep the two-liner, because the regex version cannot fail.

`assertRaisesRegex` searches `str(exception)`, and the message itself contains the word "team":

```
str(ValidationError({"team": [msg]}))  == "{'team': ['The selected team does not belong to this organization.']}"
str(ValidationError(msg))              == "['The selected team does not belong to this organization.']"
```

So `"team"` matches whether or not the error is scoped to the field:

| regex | field-scoped | non-field | scoped to wrong field |
| --- | --- | --- | --- |
| `"team"` | PASS | PASS | PASS |
| `"'team':"` | PASS | fail | fail |

The field scoping is the thing under test — `clean()` deliberately re-raises as `ValidationError({"team": exc.messages})` so the admin renders the error on the `team` field and GraphQL reports the right field. `assertIn("team", ctx.exception.message_dict)` asserts exactly that, and `message_dict` raises `AttributeError` on a non-field error, so it fails loudly.

The tighter `r"'team':"` does discriminate correctly, but it asserts against Python's `repr()` of a dict rather than the dict, and the quotes-and-colon pattern needs a comment to explain itself — which costs back the line it saves. Happy to switch if you feel strongly.

PR #2340, opened 2026-08-19
fix(teams): the teams query requires org membership, not just the header (DEV-2556)

vecchp:
Splits the security fix from the client-breaking change that was bundled with it. The breaking half is #2345.

## The gap

The `teams` query trusted `X-Organization-ID` outright — it looked the organization up by pk and listed its teams. Any authenticated user could name any organization and read its teams.

## The fix

Membership is now required when the header is present. **Membership**, not a `teams.*` permission: CASEWORKER holds no Team perms, so gating on `Team.perms.VIEW` would break the team picker for exactly the people who use it.

`accounts.selectors.organization_get_for_member` returns `None` for an unknown id, an organization the user does not belong to, and an unusable id alike, so one denial covers all three and the error cannot be used to probe for which organizations exist.

## What this deliberately does *not* change

The headerless path keeps its first-match fallback. Builds older than #2330 do not send the header and have queried `teams` since #2167, so denying them would remove the team picker, task-form team selector and team filter from every build still installed — and `runtimeVersion` is per build, so an OTA update cannot reach them.

**Until #2345 merges, a multi-org user on an old build can still be served the wrong organization's teams.** That is a smaller exposure than the one closed here (which needed no old build at all, just a forged header), and it is the part that has to wait on a release.

## Also in this PR

- `team_get` tolerates an unusable pk instead of raising `ValueError`, so a junk id in `updateTeam`/`deleteTeam` reads as a denial rather than a server error. That removes the `int(data.id)` coercion from both call sites.
- `team_list` drops `.order_by("name")`, which `Team.Meta.ordering` already provides.
- The shared fixture gains an org_2 team. The cross-org assertions compare against a live query for org_2's teams, so without one they would have compared two empty sets and passed while testing nothing.

## Verification

- **1364 passed**, 31 skipped
- mypy clean over 377 files; `ruff check` and `ruff format --check` clean
- Mutation-tested: trusting the header again fails 2 tests; dropping the org filter in `team_list` fails 10

No longer depends on #2328 — the headerless branch must not raise, so this reads `request.organization_id` directly. #2345 is where `get_current_organization` gets used, and that is where #2328 is a prerequisite.

Running the suite locally: allauth's rate limits live in the shared Redis cache and are not reset between runs, so back-to-back runs produce spurious `test_headless_views.py` failures — see #2346. Clear the cache first.

## Summary by Sourcery

Enforce organization membership for header-scoped team access without breaking legacy headerless clients.

Bug Fixes:
- Require organization membership when the teams query receives an organization header, preventing users from reading teams from organizations they do not belong to.
- Return permission denials instead of server errors for malformed or cross-organization team update and deletion requests.

Enhancements:
- Preserve the legacy headerless teams-query fallback for older application builds while enforcing active-organization scoping when a header is present.
- Centralize organization membership lookup and strengthen team selector coverage for organization scoping, ordering, malformed identifiers, and cross-organization access.

Tests:
- Add regression coverage for organization-header authorization, missing and malformed headers, cross-organization visibility, and team selector behavior.

### DIFF
--- a/apps/betterangels-backend/teams/tests/test_queries.py
+++ b/apps/betterangels-backend/teams/tests/test_queries.py
@@ -15,7 +15,7 @@ from accounts.tests.baker_recipes import organization_recipe
 from model_bakery import baker
 from teams.models import Team
 
-from .utils import TEAM_FIELDS, TeamGraphQLBaseTestCase, TeamGraphQLUtilsMixin
+from .utils import TeamGraphQLBaseTestCase, TeamGraphQLUtilsMixin
 
 
 class TeamsQueryTestCase(TeamGraphQLUtilsMixin):
@@ -83,23 +83,12 @@ class TeamsQueryTestCase(TeamGraphQLUtilsMixin):
 
 
 class TeamQueryOrgScopingTestCase(TeamGraphQLBaseTestCase):
-    def _teams_query(self) -> Dict[str, Any]:
-        query = f"""
-            query {{
-                teams {{
-                    totalCount
-                    results {{ {TEAM_FIELDS} }}
-                }}
-            }}
-        """
-        return self.execute_graphql(query)
-
     def _assert_denied(self, response: Dict[str, Any]) -> None:
         self.assertIsNotNone(response.get("errors"))
         self.assertIsNone((response.get("data") or {}).get("teams"))
 
     def test_returns_only_the_active_orgs_teams(self) -> None:
-        response = self._teams_query()
+        response = self.execute_graphql(self.get_teams_query())
 
         results = response["data"]["teams"]["results"]
         returned_ids = {int(row["id"]) for row in results}
@@ -115,7 +104,7 @@ class TeamQueryOrgScopingTestCase(TeamGraphQLBaseTestCase):
         OrgRoleManager(self.org_2).add_roles(self.org_1_admin, ORG_ADMIN)
         self._set_active_org(self.org_2)
 
-        results = self._teams_query()["data"]["teams"]["results"]
+        results = self.execute_graphql(self.get_teams_query())["data"]["teams"]["results"]
 
         returned_ids = {int(row["id"]) for row in results}
         org_2_ids = set(Team.objects.filter(organization=self.org_2).values_list("pk", flat=True))
@@ -125,7 +114,7 @@ class TeamQueryOrgScopingTestCase(TeamGraphQLBaseTestCase):
         """The server must not guess — first-match is how other orgs leaked."""
         del self.graphql_client.defaults["HTTP_X_ORGANIZATION_ID"]
 
-        self._assert_denied(self._teams_query())
+        self._assert_denied(self.execute_graphql(self.get_teams_query()))
 
     def test_denies_an_org_the_user_does_not_belong_to(self) -> None:
         """The header names the org; it does not grant access to it.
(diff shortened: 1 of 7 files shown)


apps/betterangels-backend/teams/tests/test_selectors.py
```diff
@@ -0,0 +1,78 @@
...
+
+
+class TeamListTestCase(GraphQLBaseTestCase):
+    def test_returns_only_the_given_organizations_teams(self) -> None:
+        teams = team_list(organization=self.org_1)
+
+        self.assertTrue(teams.exists())
+        self.assertEqual({t.organization_id for t in teams}, {self.org_1.pk})
+
+    def test_orders_by_name(self) -> None:
```
mikefeldberg: should this be a selectors test if ordering is set on the model? Does it make sense to test something as basic as model ordering?

apps/betterangels-backend/teams/tests/test_queries.py
```diff
@@ -69,3 +80,68 @@ def test_teams_query_filter(self) -> None:
         response = self.execute_graphql(self.get_teams_query(), variables)
 
         self.assertEqual(response["data"]["teams"]["totalCount"], 1)
+
+
+class TeamQueryOrgScopingTestCase(TeamGraphQLBaseTestCase):
+    def _teams_query(self) -> Dict[str, Any]:
```
mikefeldberg: this (`_teams_query` and `TEAM_FIELDS` constant) adds yet another pattern to our already inconsistent test design. I'm not opposed to it, I'd just like us to deliberately settle on something. I'm surprised claude came up with this given what was already there. I'd expect it to either use this pattern or replace it. Odd that it added another effectively identical one right next to it.

```py
class TeamGraphQLUtilsMixin(GraphQLBaseTestCase):
    def get_team_fields(self) -> str:
        return """
            id
            name
            isActive
        """

    def get_teams_query(self, fields: Optional[str] = None) -> str:
        return f"""
            query ($filters: TeamFilter) {{
                teams (filters: $filters) {{
                    totalCount
                    results {{
                        {fields or self.get_team_fields()}
                    }}
                }}
            }}
        """
```

vecchp: Agreed, and it was worse than a near-duplicate — `TeamQueryOrgScopingTestCase` extends `TeamGraphQLBaseTestCase`, which extends `TeamGraphQLUtilsMixin`, so `self.get_teams_query()` was already on `self`. It was shadowing an inherited helper rather than filling a gap, and `TEAM_FIELDS` had exactly one consumer in the repo: that one query.

Fixed in 829402b0 by using the mixin. I inlined the local wrapper entirely rather than leaving a one-line alias, so there is now one way to build a teams query in that file:

```py
response = self.execute_graphql(self.get_teams_query())
```

The only real difference was that `get_teams_query()` declares `query ($filters: TeamFilter)` while the local one declared no variables — not a reason to fork, since `test_teams_query` in the same class already calls it with no variables.

apps/betterangels-backend/teams/tests/test_selectors.py
vecchp: You are right — removed in 829402b0.

`team_list` is `Team.objects.filter(organization=organization)` with no `order_by`, and `Team.Meta.ordering = ["name"]`, so the assertion was exercising Django applying `Meta.ordering` and nothing the selector does. It was meaningful against the older version of the selector, which ordered explicitly; this PR is what removed that `order_by`, and the test became a framework test at the same moment without anyone noticing.

The tell was in the test itself: it needed a three-line docstring defending its fixture names against Postgres collation ordering vs Python codepoint ordering. A test that needs a paragraph to justify its setup, in order to assert a framework default, is not earning its place.

I also dropped the "in `Team.Meta.ordering`" clause from `team_list`s docstring — it only restated the default, and with the test gone it was an untested claim.

Nothing asserts team ordering now. I think that is the right trade: it is cosmetic, and pinning it would mean asserting a framework default somewhere else instead. Say the word if you would rather it lived as a model test.

mikefeldberg (approved):

Issue #2341, opened 2026-08-19
Remove the deprecated team compatibility fields: TeamType.slug and NoteType/TaskType.currentTeam


vecchp:
Three GraphQL fields exist only so that documents baked into already-released native app builds keep validating. They should come out together, once no build in the field selects them.

## The fields

| Field | Current behaviour |
| --- | --- |
| `TeamType.slug` | resolver returning `null` — the `Team.slug` column was dropped in #2331 |
| `NoteType.currentTeam` | deprecated alias for `NoteType.team`, same FK |
| `TaskType.currentTeam` | deprecated alias for `TaskType.team`, same FK |

## Why they must go together

Every `currentTeam` selection shipped in the native app is the same block:

```graphql
currentTeam {
  id
  slug
  name
}
```

All 12 of them, across 10 documents in `libs/expo/betterangels` — verified against `main` at the time #2331 was written. There is not one that omits `slug`. GraphQL validation rejects the **entire document** on an unknown field, so removing either field alone breaks exactly the same requests as removing both. Keeping one without the other buys nothing.

`libs/expo/betterangels` is consumed only by `apps/betterangels`, the store-distributed native app, which is why this needs a rollover window at all. No React web app selects either field — `libs/react/betterangels-admin`'s teams page selects `id name isActive createdAt`.

## When this is safe to do

When no app build still in use selects `currentTeam`. `slug` entered those documents in #2167 and the replacement (`team`) shipped in #2331, so the window is bounded by how long builds from that range stay in the field.

## What to do

1. Delete the `slug` resolver from `TeamType` (`apps/betterangels-backend/teams/types.py`).
2. Delete `current_team` from `NoteType` and `TaskType` (`notes/types.py`, `tasks/types.py`).
3. Delete `test_note_query_legacy_team_selection_still_resolves` and its task counterpart — they exist to stop this happening early, and both fail if either field is removed.
4. Regenerate `schema.graphql` and the frontend types.
5. This is a breaking schema change, so it needs the `graphql-inspector:approved-breaking-change` label.

No frontend work: current documents already select `team` and none select `slug`.

## Context

#2331 removed the legacy team enum and `Team.slug`. It initially deleted `currentTeam` outright; that was caught in review because shipped builds consume it. `slug` is the same hazard in the same selection block and was missed on the first pass.

vecchp:
Superseded by #2342, which does the removal as a draft PR rather than tracking it as an issue.

PR #2342, opened 2026-08-19
feat(api)!: remove the deprecated team compatibility fields

vecchp:
Draft on purpose — **do not merge until no app build in the field selects `currentTeam`.**

## What

Removes the three fields #2331 kept alive only so documents baked into already-released native app builds would keep validating:

| Field | What it was |
| --- | --- |
| `TeamType.slug` | a resolver returning `null` — the `Team.slug` column was dropped in #2331 |
| `NoteType.currentTeam` | deprecated alias for `NoteType.team`, same FK |
| `TaskType.currentTeam` | deprecated alias for `TaskType.team`, same FK |

Also deletes `test_note_query_legacy_team_selection_still_resolves` and its task counterpart. Their only job was to fail if this landed early.

## Why all three, or none

They were never independent. Every `currentTeam` selection shipped in the native app is the same block:

```graphql
currentTeam {
  id
  slug
  name
}
```

All 12 of them, across 10 documents in `libs/expo/betterangels` — checked against `main`; there is not one that omits `slug`. GraphQL validation rejects the **entire document** on an unknown field, so removing any single field breaks exactly the same requests as removing all three. Keeping one without the others protects nothing, which is the mistake #2331 made on its first pass: it restored `currentTeam` after review while still dropping `slug` from the same block.

## When it is safe to merge

`libs/expo/betterangels` is consumed only by `apps/betterangels`, the store-distributed native app — that is the only reason a rollover window is needed at all. No React web app selects either field (`libs/react/betterangels-admin`'s teams page selects `id name isActive createdAt`), so nothing web-side is waiting on this.

`slug` entered those documents in #2167 and the replacement `team` field shipped in #2331, so the window is bounded by how long builds from that range stay in use.

## Frontend

None required. Current documents already select `team`, and none select `slug` — the only frontend churn here is regenerated types.

## Verification

- **1346 passed, 31 skipped** (two fewer than #2331: the two legacy-selection tests)
- `mypy` clean over 376 files; `ruff check` and `ruff format --check` clean
- `export_schema` reproduces the committed `schema.graphql`; the diff is exactly the three field lines
- `expo-betterangels`, `react-betterangels-admin` and `ba-platform` pass typecheck, lint and test

This is a breaking schema change, so it needs the `graphql-inspector:approved-breaking-change` label before CI will pass.

Replaces #2341.

## Summary by Sourcery

Remove obsolete team compatibility fields and their legacy validation coverage from the GraphQL API.

Enhancements:
- Remove the deprecated GraphQL compatibility fields for team slugs and legacy current-team aliases from notes, tasks, and teams.
- Regenerate affected GraphQL type and schema artifacts after the breaking schema cleanup.

Tests:
- Remove legacy query tests that exclusively verified continued support for the deprecated team field selections.


vecchp:
Replayed onto `main` — it had fallen behind, and its diff against the new `main` was carrying reversals of #2351's `full_clean()` work and #2340's membership check. The net contribution is unchanged: the three field definitions, the two legacy-selection tests, and regenerated types.

Re-verified on `main`: **1392 passed, 31 skipped** (two fewer than `main`'s 1394 — exactly the two deleted tests); mypy clean over 382 files; `ruff check` and `ruff format --check` clean; `export_schema` reproduces `schema.graphql` with the three field lines gone; `nx typecheck` clean for `betterangels-admin`, `expo-betterangels` and `betterangels`.

**Still a draft on purpose.** Nothing about the rollover gate has changed: `app.config.js` sets `runtimeVersion` per build so OTA cannot reach older builds, and there is no minimum-version gate anywhere in the repo to make "every build in the field sends the current documents" observable.

One thing that *did* change: #2345 no longer waits on this gate. It was rewritten to stop the `teams` query guessing an organization while still serving single-org callers who send no header, so the security fix ships now and only the genuinely client-breaking removals — this PR, and dropping `allow_implicit_org` — sit behind the gate. That leaves two PRs blocked on it rather than three.

PR #2347, opened 2026-08-19
update DocumentModal to handle OperationInfo and unmount later (DEV-2510)

tglaz:
### [Doc Library] Can't delete the file from Doc Library
DEV-2510

* could never repro the bug, but possible fail points:
  * calling `closeModal` before mutation ran unmount the modal, which can cause random issues
  * if there was an `OperationInfo` error it wasn't handled, and `catch` block wouldn't catch it

## Summary by Sourcery

Make document deletion resilient by validating all mutation failure formats and delaying modal unmounting until the operation settles.

Bug Fixes:
- Handle GraphQL OperationInfo and top-level mutation failures consistently, including permission-specific errors and server-provided messages.
- Keep the document modal mounted until deletion completes to prevent premature unmounting and improve deletion error reporting.

Enhancements:
- Add a reusable mutation response validator for non-form mutations with comprehensive automated coverage.

Tests:
- Add coverage for successful mutations, permission failures, OperationInfo errors, GraphQL errors, fallback messages, and unclassifiable responses.


tglaz:
e2e passed!

shiozakiyuko (commented):
Hi @tglaz 

I was able to test only with iOS this time. For some reason, the Expo Go on Android launched with blank screen no matter how many times I uninstalled/installed the app. 

Overall, I think it's good. The reason why I said "I think" was after a few files of deletion, it gets a little hiccups like the screen gets stuck but I'm assuming that's because the file I tried to delete was bigger one than the others. 

I will try again on the android and I can let you know.

shiozakiyuko (commented):
Hi @tglaz 

On Android, I was able to launch the Expo Go app and scanned the QR code. Then it gets white, blank page. It doesn't show anything on the screen so I don't know what's going on. 

Do you get the same experience through your Android phone (not emulator)?

tglaz:
@shiozakiyuko - not sure why you'd see any special `android` issues.
On my android emulator it's fine. The `e2e` tests also pass on both iOS and Android, so the app is able to run on both in that environment.
Can you maybe try reinstall the app?

as of `iOS` - can you explain what "stuck" means? 
does it unresponsive while deleting, and once you see the success snackbar it's responsive again?

shiozakiyuko (approved):

vecchp:
Before the diff itself: I don't think DEV-2510 is fixable from the client. `deleteClientDocument` **deletes the row and then returns a permission error**, and `deleteNote` does the same thing. Verified against the real schema:

```
permitted user  → row deleted, OperationInfo{PERMISSION, field: "deleteClientDocument"}
real denial     → row kept,    OperationInfo{PERMISSION, field: null}
```

<details>
<summary>Repro (no files written, uses the suite's own fixtures)</summary>

```bash
cd apps/betterangels-backend && ../../.venv/bin/python - <<'EOF'
import json, os, django, logging
os.environ.setdefault("DJANGO_SETTINGS_MODULE", "betterangels_backend.settings")
django.setup(); logging.disable(logging.CRITICAL)
from django.test.utils import setup_databases, setup_test_environment
setup_test_environment(); setup_databases(verbosity=0, interactive=False, keepdb=True)
import unittest
from clients.tests.utils import ClientProfileGraphQLBaseTestCase
from common.models import Attachment
class Probe(ClientProfileGraphQLBaseTestCase):
    def test_probe(self):
        doc_id = self.client_profile_1_document_1["id"]
        self._handle_user_login("org_1_case_manager_1")
        r = self.execute_graphql("""
          mutation ($id: ID!) { deleteClientDocument(data: {id: $id}) {
            __typename
            ... on OperationInfo { messages { kind field message } }
            ... on ClientDocumentType { id } } }""", variables={"id": doc_id})
        print("RESP", json.dumps(r))
        print("ROW STILL THERE?", Attachment.objects.filter(id=doc_id).exists())
unittest.main(argv=["p", "Probe"], exit=False, verbosity=0)
EOF
```
</details>

**Cause.** `HasRetvalPerm` on a `mutations.delete` runs the object-permission check twice — once inside the resolver via `get_with_perms` before anything is written, and again on the returned instance afterwards. The second check resolves against the database row (the returned instance still carries its pk), so the identical check that passed before the delete fails after it, and with `fail_silently=True` (the default) that surfaces as an `OperationInfo` rather than an error. Traced:

```
success path:  get_with_perms OK pk=3242              ← pre-write check passes
               [row deleted]
               _resolve_obj obj=Attachment pk=3242  → DjangoNoPermission
               handle_no_permission fail_silently=True → OperationInfo{PERMISSION, field:"deleteClientDocument"}

denial path:   get_with_perms RAISED PermissionDenied → row kept, field: null
```

To be clear, nothing is unprotected — the pre-write check is what stops a genuine denial.

**Why that matters for this PR.** The false denial is indistinguishable from a genuine one except for `field` being non-null, which is an internal of strawberry-django's `handle_no_permission`. So the client can't tell them apart — and with `persist: true`, this PR turns today's transient "An error occurred while deleting the document" into a permanent "You don't have permission to access this app." on a delete that succeeded, with no success snackbar and the file vanishing from the list behind it.

**The fix belongs on the two backend fields**, and both correct patterns already exist in the repo: `PermissionedQuerySet` + `get_object_or_permission_error` (`delete_service_request`), or `filter_for_user(...).get()` + `PermissionDenied` (`delete_task`). Both check before the write, return a clean success payload, and let us write a real message ("You do not have permission to delete this document") instead of the extension's default. Worth noting that `PermissionedQuerySet`'s own docstring already explains that it skips the parent retval check — same problem, other guise.

I'm happy to open that backend PR (both fields, plus payload assertions — `test_delete_client_document` currently only asserts the row is gone, so this bug passes CI today) so you can rebase this on top.

The client-side hardening here is still worth having, and the inline notes I've left stand on their own regardless of sequencing. The one I'd treat as blocking is `persist: true`, since it makes the backend's false denial permanent rather than transient.

vecchp:
Separate from the bug, on the new shared helper rather than the fix: we now have three families doing this job — `libs/expo/betterangels/src/lib/apollo/graphql/response/` (9 `extract*` helpers, one of which, `hasGqlCombinedApiError`, already handles Apollo v4's error shape), `getFieldErrorsOrThrow`, and now `throwOnMutationFail`, whose first three steps are near-verbatim copies of `getFieldErrorsOrThrow`'s.

Which one is the intended survivor? Worth saying in the description so the other two get retired rather than accumulating.

vecchp:
@shiozakiyuko a possible explanation for the iOS "stuck" report, if it helps narrow the repro: `Documents.tsx` renders `{!!selectedDocument && <DocumentModal/>}` with no `key`, so tapping a different file while a delete is still in flight reuses the same component instance — `modalHidden` is still `true`, so the sheet never opens, and when the delete settles `finally { closeModal() }` clears the file that was just picked.

Worth trying "tap a second file right after confirming a delete" to see if that matches what you saw.

libs/expo/betterangels/src/lib/ui-components/DocumentModal.tsx
```diff
@@ -45,15 +53,41 @@ export default function DocumentModal({
   });
 
   const deleteFile = async () => {
-    closeModal();
+    // Hide immediately but stay mounted; unmount only after the mutation settles.
+    setModalHidden(true);
+
     try {
-      await deleteDocument({ variables: { id: document.id } });
+      const result = await deleteDocument({
```
vecchp, line 65: Apollo Client v4's mutate result is `{ data, error, extensions }` — there is no `errors` array, and with the default `errorPolicy: 'none'` a GraphQL error rejects this promise before `throwOnMutationFail` ever runs, so the catch falls back to the hardcoded string and the server's message is lost.

Adding `errorPolicy: 'all'` here is what makes step 1 of the helper reachable. `ClientContactDeleteBtn` does exactly that for the same job (non-form delete), and then checks `response.error`.

libs/ba-platform/src/lib/apollo/graphql/response/throwOnMutationFail.ts
```diff
@@ -0,0 +1,85 @@
...
+
+  // 1. Top-level errors (e.g. UNAUTHENTICATED, transport/server failures)
+  if (response.errors?.length) {
+    if (isUnauthenticatedError(response.errors)) {
+      throw new BaPermissionError(
+        response.errors.find((e) => e.message)?.message || undefined,
+      );
+    }
+
+    throw new Error(composeErrorMessage(response.errors.map((e) => e.message)));
```
vecchp: This branch can't be reached from a v4 mutation result — `ApolloClient.MutateResult` carries `error?: ErrorLike`, never `errors`, and `QueryManager.mutate` sets `result.error = new CombinedGraphQLErrors(...)`.

Suggest reading `response.error` and unwrapping it the way `hasGqlCombinedApiError` already does in this repo:

```ts
if (CombinedGraphQLErrors.is(response.error)) {
  const errors = response.error.errors;
  // UNAUTHENTICATED → BaPermissionError, else composed message
}
```

Worth keeping in mind that the app-wide `createErrorLink` already redirects on `UNAUTHENTICATED`, so that particular classification may be redundant here either way.

libs/ba-platform/src/lib/apollo/graphql/response/throwOnMutationFail.test.ts
```diff
@@ -0,0 +1,276 @@
...
+      });
+    });
+  });
+
+  describe('throws Error', () => {
+    describe('with response errors', () => {
+      describe('with server message', () => {
+        const testCases: TestCase[] = [
+          {
+            name: 'throws Error with server message',
```
vecchp: These six cases (105, 113, 167, 179, 194, 206) build `{ data, errors }`, which Apollo v4 never produces for a mutation — so they're covering a shape the client can't hand the helper. The real thing is a `CombinedGraphQLErrors` on `error`:

```ts
{ data: undefined, error: new CombinedGraphQLErrors({ data: null, errors: [{ message, extensions }] }) }
```

(`constructor(result: FormattedExecutionResult)` is the overload to use.)

libs/ba-platform/src/lib/apollo/graphql/response/throwOnMutationFail.test.ts
```diff
@@ -0,0 +1,276 @@
...
+  it.each(cases.map((c) => [c.name, c] as const))('%s', (_, tc) => {
+    const mergedParams = { ...defaultParams, ...tc.params };
+
+    if (tc.throws) {
+      expect(() =>
+        throwOnMutationFail({
+          response: tc.response,
+          ...mergedParams,
+        }),
+      ).toThrow(tc.throws.type);
```
vecchp: `toThrow(Error)` also passes for `BaPermissionError`, since it extends `Error` — so on the `type: Error` cases the class assertion is vacuous, and the `toContain` check at line 71 is the only thing discriminating. That check can't catch a wrongly-thrown permission error whose message is the server's pass-through text (167, 179, 206).

Adding `expect(e).not.toBeInstanceOf(BaPermissionError)` to the existing catch would close it. Worth noting that `toThrow` only does an `instanceof` check when given a class — given an error *instance* it deep-compares, so that form wouldn't assert what it looks like it asserts.

libs/expo/betterangels/src/lib/ui-components/DocumentModal.tsx
```diff
@@ -35,8 +37,14 @@ export default function DocumentModal({
...
 }: IDocumentModalProps) {
+  const fileTypeText = getFileTypeText(document.mimeType);
+
   const { showSnackbar } = useSnackbar();
-  const [modalState, setModalState] = useState<ModalState>('mainVisible');
+  const [modalState, setModalState] = useState<ModalState>('menuVisible');
+  // Hidden (but still mounted) while the delete mutation is in flight: the
+  // parent only unmounts us via closeModal(), which we call once the mutation
+  // settles — so the mutation always runs with the component alive.
+  const [modalHidden, setModalHidden] = useState(false);
```
vecchp: `useMutation`'s unmount cleanup only sets `isMounted = false`, which suppresses `setResult`; the promise from `client.mutate` still settles and the awaited code still runs. So the component doesn't need to stay mounted for the mutation to complete — and `DeleteModal` already hides itself on confirm (`onDelete(); setVisible(false);`), so the `return null` at line 167 is unmounting an already-invisible tree.

It does have a cost: `Documents.tsx` renders `{!!selectedDocument && <DocumentModal/>}` with no `key`, so tapping a different file during an in-flight delete reuses this instance with `modalHidden` still `true` — the sheet never opens, and `finally { closeModal() }` then clears the newly selected file.

Moving the mutation up into `Documents.tsx` (mounted regardless) would delete this state, the early return, and that stuck state together.

libs/expo/betterangels/src/lib/ui-components/DocumentModal.tsx
```diff
@@ -45,15 +53,41 @@ export default function DocumentModal({
...
+
+      if (err instanceof BaError) {
+        errorMessage = err.message;
+      }
+
       showSnackbar({
-        message: 'An error occurred while deleting the document',
+        message: errorMessage,
         type: 'error',
+        persist: true,
```
vecchp, line 106: See my top-level comment: the backend currently answers a *successful* delete with `OperationInfo{PERMISSION}`, so with `persist: true` a successful delete leaves a permanent "You don't have permission to access this app." on screen. Without it, worst case is today's behavior with a better message.

This is the one change I'd treat as blocking until the backend is fixed.

libs/expo/betterangels/src/lib/ui-components/DocumentModal.tsx
```diff
@@ -45,15 +53,41 @@ export default function DocumentModal({
...
-      await deleteDocument({ variables: { id: document.id } });
+      const result = await deleteDocument({
+        variables: { id: document.id },
+      });
+
+      throwOnMutationFail({
+        response: result,
+        operationKey: 'deleteClientDocument',
+        successTypename: 'ClientDocumentType',
+      });
```
vecchp: These two literals aren't checked by anything — a typo in either turns every response, including successful ones, into a generic error, and TS can't catch it.

`tools/codegen/plugins/operation-meta-plugin.cjs` already generates typed meta for exactly this (`deleteBedsMeta` in shelter-operator). Adopting it here needs `nonOptionalTypename: true` in `libs/expo/betterangels/codegen.ts` first — with optional `__typename` the generated `Extract<NonNullable<T[key]>, { __typename: 'X' }>` resolves to `never` (verified with the repo's tsc: `TS2322`). That's probably a follow-up; deriving the literals from `DeleteClientDocumentMutation` locally would do for now.

libs/ba-platform/src/lib/apollo/graphql/response/throwOnMutationFail.ts
```diff
@@ -0,0 +1,85 @@
...
+    );
+
+    if (permissionMsg) {
+      throw new BaPermissionError(permissionMsg.message || undefined);
+    }
+
+    // Compose ALL messages — no field filtering (non-form)
+    throw new Error(
+      composeErrorMessage(operationInfo.messages.map((m) => m.message)),
+    );
```
vecchp: The only caller surfaces `err.message` just for `BaError`, so these composed messages can't reach a user. That may well be right — the real `kind: ERROR` message here is `"Attachment matching query does not exist."`, which isn't something to show — but then composing them is dead weight. Either throw `BaError` for classified failures, or drop the composition on this path and keep the generic fallback.

libs/ba-platform/tsconfig.spec.json
```diff
@@ -8,6 +8,8 @@
     "types": ["node", "vitest/globals"]
   },
   "include": [
+    "src/**/*.ts",
+    "src/**/*.tsx",
```
vecchp: What error did this fix? `getFieldErrorsOrThrow.test.ts` imports its module under test the same way, and `ba-platform` has no `typecheck` target (only `test`, `lint`, `generate-graphql-types`) with nothing referencing its tsconfig, so `nx affected -t typecheck` never builds the spec project. Not objecting — it matches `libs/react/shared` — just curious whether it was an editor error, since it now makes every source file an input to both composite projects.

vecchp (commented):
Reviewed the diff; notes inline. Please read the top-level comment first — the root cause of DEV-2510 is server-side (`deleteClientDocument` deletes the row and then returns a permission error), so this PR can't close the ticket, and the `persist: true` change would make the user-visible symptom worse until that's fixed.

Everything else inline is independent of the backend and worth doing either way. Happy to pair on any of it.

libs/expo/betterangels/src/lib/ui-components/DocumentModal.tsx
tglaz: yeah. guess it's time to add `operation-meta-plugin` to other repos
i'll just do that now

vecchp:
Backend fix is up: #2387. Both fields check the permission once before the write and return a real success payload, so the false `OperationInfo{PERMISSION}` on a successful delete is gone and a genuine denial carries a written message.

Once that lands this can rebase on it, and `persist: true` is safe to keep.

libs/ba-platform/tsconfig.spec.json
tglaz: removed

libs/ba-platform/src/lib/apollo/graphql/response/throwOnMutationFail.ts
tglaz: right, i missed the CombinedGraphQLErrors.
for this  I just removed the file completely and will handle more of this ba-platform response handling consolidation via DEV-2538

libs/ba-platform/src/lib/apollo/graphql/response/throwOnMutationFail.test.ts
tglaz: file removed

libs/ba-platform/src/lib/apollo/graphql/response/throwOnMutationFail.test.ts
tglaz: file removed

libs/expo/betterangels/src/lib/ui-components/DocumentModal.tsx
tglaz: given other changes - I just removed this

libs/expo/betterangels/src/lib/ui-components/DocumentModal.tsx
tglaz: this call was removed

shiozakiyuko:
Hey @tom, as usual, I have to wait for a while to get to the app on Android but so far, this was what I experienced in iOS. Please see the video. Some hiccups to delete files.

Addition. On Android side, I can delete files smoothly.


https://github.com/user-attachments/assets/d1a39e1f-12d2-4319-9cfa-d5899b1e9732

libs/expo/betterangels/src/lib/ui-components/DocumentModal.tsx
tglaz: i updated this part to only check for success or OperationInfo (permissions).
Otherwise opened [DEV-2538](https://betterangels.atlassian.net/browse/DEV-2538) to consolidate all of the error response code.

vecchp (approved):

PR #2352, opened 2026-08-20
fix(accounts): make admin-created organizations usable, and fix the org role pipeline (DEV-2552)

vecchp:
## What

An organization created in the Django admin could hold no roles and accept no members — adding one returned **HTTP 500**. This makes that flow work, and fixes seven further defects in the org → role pipeline: four the fix would otherwise have made reachable, one in how the API validates a role, and two this branch introduced that review caught. Managing membership in the admin — listing members, editing roles, inviting with several — is **#2354**, stacked on this.

| Before | After |
|---|---|
| Create org in admin → add member → `500` | Create org → tick an org type → **Add member** → invite sent |
| Org has empty `org_types`, zero permission groups | Org type is required at creation; groups reconciled on save |
| A misconfigured org 500s | Explains that a type is needed first |

## Why it grew past the admin

The pipeline conflated **derived** state with **hand-managed** state. Groups implied by an organization's `org_types` are derived and reconciliation owns them; `Global Shelter Operator` and ad-hoc named groups are granted by hand and must be left alone. Reconciliation assumed everything was derived and deleted the rest. Naming that distinction is most of this change.

## Defects fixed

**1. A read performed a write.** `OrganizationProfile.organization` was an `AutoOneToOneField`, so `org.profile` *inserted* a row as a side effect of reading it. Every organization therefore looked configured while carrying an empty `org_types` — which is why the bug was silent — and the four `if hasattr(org, "profile")` guards could never be `False`. It is now a plain `OneToOneField` with a check constraint requiring at least one org type.

**2. Group names collided and went stale.** `auth.Group.name` is unique and capped at 150 characters; `Organization.name` is neither. A second organization sharing a name failed with `IntegrityError: duplicate key value violates unique constraint "auth_group_name_key"`, and renaming an organization left its groups named after the old name forever. Names are now `Better Angels [1] · Caseworker` — the pk makes them unique, the organization name makes them legible in the group picker, and `reconcile_org_groups` re-applies them so a rename does not go stale.

**3. Reconciliation orphaned groups, making roles unrevokable.** It deleted through a queryset, which skips the model's `delete()` and left the `auth.Group` behind — the member kept the permissions while the row needed to revoke them was gone. Group teardown moved to a `post_delete` receiver, the only mechanism that also covers cascades (deleting an organization previously orphaned every one of its groups).

**4. `Global Shelter Operator` would not survive a reconcile.** It belongs to no org type, so reconciliation treated it as stale. Reconciliation now only touches derived templates.

**5. `PermissionGroup.save()` minted a new `auth.Group` on every save.** The guard was `hasattr(self, "template")`, which is `True` even for a null FK, so re-saving a template-less row created a second group and orphaned the first — while a templated row could not be re-saved at all, raising `Updating a PermissionGroup with a template is not allowed.` The group is now created once, when the row is new.

**6. A role the organization cannot hold was accepted by the API.** `REGISTRY.get_template_or_raise` read as a validator and was not one — it resolved the name against the whole registry and used the organization only to word its error. `PermissionTemplateEnum` offers every invitable role globally, so `addOrganizationMember` and `changeOrganizationMemberRole` accepted e.g. `Caseworker` for a shelter-only organization, then failed in `OrgRoleManager.add_roles` and surfaced `PermissionGroup matching query does not exist.` to the client. It now validates against the roles the organization's org types allow, at no extra query — both mutations `select_related("profile")` on the organization they already fetch. Neither mutation had a test for this, and `changeOrganizationMemberRole` had none at all.

**7. This branch widened the membership admin form.** `main` registered `CustomOrganizationUserAdmin` with `OrganizationUserForm`, whose `exclude = ("user", "is_admin")` kept both off the form. Replacing that form with a bare `ModelAdmin` fell back to the default `ModelForm`, which offers every editable field — so `user` became repointable, and `is_admin` came back as a checkbox that looks like it grants admin and does not. Repointing either foreign key moves the membership while its org-scoped groups stay on the old pair, leaving a user holding every `org:<pk>:<role>` group for an organization they are no longer in. `fields` drops `is_admin` again and `get_readonly_fields` fixes both keys once the row exists; adding is already refused here, so nothing needs them writable.

**8. "A template or a name" held for the admin and nothing else.** `PermissionGroup.clean()` states the rule, but `clean()` only runs through a ModelForm — `objects.create()` and reconciliation's `get_or_create` bypass it. A row with neither leaves its `auth.Group` named `<organization> [<pk>] · `, an empty role segment that collides with the next such row on the unique `auth.Group.name`. It is now a check constraint (`0006`), the same shape as `org_profile_has_org_type` above, which is what makes the rule true for `bulk_create`, `loaddata` and the shell. Not hypothetical: adding it immediately failed three fixtures in `notes/tests/test_note_attachment_mutations.py` that were calling `baker.make(PermissionGroup, organization=...)`, which leaves both fields empty because both are optional. `clean()` stays, because it is what words the error on the inline — `_post_clean` runs with `validate_constraints=False` — and its docstring says which of the two calls is the redundant one.

### Defect 4 was latent, and this branch is what would have armed it

Worth stating precisely, because it drove the design. The old reconcile bailed at `if not expected: return`, and **94 of 99** production organizations carry `org_types = []` — including `Better Angels`, which holds the only Global Shelter Operator group. So reconciliation never reached its delete and nothing was ever wiped. That also means defect 3 produced **no orphans** in production; the one orphaned group predates multi-tenancy and has no users.

The exposure runs forward. Once the backfill populates `org_types`, `expected` becomes non-empty and the *old* reconcile would have deleted that GSO permission group, orphaning its `auth.Group` and leaving **5 users with permissions nobody could revoke**. The derived/hand-managed split is a prerequisite for the backfill, not incidental cleanup.

## Simplifications

- **Reconciling an organization is now a single pass.** `reconcile_org_groups` used to skip an organization with no profile outright, so its hand-granted groups never had permissions applied — which is why `post_migrate` ran a second, unscoped `sync_group_permissions()` after its per-org loop. The profile-less case now skips only the *derived*-group work and always applies permissions, so one call leaves an organization consistent and the second pass is gone.
- `REGISTRY` is now the single list of templates; `ALL_TEMPLATES` is gone.
- `group` is no longer offered in the permission-group admin or its inline. It is created by `PermissionGroup.save()` and torn down on delete, so selecting an existing one let staff bind a row to a group used for something else — whose deletion then took that group's members with it.
- Deleted as unused: `Registry.template_names_for`, `Registry.invitable_templates_for`, `Registry.invite_template`, `accounts.services.create_default_org_permission_groups`, `accounts.utils.create_missing_groups_for_org`.
- **`django-annoying` is removed from the dependency tree.** It was only ever used for the `AutoOneToOneField` above, and once that was gone it survived on two lines in historical migrations — one a dead import. `AutoOneToOneField` overrides only `contribute_to_related_class`, so swapping it for `OneToOneField` in `0002` changes recorded migration state, not the column; verified by uninstalling the package and building a database from zero.
- `uv.lock` is regenerated. Dropping `django-annoying` from the member `pyproject.toml` left the workspace-root lockfile stale, failing `uv lock --check` in CI.
- Adding a member is now a view on the organization page, where the workflow starts. It replaces a `ModelForm` that reassigned `self.instance.pk` to turn an INSERT into an UPDATE, declared no `Meta.model`, needed `request` smuggled in by a mixin carrying a file-wide `# mypy: ignore-errors`, and sent the invite email inside the admin's transaction. `accounts/admin_request_mixin.py` is deleted and the email is sent on commit.

## Roles come from one of two places, and both still work

Worth stating because an earlier revision of this branch broke the second one.

| Tier | Permissions come from | Refreshed |
|---|---|---|
| A role the code defines (`Caseworker`, `Shelter Operator`, …) | its `TemplateConfig`, mirrored onto the template row | every `migrate` |
| A role someone creates in the admin | the `PermissionGroupTemplate.permissions` set on it | every `migrate`, preserved |

The second tier is what makes a template a template: define a role once and it reaches every organization holding it. `sync_group_permissions` resolves which tier applies per template and writes the result onto the `auth.Group` that grants it, scoped to one organization when reconciliation calls it — which is what stops a newly created group granting nothing until the next deploy.

## Why group naming is in scope

It looks like drive-by cleanup and is not. `auth.Group.name` is unique; on `main` a group is named `f"{organization.name}_{template.name}"`. This PR makes creating an organization in the admin actually create its groups — so typing a name that already exists would raise `IntegrityError` instead of producing a working organization, and production holds **5 exact duplicate-name pairs**. Uniqueness has to land with the flow it protects.

## Migrations, and what they do to production

A read-only census informed both. Production: 99 organizations, 99 profiles, 94 with an empty `org_types`.

**`0004`** resolves org types from evidence rather than a blanket default:

| Organizations | Outcome |
|---|---|
| 5 already typed | untouched |
| 12 with a Caseworker group — these hold **all 129** memberships | resolved to `outreach` |
| 0 with members but no signal | would fall back to `outreach` **and be printed** |
| 82 with no members, no permission groups, nothing to infer | **empty profile removed** — left unconfigured |

No populated organization receives a guessed type. The 82 are left unconfigured deliberately: the constraint governs profiles that exist and an `Organization` may have none, so "no profile" states the truth — not configured as a tenant yet. Backfilling them to `outreach` would have asserted a type for organizations we know nothing about (several are shelters by name) and created **+246 `PermissionGroup` and +246 `auth.Group`** rows for organizations nobody uses. The admin requires a type the first time one is actually edited.

Consequently "no profile" is a supported state: `templates_for` returns `[]` for such an organization and reconciliation skips it, because *unconfigured* is not *no roles allowed* — treating it as the latter would delete roles.

**`0005`** renames groups to `<organization> [<pk>] · <role>`. It does not re-attach orphans: the census reports nothing to recover, and recovering by name would mean matching on the non-unique `Organization.name`. Orphans are reported for a human instead.

Both reverse cleanly, including with duplicate organization names — production has **5** such pairs, which would otherwise make the reverse collide on `auth.Group.name`. Note that `reconcile_org_groups` also maintains group names on every `post_migrate`, which makes `0005`'s forward rename belt-and-braces and means reversing it while running this code does not restore the old names; both are documented in the migration.

## Testing

- `1410 passed, 31 skipped` — full suite, on a database built from zero.
- `mypy` clean on 383 files with the cache cleared; `ruff check` and `ruff format` clean.
- `accounts/tests/test_admin.py` is new — nothing in the repo previously exercised a Django admin view.
- Migration decisions that read a data state the constraint forbids are unit-tested as pure functions; their effects are verified by running the migrations.
- **Rehearsed on a real database** reshaped to the production shape (types blanked, unused shells added): 5 resolved from permission groups, 3 shells left unconfigured, 0 rows created for them.
- `0006` reports offending rows and stops rather than resolving them. A row with neither field has nothing left saying which role it was meant to be, so guessing a name or deleting a row that may still grant access are both worse than asking. The dev database has none.
- Each new test mutation-tested: with its fix reverted, it fails. For the constraint that means removing it from the migration as well as the model — a rebuilt database still enforces it otherwise, and the test passes vacuously.

## Known, not addressed

- Unchecking an org type now correctly strips that role from every member. Correct, but it deserves a confirmation step — follow-up.
- `Organization.add_user` makes the first user the owner and `organization_remove_member` refuses to remove the owner, so the first person invited to a new org becomes its un-removable owner. Pre-existing and identical on the GraphQL path.
- **`SITE_ID` resolves to `example.com`** on dev, and `member_invite` passes that `Site` into the invitation email, so invite links there point at example.com. The GraphQL path always did this too, so if production's `django_site` row is likewise unconfigured, invitation links are already broken — worth confirming now that admin-initiated invites are a real workflow.
- A rename made outside a reconcile leaves group names stale until the next one. Acceptable only while nothing reads them as data, which is true today and noted in the code.
- The census surfaced near-duplicate organizations that exact-name matching misses — notably `Assured Life Style Housing` (10 members, untyped) vs `Assured Lifestyle Housing Inc` (0 members, typed), which looks like a re-creation. Both are handled correctly here, but the data hygiene is worth its own look.

## Summary by Sourcery

Make admin-created organizations configurable and usable while hardening organization role reconciliation, permissions, invitations, and membership lifecycle management.

New Features:
- Add organization-admin workflows for configuring org types, inviting members with selected roles, changing member roles, and displaying organization roles.

Bug Fixes:
- Fix admin-created organizations failing to accept members by requiring valid organization types and reconciling their permission groups.
- Preserve hand-managed roles during reconciliation, clean up backing groups reliably, and apply configured permissions to newly created groups.
- Prevent permission-group name collisions and stale names when organizations share or change names.
- Ensure invitations are sent only after successful transactions and prevent membership deletion from leaving organization roles behind.

Enhancements:
- Separate derived permission groups from hand-managed groups and make organization reconciliation a single consistent operation.
- Remove duplicated template permission storage and replace the automatic one-to-one profile behavior with an explicit constrained profile.
- Remove the obsolete django-annoying dependency and simplify organization and permission-group management APIs.

Build:
- Remove django-annoying from the backend dependency set and lockfile.

Tests:
- Add comprehensive Django admin, migration, reconciliation, permission synchronization, invitation, role-management, and membership-deletion coverage.

Chores:
- Add migrations to backfill organization types, deterministically rename permission groups, and remove duplicated template permissions.


apps/betterangels-backend/accounts/migrations/0004_require_org_type_on_profile.py
```diff
@@ -0,0 +1,161 @@
...
+and a check constraint requires at least one org type on the profiles that remain.
+
+Existing rows are resolved from evidence rather than a blanket default.  An
+organization whose permission groups imply a type gets that type; one with members
+but nothing to infer from falls back to outreach and is printed, because that is a
+guess someone should see.  An organization with no members and nothing to infer
+from keeps no profile at all: the constraint governs profiles that exist, and an
+``Organization`` is allowed to have none.  That state means "not configured as a
+tenant yet", which is the truth for these rows — inventing a type for them would
+assert something we do not know, and would provision roles nobody uses.
```
mikefeldberg: not the clearest of docstrings. Could be shorter and simpler. Maybe just bullet points rather than paragraph form

mikefeldberg (commented):
## Soft / optional findings

### 1. `PermissionGroup.clean()` may not run from admin inlines

Validation for “template or name required” is only enforced if `full_clean()` runs. Django admin inlines often skip it. Empty rows can still create colliding/` · `-suffix group names. Optional: inline form validation or require template when name is blank.

### 2. Removed “no update when templated” guard (low likelihood)

Main blocked re-saving a templated `PermissionGroup`. That is gone. Reproducible only by deliberately changing **Template** on an existing inline row to a role the org doesn’t already have — reconcile then recreates the expected role as a new empty group and deletes the edited one (members on the old `auth.Group` are dropped). Changing to a template already on the org usually fails earlier on `unique_together`. Soft follow-up: make `template` read-only when `pk` is set — not a merge blocker.

---

## Lower findings

- **`templates_for` uses `hasattr(org, "profile")`:** Works (RelatedObjectDoesNotExist ⊆ AttributeError) and is tested, but `try/except ObjectDoesNotExist` is clearer and avoids descriptor quirks.
- **Invite success redirects to the org changelist**, not the org change page — minor UX regression vs staying on the org.
- **No queryset-delete orphan test:** Instance delete is covered; the production bug was queryset `.delete()`. One `PermissionGroup.objects.filter(...).delete()` assertion would lock the signal path.

vecchp:
Thanks — going through these in order.

### 2. Removed "no update when templated" guard — agreed, fixed in `eb0de8b6`

You're right that dropping `main`'s `ValidationError("Updating a PermissionGroup with a template is not allowed.")` opened this, and the repro is exactly as you describe. One correction to the outcome, which I only found by writing the test: it doesn't drop the members, it **500s**.

Repointing the row leaves it holding a group still named after the old role, so reconciliation's `get_or_create` for that role tries to create a second `auth.Group` with the same name and hits `auth_group_name_key` — *before* it reaches the delete that would have freed the name:

```
DETAIL:  Key (name)=(Test Org [39260] · Caseworker) already exists.
```

The admin's transaction rolls back, so no access is lost. The deterministic naming accidentally protects the data — but a plausible edit crashing is still a bug, so `template` is now `disabled` once `pk` is set (disabled rather than read-only, so submitted data is ignored and a hand-crafted POST can't change it either). `name` stays editable; renaming a hand-managed role is harmless and `_refresh_group_names` relabels its group on the next reconcile.

Worth knowing: that test also surfaced a crash in the stacked PR, where the same form renders the Members inline — its display callables reversed a URL with `None` for Django's blank row, so *any* validation error on the organization form would have 500'd. Fixed in #2354.

### 4. Invite redirects to the changelist — correct here, fixed in #2354

Good catch, and accurate for this PR. It moved to #2354 along with the Members inline, since redirecting to the organization page is only an improvement once that page lists members.

### 1. `clean()` from admin inlines — I don't think this one holds

Admin inlines are ModelForms, and `ModelForm._post_clean` does call it:

```python
self.instance.full_clean(exclude=exclude, validate_unique=False, validate_constraints=False)
```

Uniqueness and constraints are skipped there, but `Model.clean()` runs. And a blank extra row never reaches validation or save at all — `BaseForm.full_clean` returns early at `django/forms/forms.py:334` when `empty_permitted and not self.has_changed()`, and with the inline's `fields = ("template", "name")` a row with both blank can't be "changed". So I don't think the ` · `-suffix collision is reachable from the inline.

There *is* a real version of this, though, and it's the programmatic path rather than the admin: `PermissionGroup.objects.create(organization=org)` bypasses `clean()` entirely, and `reconcile_org_groups` uses `get_or_create` without `full_clean()` — which our styleguide's "services call `full_clean()` before `save()`" would flag. It's unreachable today only because reconcile always supplies a template. Happy to tighten that if you'd rather not rely on that.

### 5. Queryset-delete orphan test — there is one

`test_removing_an_org_type_deletes_its_groups_without_orphaning` goes through `reconcile_org_groups`, whose delete *is* a queryset delete:

```python
PermissionGroup.objects.filter(organization=org, template__name__in=derived - expected).delete()
```

then asserts `Group.objects.filter(pk=group_id)` no longer exists. So the path is locked, just via its real caller rather than a bare `.filter().delete()`. If you'd prefer an explicit one that names the mechanism I'm glad to add it.

### 3. `hasattr(org, "profile")` → `try/except ObjectDoesNotExist`

Fair, and agreed it reads better. Leaving it for now because #2354 leans on the same trick in the Members inline (`hasattr(group, "permissiongroup")`), so I'd rather change both together than split the idiom across two PRs.

---

Also: the CodeQL alert on this PR (`py/redos`, alert 52) is a false positive. It points at `0005_deterministic_permission_group_names.py:20`, which is prose inside the module docstring — no version of that file contains a regex or imports `re`. Worth dismissing.

vecchp:
Thanks — all five answered. Two of them shifted once I dug in, and I found one thing you didn't flag.

## The membership form was wider than `main`, and that's on this PR

Not something you raised, but it belongs here. `main` registered `CustomOrganizationUserAdmin` with `OrganizationUserForm`, whose `exclude = ("user", "is_admin")` kept both off the form. Replacing that form with a bare `ModelAdmin` falls back to the default `ModelForm`, which offers every editable field — so `user` became repointable, and `is_admin` came back as the checkbox that looks like it grants admin and doesn't.

Repointing either FK moves the membership while its org-scoped groups stay on the old pair. `fields` drops `is_admin` again and `get_readonly_fields` fixes both keys once the row exists — adding is already refused here, so nothing needs them writable. Fixed in `5cfbb8af`.

## 1. `clean()` from admin inlines — answered with a constraint, and you were more right than I was

I still think the admin path is safe: `BaseModelForm._post_clean` calls `full_clean(exclude=…, validate_unique=False, validate_constraints=False)`, so `Model.clean()` does run, and a blank extra row returns early from `BaseForm.full_clean` before validation.

But the programmatic hole I conceded turned out to be real rather than theoretical, so I fixed it properly instead of adding `full_clean()` to the service. Per the styleguide's "must hold for every writer → `Meta.constraints`", `PermissionGroup` now carries a check constraint requiring a template or a name — the same shape this PR already uses for `org_profile_has_org_type`. That covers `objects.create()`, `bulk_create`, `loaddata` and the shell, which `full_clean()` in one service never would.

The constraint immediately failed the suite: three fixtures in `notes/tests/test_note_attachment_mutations.py` were calling `baker.make(PermissionGroup, organization=...)`, which leaves `name` blank and `template` null because both are optional. So the codebase was already creating rows that identify no role — just in tests. `permission_group_recipe` exists to supply a generated name and is now used there.

`clean()` stays, because it's what words the error on the inline (`_post_clean` runs with `validate_constraints=False`). The docstring says which of the two is the redundant call, so the right one gets deleted later.

`6d99dcf0` (constraint + migration `0006`), `83e073d8` (fixtures).

## 5. Queryset-delete orphan test — added

You were right that naming the mechanism matters even though the path was covered. `test_deleting_through_a_queryset_still_deletes_the_group` does a bare `PermissionGroup.objects.filter(...).delete()` and asserts the `auth.Group` goes with it. `63749472`.

## 3. `hasattr` → `try/except ObjectDoesNotExist` — done, in #2354

All three sites together, as discussed: `templates_for` here, plus the Members inline's `permissiongroup` and the `organizationowner` lookup that #2354's N+1 fix added. It's `50a7f071` on that branch.

## 2 and 4

Already landed — the `template` guard in `eb0de8b6`, and the invite redirect with the Members inline in #2354.

---

The migration reports offending rows and stops rather than resolving them: a row with neither field has nothing left saying which role it was meant to be, so guessing a name or deleting a row that may still grant access both seemed worse than asking. Dev DB has none.

Full suite green on this branch (1410 passed, 31 skipped), mypy clean on 383 files with the cache cleared, ruff clean. Each fix mutation-tested.

mikefeldberg (approved):

PR #2353, opened 2026-08-20
chore(android): ignore generated Android build output

guerrini:
`expo prebuild` / `run:android` generate `apps/betterangels/android/` and
per-module `android/build/` directories. Without these entries a local Android
build leaves thousands of untracked files in `git status`.

The `expo-splash-screen` dependency that originally accompanied this fix is
already on main, so only the ignore rules remain.

Split out from in-progress referral module work so the build fix can land on its
own. No behaviour change — four lines added to `.gitignore`, nothing else.

## Summary by Sourcery

Build:
- Ignore generated Android project and build output to keep local prebuild and Android build artifacts out of version control.


guerrini:
Superseded by #2466, opened from a branch in this repository.

PR #2354, opened 2026-08-20
feat(accounts): manage organization membership in the admin

vecchp:
## What

Stacked on **#2352**, which makes an admin-created organization usable. This adds the membership management the admin was missing: seeing who belongs to an organization, changing what they can do there, and removing them safely.

Adding a member used to redirect to the organization *changelist*, and the organization page had no members list — so there was nothing to see afterwards. *Organization users* offered no **Add** at all, because the default ModelForm would have created a membership with no role: the state that made an organization unusable in the first place. And changing an existing member's role had no admin route, while the API has had `changeOrganizationMemberRole` all along.

| Action | Where |
|---|---|
| Grant roles initially | Organization → **Add member**, or Organization users → **Add** (which asks for the organization too) |
| Change a member's roles | **Change roles** on the member's row, or from the Organization users list |
| Revoke everything, keep membership | **Change roles**, clear every box |
| Remove a member | Organization users → delete, which clears their roles first |

## Defects fixed

The first two are the same class as the ones #2352 fixes. The rest came out of review, and three of them are defects this branch introduced.

**1. Removing a membership left the roles granted.** `organization_remove_member` clears a member's org-scoped roles before deleting the row; the admin deleted the row alone, leaving the user holding every group for an organization they no longer belonged to. Deletion now routes through that service — the service commit `cd10a5aa` centralised this rule deliberately, so the admin calls it rather than reinstating the `post_delete` receiver that commit removed.

**2. Removing the owner's membership un-owned the organization.** `OrganizationOwner.organization_user` is `CASCADE`, so the ownership row went with it. `AbstractOrganizationUser.delete` does raise `OwnershipRequired`, but only from `Model.delete()` — so the single-object view turned it into a 500, and the changelist's bulk delete, which goes through `queryset.delete()`, was unguarded. Deletion permission is now withheld for exactly the rows the service refuses: the owner's, and your own. `get_deleted_objects` consults that per selected row, so `delete_selected` refuses the whole batch rather than partially deleting.

**3. Change roles revoked every role it could not offer.** The form lists `REGISTRY.invitable_template_names_for`, but `member_roles_replace` went through `OrgRoleManager.replace_roles`, which clears *every* group with a `PermissionGroup` on the organization before re-adding the posted ones. Any save therefore dropped what has no checkbox — `Organization Admin`, `Organization Superuser`, `Global Shelter Operator`, and rows created by hand with a name and no template. A save that changed nothing still demoted, and the owner holds `Organization Admin` and is offered this form from the Members inline. It was invisible in both directions: `initial` carried the held names, but they are not in `choices`, so nothing rendered.

Merging the held roles into the target set does not fix it. `add_roles` resolves a `PermissionGroup` by `template__name`, so a template-less row has no `TemplateConfig` to re-add it with — clearing destroys it for good. The operation is inverted instead: add the selected templates, remove the offered-but-unselected ones, and never touch anything outside the set the form manages, which leaves nothing to reconstruct. Silently keeping a role is the mirror image of silently dropping one, so the page now names what it is not showing checkboxes for.

**4. `changeOrganizationMemberRole` had the same defect.** It called `replace_roles` directly, and `PermissionTemplateEnum` cannot name a non-invitable role — so no caller could ask to keep `Organization Admin`, and every call stripped it. Granting an org admin Caseworker demoted them in the same breath. Live on `main` rather than introduced here, but leaving a broken sibling of the thing this PR fixes made no sense, so it routes through `member_roles_replace` too. The `PermissionDenied` membership check stays in front of it, because that is the error the mutation has always returned for a non-member. Its docstring claimed it could promote; it could only ever demote.

**5. The Members inline queried twice per member.** `owner` does a hasattr on `organizationowner`, the reverse of a OneToOneField, and django-organizations' `AbstractOrganizationUser.__str__` reads `self.organization.name` — which the inline renders once per row. That is 180 queries on the 90-member organization cited below. Both are in `select_related` now.

**6. The role editor rendered for people who are not members.** The view loaded the target by `User`, so a GET for someone outside the organization showed a role editor for a membership that does not exist; only the POST was refused. It looks the membership up instead and 404s.

## A parallel CRUD surface was routed, and is now gone

Chasing the broken "View on site" link found the reason it was broken. `accounts/urls.py` mounted `organizations.urls`, exposing 11 URL names **this project uses none of**. The package ships templates for them, so they rendered:

| URL | Guard (the package's own) | Effect |
|---|---|---|
| `/accounts/` | none | lists every organization |
| `/accounts/add/` | none — *"Allows any user to create a new organization"* | any authenticated user creates one |
| `/accounts/<pk>/` | any **member** | enumerates that organization's members |
| `/accounts/<pk>/delete/` | **owner** | deletes the organization |
| `/accounts/<pk>/people/<pk>/delete/` | `is_admin` or superuser | deletes a membership **without clearing roles** |
| `/accounts/<pk>/people/add/` | same | creates a membership with **no role** |

The last two are the defects above, reached by a path a `ModelAdmin` guard cannot see — and a role-less membership is the state that caused the original 500. `is_admin` is never set in this codebase, so those gate to superuser in practice; the organization-level rows do not.

Only the invitation flow is kept, from `invitation_backend().get_urls()`, which was already mounted separately. Nothing reversed the removed names: the sole caller of `InvitationBackend.get_success_url` — which reverses `organization_list` — is `BaseBackend.activate_view`, and `CustomInvitations` overrides that to raise `Http404`.

Worth knowing for review: the domain never mattered here. `contenttypes.views.shortcut` falls back to a **relative** redirect when it cannot resolve a domain, exactly like every other admin link; it only produced `example.com` because `django.contrib.sites` is installed *and* `SITE_ID = 1` is set, which makes `get_current_site()` prefer the database row over the request host.

## Behavior change

The old form assigned **every** invitable template for the organization, so a member added to a dual-type org silently received both Caseworker *and* Shelter Operator. Staff now select the roles explicitly — one or several.

Because a single invitation email is sent however many roles are granted, `invitation_role` chooses it: a role whose invite template differs from the generic organization invitation wins, since that body carries something role-specific (Shelter Operator's links to the shelter app). Caseworker names the generic template explicitly, so having an `invite_html` does not distinguish them — it is compared against the backend's default. The choice is named in the success message rather than left to be guessed.

## Also here

- **`hasattr` on a related descriptor is gone**, in all three places: `templates_for`, the Members inline's `permissiongroup`, and the `organizationowner` lookup above. It works only because `RelatedObjectDoesNotExist` subclasses `AttributeError`, so it reads as a presence check while swallowing that one exception. Raised in review on #2352 against the first of those; done here because this branch is where the other two live.
- **"View on site" is off** for organizations, memberships, invitations and the Members inline — see below; with the generic views unrouted, `get_absolute_url` raises `NoReverseMatch`, so offering the link would error.
- The organization page links to the filtered, paginated member list, because an inline renders every row and one production organization has 90 members.
- Both invite entry points share a `MemberInviteAdminMixin` and one `member_invite` service; the two single-object role displays share selectors in `accounts/selectors.py`. The Members inline keeps reading off its prefetch and says why — a per-row selector would be an N+1.

## Testing

- `1442 passed, 31 skipped` — full suite.
- `mypy` clean on 383 files with the cache cleared; `ruff check` and `ruff format` clean.
- Every new behaviour is mutation-tested: reverting the fix fails the test. That includes the bulk-delete path, which a `delete_model` override would miss, and the unrouted views, asserted at `reverse()` level so a view merely starting to refuse would not pass.
- The inline's query count is pinned by rendering the page at two member counts and asserting they match, rather than against a fixed number that would drift with unrelated admin changes.

## Known, not addressed

- Unchecking an org type strips that role from every member. Correct, but it deserves a confirmation step.
- `Organization.add_user` makes the first user the owner, so the first person invited to a new organization becomes its owner. Ownership is transferable via the `OrganizationOwner` admin, which is what the refusal message points at.
- A member with no roles is a supported state — it is what **Add member** starts from — so the role form allows clearing every box while the invite form still requires at least one.

## Summary by Sourcery

Manage organization membership and roles through safe, role-aware admin workflows.

New Features:
- Add admin workflows to invite organization members with one or more roles and edit existing members’ organization-scoped roles.
- Expose organization membership and role information on organization, membership, and user admin pages.

Bug Fixes:
- Route membership removal through the service layer so organization roles are revoked and protect owner and self-memberships from individual and bulk deletion.
- Preserve non-invitable organization roles when changing a member’s invitable roles.
- Remove the unused generic organization CRUD routes that allowed unsafe parallel membership operations.

Enhancements:
- Add member listings, role editor links, filtered member navigation, and invitation feedback across the admin.
- Disable invalid admin “View on site” links and centralize invitation and role-management behavior in shared services and forms.
- Update the GraphQL role mutation to use the shared role replacement behavior.

Tests:
- Add comprehensive coverage for multi-role invitations, role changes, protected deletion paths, admin navigation, unrouted generic views, role preservation, and inline query performance.


mikefeldberg (commented):
## High: Change roles silently demotes Org Admin / Superuser

`[OrganizationMemberRoleForm](apps/betterangels-backend/accounts/forms.py)` only offers **invitable** roles (`REGISTRY.invitable_template_names_for`). `[member_roles_replace](apps/betterangels-backend/accounts/services.py)` then calls `[OrgRoleManager.replace_roles](apps/betterangels-backend/accounts/role_manager.py)`, which **clears every org-scoped group** and re-adds only the posted templates.

Promotion roles (`Organization Admin`, `Organization Superuser`, `is_invitable=False`) never appear as checkboxes. Saving the form — even with no intentional change — revokes them.

```mermaid
flowchart LR
  held["Holds Caseworker + Org Admin"] --> form["Form shows only Caseworker"]
  form --> post["POST permission_templates=Caseworker"]
  post --> clear["replace_roles: clear all"]
  clear --> result["Only Caseworker left"]
```



Owners created via the normal path get `ORG_ADMIN`. The Members inline still offers **Change roles** on those rows. One accidental save demotes the owner.

Initial values include non-invitable names that are not in `choices`, so the UI never shows them as held — the loss is invisible.

Tests cover invite/revoke among invitable roles only; no case with `ORG_ADMIN` + Change roles.

**Fix direction:** Preserve non-invitable roles across replace (merge held promotion roles into the target set), or split editable vs locked roles in the form/UI and never clear locked ones. Add a regression test: member with `(CASEWORKER, ORG_ADMIN)` → save with Caseworker checked → still holds Org Admin.

Bugbot: [Bugbot](ab7228fe-b12c-4edf-af76-a98e6090b496) — same finding.

---

## Medium

1. **Members inline N+1 on Owner** — `[owner()](apps/betterangels-backend/accounts/admin.py)` uses `hasattr(obj, "organizationowner")` without `select_related`/`prefetch_related` for that relation, while roles carefully avoid N+1. With ~90 members this is one query per row.
2. **Change-roles GET does not require membership** — loads any `User` by id; only POST fails via `member_roles_replace`. Prefer 404 when no `OrganizationUser` for that org/user.
3. **OrganizationUser change form still edits `organization` / `user`** — moving a membership by FK can leave org-scoped groups on the old org (same class of bug deletion just fixed). Consider making those readonly and routing moves through a service, or at least documenting the footgun.**OrganizationUser**

vecchp:
Thanks — all four confirmed and fixed. Three of them turned out worse than described, so flagging what changed beyond what you wrote.

## High — Change roles silently demotes

Confirmed, and it's broader than promotion roles. `clear_roles` removes every group with a `PermissionGroup` on the org, so it also takes `Global Shelter Operator` (also `is_invitable=False`) and **hand-created template-less `PermissionGroup` rows** — the ones `reconcile_org_groups` documents as deliberately left alone.

That last case rules out the merge-held-roles-into-the-target-set fix: `add_roles` resolves a `PermissionGroup` by `template__name`, so a row with a name and no template has no `TemplateConfig` to re-add it with. `clear_roles` destroys it and nothing puts it back.

So I inverted the operation instead. `member_roles_replace` now adds the checked templates and removes the *offered-but-unchecked* ones, never touching anything outside the set the form manages — so nothing has to be reconstructed, and it needs no extra argument: "edits the org's invitable roles, leaves everything else alone" is one rule.

On the invisibility: you're right that `initial` carried names that aren't in `choices`. Silently preserving them is the mirror image of silently dropping them, so the form exposes `locked_role_names` and the page says *"Also holds, and keeps: Organization Admin. These are granted from the group picker on the user's page."* The success message names them too.

Regression tests are the case you asked for — `(Caseworker, Org Admin)` → save with Caseworker checked → still holds Org Admin — plus the template-less variant, and one asserting the page names what it isn't showing checkboxes for. `342372d0`.

## Medium 1 — Members inline N+1

Confirmed, and there were two. Fixing only `organizationowner` left the query count still growing 1:1 with members: django-organizations' `AbstractOrganizationUser.__str__` reads `self.organization.name`, and the inline renders it per row. Both are in `select_related` now, with a test that renders the page at two member counts and asserts the count is identical — a fixed expected number would drift with unrelated admin changes. `10751ab0`.

## Medium 2 — change-roles GET

Confirmed. It now looks up the `OrganizationUser` rather than the `User`, so a non-member 404s on GET instead of only being refused on POST. Sourcery separately asked for a direct unit test of the service's non-member branch — with the view 404ing first that guard is defense-in-depth, which is exactly the case for testing it directly, so it's there. `cd04f923`.

## Medium 3 — editable `organization` / `user`

Confirmed, and it moved **down to #2352**, because that's the PR that introduced it: `main` had `OrganizationUserForm` with `exclude = ("user", "is_admin")`, and replacing that form with a bare `ModelAdmin` re-exposed both. Fixed there in `5cfbb8af`; this branch just adds `roles` to `fields`.

## Also fixed here rather than deferred

`changeOrganizationMemberRole` called `OrgRoleManager.replace_roles` directly and had the identical defect — granting an org admin Caseworker demoted them in the same breath, since `PermissionTemplateEnum` can't name `ORG_ADMIN` and so no caller could ask to keep it. It's pre-existing on `main`, but leaving a known-broken sibling of the thing this PR fixes didn't seem right. It now routes through `member_roles_replace`, keeping the `PermissionDenied` membership check in front so the API error is unchanged. Its docstring claimed the mutation could promote; it could only ever demote. `dbf81d61`.

Your `hasattr` finding from #2352 also landed here (`50a7f071`), across all three sites — `templates_for`, the inline's `permissiongroup`, and the `organizationowner` lookup the N+1 fix added.

---

Full suite green (1442 passed, 31 skipped), mypy clean on 383 files with the cache cleared, ruff clean. Every fix mutation-tested — revert it, confirm the test fails, restore.

PR #2356, opened 2026-08-20
fix(graphql): treat an unmatchable ID as not found, and stop coercing ids

### LINKED ISSUE #2337: int(data.id) on GraphQL IDs returns a 500 for a non-numeric id
## What happens

GraphQL `ID` is a string, so a client may send any string. Several mutations cast it with `int()` before anything validates it:

- [`teams/schema.py`](../blob/main/apps/betterangels-backend/teams/schema.py) — `update_team`, `delete_team` (three casts, including `DeletedObjectType(id=int(data.id))`)
- [`accounts/schema.py`](../blob/main/apps/betterangels-backend/accounts/schema.py) — `user_id=int(data.id)`

`int("abc")` raises `ValueError`, and `teams/schema.py` already documents the consequence in a neighbouring comment:

> A bare `ValueError` is not one of the exceptions strawberry-django turns into `OperationInfo`, so it surfaced as an internal server error instead of a denial.

So `updateTeam(data: {id: "abc"})` is a 500 rather than the denial that `updateTeam(data: {id: "999999"})` correctly produces.

## Why it isn't caught today

The existing tests use `"999999"` — numeric, so they exercise the handled "no such team" path and never reach the cast. Nothing sends a non-numeric id.

## Suggested fix

Treat an unparseable id the same as an id that names nothing, since a caller cannot tell the difference anyway and shouldn't be able to:

```python
try:
    pk = int(data.id)
except ValueError:
    raise PermissionDenied("You do not have permission to update this team.")
```

or push the coercion into the selector (`team_get`) so every caller inherits it. A parametrized test over `("abc", "", "1.5", "999999")` would cover the family.

## Related

Same class of bug as the malformed `X-Organization-ID` header fixed in #2310 — a client-supplied string reaching a coercion that assumes it is well-formed. That fix used `Organization._meta.pk.to_python()`, which is field-driven and would work here too.

Pre-existing, and small. Filed rather than folded into #2310 to keep that PR's scope.

### LINKED ISSUE #2343: A non-numeric teamId returns a 500 instead of a validation message
`teamId` is a GraphQL `ID`, so a client can send any string. A non-numeric one reaches `int()` unguarded and surfaces as a top-level error with `data: null`, rather than as an `OperationInfo` `VALIDATION` message like every other bad input on these mutations.

Reproduced against `fix/team-belongs-to-org`, on both paths that accept a team:

```
createNote(data: {purpose: "p", teamId: "abc"})
  -> {"data": null, "errors": [{"message": "invalid literal for int() with base 10: 'abc'"}]}

createNote(data: {purpose: "p", tasks: [{summary: "s", teamId: "abc"}]})
  -> same
```

The conversion happens in `common/graphql/utils.maybe_int_value`, which does a bare `int(maybe.value)`. The nested-task path never calls it and instead hits the same `ValueError` from Django inside `Team.objects.filter(pk="abc")`.

Both should return a validation message. The fix belongs with `maybe_int_value` so it covers every caller at once, which is why it is not in #2339 — that PR only adds the org-ownership rule on top.

No client can trigger this today: the team picker always sends an id it received from the API. It matters for API consumers, and for the error being legible in logs.

vecchp:
Fixes #2343. Also closes #2337.

Rebased onto `main` — the branch forked before #2352, which dropped `django-annoying`, so it no longer imported against a current environment.

## What the bug actually was

`teamId` is a GraphQL `ID`, which accepts **any string** by specification. The failure was not that we validated too little — it was that **our own code called `int()` on it** and raised from inside the ORM.

Those `int()` calls are also an assumption: that primary keys are integers. `NoteDataImport` and `ClientProfileDataImport` are already `UUIDField(primary_key=True)` — `Note` and `ClientProfile` themselves are still `BigAutoField` — so that is exactly the assumption a wider UUID move would have to undo.

So nothing converts an id any more. The value stays the string GraphQL parsed and Django's field coerces it for whichever column it is bound for.

**Removing the conversion is half the fix.** `Task(team_id="abc").full_clean()` records Django's own field error in `clean_fields()` — but `clean()` then ran `validate_team_in_org`, whose `Team.objects.filter(pk="abc")` raised `ValueError` and **discarded** the error already collected. Both had to go for the mutation to answer with an `OperationInfo` VALIDATION message.

## Unboxing a `Maybe`

`maybe_int_value` existed only for the conversion, so it is deleted rather than kept in a pass-through form. Its callers unbox inline:

```python
team_id = data.team_id.value if data.team_id else None
```

Truthiness rather than `is not None`, because absence has two spellings and only truthiness is correct for both — Strawberry injects `= None` for a bare `Maybe[T]`, while a hand-written `Maybe[T] = UNSET` is *not* `None` when the field was never sent, and `.value` on `UNSET` is an `AttributeError`. `shelters/types/inputs.py` carries about 45 of the second kind.

`common/graphql/utils.py` is left holding `get_object_or_permission_error` alone.

## Where a lookup can still fail: it is a miss, not a crash

An id the column cannot hold names no row — which is the answer already given for a row that does not exist. `teams/selectors.py` has worked this way since #2340; this applies the same shape to the three other shared helpers that own an ID→lookup conversion.

All three ask the same question, so they share one answer. `can_match()` and
`matchable_values()` in `common/utils.py` put it to the field itself —
`field.get_prep_value(value)`, the call `Lookup.get_prep_lookup` makes, so they
reject exactly what the lookup would have raised on. That is Django's own
coercion, correct for a UUID column without a special case, and it is why there
is no `int()` in the new code either.

| Helper | Behaviour | Reach |
| --- | --- | --- |
| `get_object_or_permission_error` | `PermissionDenied`, as for a missing row | 11 call sites |
| `get_by_pk_or_not_found` | `ObjectDoesNotExist`, as for a missing row | 7 call sites in `shelters` |
| `team_get` | `None`, as for another org's team | `updateTeam`, `deleteTeam` |
| `validate_team_in_org` | says nothing — a value that is not an id is not this rule's business | notes + tasks, incl. nested tasks |
| `make_in_filter` | drops unmatchable elements from the `__in` | 9 ID filters across notes, tasks, hmis |

The first three differ only in the exception they raise, so they share
`get_or_none(queryset, pk)` — `can_match`, then `queryset.filter(pk=pk).first()`,
which raises nothing and so needs no `except`. Returning `None` rather than
choosing an exception is what lets them share it. `get_object_or_permission_error`
moves off `.get()` as a result, so `MultipleObjectsReturned` stops being raised;
it is a primary-key lookup, so more than one row is impossible.

Asking beats catching. `qs.get(pk=...)` raises `ValueError` from anywhere in query construction, so catching it around the call would report a genuine bug as `PermissionDenied` with the traceback discarded. After this, **no `except ValueError` survives anywhere in the path** — `get_object_or_permission_error`, `team_get` and `validate_team_in_org` each lost theirs.

`get_by_pk_or_not_found` sits immediately above the predicate in `common/utils.py` and had the same hole: it went straight to `queryset.filter(pk=pk)`, so `shelterOccupancyMetrics(shelterId: "abc")` answered with a raw Python error. Seven callers reach it — `shelter_get`, `room_get`, `bed_get`, `reservation_get`, and the photo and clone services.

The predicate lives in `common/utils.py` rather than `common/graphql/utils.py` because `teams/validators.py` is one of its callers: that module exists so `notes.models` and `tasks.models` can validate without depending on a service, and pointing it at the GraphQL layer would invert the same layering the other way. It is a Django field question, not a GraphQL one.

Dropping unmatchable values rather than rejecting the whole filter is deliberate: `["<real>", "abc"]` filters on the id that can match, and `["abc"]` matches nothing rather than silently becoming *no filter*.

## Two things that make coercion the wrong answer, not just a worse one

Found while inventorying every `ID` input position:

- `clients/schema.py` keys a dict by the **raw string** id (`item_updates_by_id[str(item.id)]`), reached by every nested list on `UpdateClientProfileInput`. Coercing ids to `int` turns that into a `KeyError`.
- `make_in_filter`'s enum branch is guarded on `isinstance(v, str)`. Coercing to `int` makes every `authors` / `organizations` / `teamIds` filter raise `AttributeError: 'int' object has no attribute 'name'`.

## One message, not two

`full_clean()` runs `clean()` even after `clean_fields()` has already failed, merging both into one `ValidationError`. So `teamId: "abc"` used to answer with Django's field error **and** a second message about a rule the value never got far enough to break:

```
team: 2 message(s)
   - “abc” value must be an integer.
   - The selected team does not belong to this organization.
```

`validate_team_in_org` now returns for a value that is not an id, leaving the field error to speak for itself, and the tests assert the message count so a second one cannot come back.

A well-formed id that names no row still produces two — `ForeignKey.validate()`'s *"is not a valid choice"* alongside the org rule. Both are true there, and that behaviour predates this PR.

## `asdict` already handled the update paths

`update_note`, `update_task` and `create_task` each patched `team_id` into the dict they had just built with `strawberry.asdict`. Those patches were no-ops: `asdict` omits an absent `Maybe`, keeps an explicit null as `None`, and unwraps `Some(x)` to `x` — the three-way behaviour the update mutations need, and what `shelters` has relied on alone all along. They are gone.

`update_task` and `delete_task` were reaching the ORM directly — a bare `qs.get()` and a hand-rolled `DoesNotExist` branch — so neither got the treatment the rest of this PR describes. Both now use `get_object_or_permission_error`.

## No schema change

`schema.graphql` is **byte-identical** — verified with `nx run betterangels-backend:validate-graphql-schema`. No `graphql-inspector:approved-breaking-change` label, no regenerated client artifacts.

For the record, since it came up: a named scalar (`DatabaseID`) was considered and rejected. It would have retyped ~110 SDL input fields, and it bakes in the same `int` assumption. Attaching validation to the built-in `ID` instead is **impossible** — graphql-core raises `TypeError: Redefinition of reserved type 'ID'`.

## Tests

**1533 passed, 31 skipped** on a freshly created database. mypy adds no errors over `main` (a control run of unmodified `main` in its own worktree reports the same set); `ruff check` and `ruff format --check` clean.

Fifteen tests. Behavioural, via `execute_graphql`:

| Case | Expected |
| --- | --- |
| `createNote(data: {teamId: "abc"})` | `OperationInfo` VALIDATION on field `team`, naming the value |
| `createNote(data: {tasks: [{teamId: "abc"}]})` | same — the nested path |
| `createTask(data: {teamId: "abc"})` | same |
| `updateNote(data: {id: "abc"})` | asserts the response is **identical** to `id: "99999999"` |
| `updateTask` / `deleteTask` with `id: "abc"` | same parity |
| `tasks(filters: {teamIds: ["<real>", "abc"]})` | no error, same single result |
| `tasks(filters: {teamIds: ["abc"]})` | no error, **zero** results — fails closed |
| `shelterOccupancyMetrics(shelterId: "abc")` | the same not-found message a shelter in another org gets |

They assert the message's `kind` and `field` and that it names the offending value, not Django's exact localized sentence.

Plus `common/graphql/tests/test_id_handling.py` on `_matchable_values` directly, which is the only way to reach a UUID column: no `ID` input in the schema currently targets one, so the claim cannot be proved end-to-end. `NoteImportRecord.import_job` points at a UUID primary key, and a well-formed UUID string is kept while a malformed one is dropped — with no `int()` in the path.

Mutation-tested, each helper independently:

| Mutation | Fails |
| --- | --- |
| restore `int()` in `create_note` | **1** — the unmatchable-team test |
| revert `make_in_filter` to the raw list | **2** — both filter tests |
| `validate_team_in_org` opines on a value that is not an id | **3** — the message-count assertions on all three create tests |
| `matchable_values` falls back to the raw list when nothing matches | **1** — the fail-closed test |
| `can_match` always returns `True` | **11** — across mutations, filters and the helper's own tests |
| `get_or_none` drops its `can_match` guard | **5** — note, task, team-selector and shelter cases together |
| `team_get` goes back to a raw filter | **1** — the unmatchable-pk selector test |

That last one is worth calling out: an explicit `Q(pk__in=[])` guard for "every value was unmatchable" turned out to be dead code — Django already renders `__in []` as matching nothing — so it is not here. The fail-closed test still holds, and still fails if the helper returns the raw list instead.

**Not mutation-tested, deliberately:** the truthiness idiom. Every note/task/team input uses the bare `Maybe[T]` spelling, where truthiness and `is not None` behave identically; the difference only shows on the `= UNSET` fields in `shelters`, which no resolver here touches. A test for it would be asserting `bool(UNSET) is False`, which is Strawberry's behaviour rather than ours.

## Not covered

19 top-level `pk: ID!` / `id: ID!` arguments — `note(pk:)`, `task(pk:)`, `shelter(pk:)`, `deleteHmisNote(id:)`. strawberry-django resolves those inside its own field resolver, so no helper here sees them, and a bad value still surfaces a raw Python message. Retyping them is not an option: deployed documents declare the variable type. If it becomes worth closing, one `SchemaExtension` alongside `DjangoOptimizerExtension` covers all 19 without touching the schema.

**And the larger half: 35 resolver bodies that put a client-supplied id straight into a query**, plus 5 hand-built `pk__in` filters in `shelters/services`. This PR fixes the shared helpers, not every caller:

| Location | Sites |
| --- | --- |
| `hmis/schema.py` | 12 |
| `clients/schema.py` | 11 |
| `tasks/schema.py` | 4 |
| `referrals/schema.py` | 4 |
| `accounts/schema.py` | 3 |
| `notes/schema.py` | 1 |
| `shelters/services/{shelter,room,bed,reservation}.py` — `pk__in` | 5 |

Every model behind them is `BigAutoField`, so `.get(pk="abc")` raises `ValueError`, which `MUTATIONS_DEFAULT_HANDLE_ERRORS` does not convert.

They are not folded in here because there is no single answer to route them through. Each is already `try: ….get(pk=id) except Model.DoesNotExist:` with its **own** chosen outcome — `ObjectDoesNotExist` in `hmis`, `PermissionError` in `clients`, a field-keyed `ValidationError` in `referrals` — and the two shortcuts are both wrong: widening each `except` to `Model.DoesNotExist, ValueError` is the catch-don't-ask pattern this PR removes from three helpers, and reusing `get_by_pk_or_not_found` would overwrite each site's error. The shape that fits is a `get_or_none(queryset, pk)` returning `Optional[M]` that asks `can_match` first and leaves each site its own `if obj is None: raise` — a helper decision plus ~35 conversions and their tests, in its own PR.


vecchp:
Cleaned up further — the change is now **9 insertions against 22 deletions** smaller than the version above, with the same coverage. Four trims, each mutation-tested again afterwards:

**`make_in_filter` keeps its original signature shape.** I had copied `make_m2m_in_filter`'s `(self, queryset, value, prefix) -> Tuple[QuerySet, Q]` form. strawberry-django accepts `(queryset, value, prefix) -> Q`, so the only actual change to that line is `info` → `queryset` — no injected `self`, no tuple return, nothing to unpack at the call site.

**`maybe_value` is one line.** `if maybe is None or maybe.value is None: return None` was redundant with the return that follows it: when `maybe.value` is `None`, `maybe.value` *is* the answer. It reduces to `return maybe.value if maybe is not None else None`.

**Dropped the `FieldDoesNotExist` branch in `_matchable_values`.** All 13 `make_in_filter` call sites resolve on their own model — verified against `_meta.get_field` for each — so it was unreachable. Worse than unreachable: if a nested filter ever did reach it, passing the values through unchanged is precisely the original bug, silently. Letting `get_field` raise makes that a loud failure while the feature is being written. Its test went with it.

**Folded a comment into the docstring.** `get_object_or_permission_error`'s docstring already argues that a missing row and a denial are the same answer; the `ValueError`/`TypeError` case is one more sentence there rather than a comment repeating it at the `except`.

Re-verified: **1403 passed, 31 skipped**; mypy clean over 383 files; ruff clean; `schema.graphql` still byte-identical. All five mutations still fail exactly one or two tests each:

| Mutation | Fails |
| --- | --- |
| `int()` back in `maybe_value` | `test_create_note_rejects_an_unmatchable_team_id` |
| narrow `get_object_or_permission_error` to `ObjectDoesNotExist` | `test_update_note_denies_an_unmatchable_id` |
| `make_in_filter` skips `_matchable_values` | both filter tests |
| `validate_team_in_org` without the catch | both `create_note` tests, incl. nested |
| `_matchable_values` falls back to the raw list | `..._of_only_unmatchable_ids_matches_nothing` |

vecchp:
Good catch, and it went further than the duplication you spotted: **`strawberry.asdict` had already produced that exact result before the `if` ran.** The whole block was a no-op.

From strawberry 0.319.0, `types/object_type.py`:

- `_prepare` line 483 — an absent `Maybe` becomes `UNSET`
- `_asdict_dict_factory` — drops `UNSET`, so the key is **omitted from the dict**
- `_prepare` lines 469-471 — unwraps `Some(x)` → `x`, recursively, including inside lists

| input | `data.team_id` | `asdict(data)` |
| --- | --- | --- |
| `{id: "1", purpose: "p"}` | `None` | `{'id': '1', 'purpose': 'p'}` — **no `team_id` key** |
| `{id: "1", teamId: "7"}` | `Some('7')` | `{..., 'team_id': '7'}` |
| `{id: "1", teamId: null}` | `Some(None)` | `{..., 'team_id': None}` |

Which is exactly the three-way behaviour the mutations need. `shelters/schema.py` has relied on `asdict` alone for its five `Maybe[ID | None]` fields all along; notes and tasks were the outliers.

The comment above each block was also backwards. It said the guard stopped an unmentioned team being nulled — but `asdict` had already decided that, so the guard was only choosing whether to rewrite the value with itself.

So: both guarded blocks are gone, plus the unconditional one in `create_task`. `maybe_value` keeps the two call sites that pass explicit keyword arguments rather than a dict (`create_note`, `import_note`), and its docstring now sends dict-building callers to `asdict`.

## On the truthiness question

`if data.team_id:` was *correct*: `Some.__bool__` returns `True` unconditionally, so truthiness means "was sent" and `Some(None)` / `Some(0)` / `Some("")` are all truthy. Strawberry defines it that way deliberately. Moot now, but worth recording — the equivalent `is not None` check elsewhere is not more correct, just more explicit.

## Something this turned up

Chasing the call sites found that **`update_task` and `delete_task` never went through `get_object_or_permission_error` at all** — a bare `qs.get(pk=data.id)` and a hand-rolled `except Task.DoesNotExist` respectively. So this PR's own claim was false on the task side: `updateTask(id: "abc")` still leaked a raw error, and a task the caller could not see raised an uncaught `DoesNotExist`. Both now use the helper, which also deletes the try/except it was duplicating. Three tests added for the task mutations, mirroring the note ones.

mypy then caught that `delete_task_fixture(task_id: int)` was annotated wrong — the mutation it sends declares `$id: ID!`, so it is a string. Widened to `int | str`, matching `team_get`.

## Verification

**1406 passed, 31 skipped** (1403 + the 3 new task tests). mypy clean over 383 files; ruff clean; `schema.graphql` still byte-identical.

The deletions are provable rather than hopeful, because the behaviour was already pinned:

- `test_update_note_omitted_team_id_preserves_team` / `..._explicit_null_team_id_clears_team`
- `test_update_task_omitted_team_id_preserves_team` / `..._explicit_null_team_id_clears_team`
- `test_nested_task_team_id_is_assigned` — the recursive unwrap inside the `tasks` list

All five unchanged. Every mutation test still fails the right cases, and the new task tests widen two of them:

| Mutation | Fails |
| --- | --- |
| `int()` back in `maybe_value` | `test_create_note_rejects_an_unmatchable_team_id` |
| narrow `get_object_or_permission_error` | note update **+ `update_task` + `delete_task`** |
| `make_in_filter` skips `_matchable_values` | both filter tests |
| `validate_team_in_org` without the catch | both note tests **+ `create_task`** |
| `_matchable_values` falls back to the raw list | the fail-closed filter test |

vecchp:
Made it generic — and there was evidence for generic specifically, which is the part I had missed: `teams/schema.py:73-74` was hand-writing the same expression for two other types, and the `ID`-typed signature is exactly why it could not call the helper.

```python
name=data.name.value if data.name else None,                 # Maybe[str]
is_active=data.is_active.value if data.is_active else None,  # Maybe[bool]
```

```python
V = TypeVar("V")

def maybe_value(maybe: Maybe[V]) -> V | None:
    return maybe.value if maybe is not None else None
```

Four call sites over three types now, replacing two hand-written copies. Verified under `mypy --strict`: `Maybe[ID | None]` → `ID | None`, `Maybe[str]` → `str | None`, `Maybe[bool]` → `bool | None`.

## On whether a library already does this

Searched properly rather than assuming. Nothing usable:

| Candidate | Why not |
| --- | --- |
| `strawberry.types.maybe` | Exports only `Maybe` and `Some`. `Some` has no `unwrap`; strawberry's own code writes the ternary inline. |
| `strawberry_django...parse_input` | Does handle a plain input (`resolvers.py:271`), but maps an **absent** `Maybe` to `None` rather than omitting it — which destroys the distinction `update_note`/`update_task` depend on. Needs an `info`, returns an untyped dict, exported from neither `__init__`. |
| `strawberry_django.mutations.update` | Cannot take a `Maybe` input. `get_vdata` is `vars(data).copy()`, so an absent `Maybe` is present as `None`; `prepare_create_update` (`resolvers.py:282`) skips only `UNSET`; so an absent field would be **written as null**. It also writes straight to the model, bypassing `team_update`'s name stripping and per-org uniqueness message. |
| `apply_maybe` (`475bfc73`, unmerged) | Written when resolvers had to strip the wrapper out of `asdict` output and absent meant `UNSET`. `asdict` does that itself now. |

The repo already draws this line and the two sets are **disjoint**: `mutations.update` appears only in `clients/schema.py`, whose inputs have **zero** `Maybe` fields; every app with `Maybe` inputs hand-writes its resolvers. So `strawberry.asdict` is the tri-state-preserving path (update), and unboxing is the collapse path (create) — with no public helper for the latter.

## Also recorded

`Maybe[T]` vs `Maybe[T | None]` is not stylistic. `MaybeNullValidationRule` (registered by default at `strawberry/schema/schema.py:137`) **rejects an explicit null for `Maybe[T]`**. So `UpdateTeamInput.name: Maybe[str]` means "cannot be cleared" and `UpdateNoteInput.team_id: Maybe[ID | None]` means "null clears it" — the annotation already says which. The docstring says that now instead of implying both are always possible.

## Verification

**1406 passed, 31 skipped** — unchanged. mypy clean over 383 files; ruff clean; `schema.graphql` byte-identical.

Two mutation checks:

- passing the raw `Maybe` through instead of unboxing fails **4** teams tests, including `test_update_team_mutation_partial` — so the substitution is genuinely covered
- restoring `int()` inside `maybe_value` still fails `test_create_note_rejects_an_unmatchable_team_id`

One note on the run: an intermediate full-suite pass reported 19 failures at 372s. That was a second suite running concurrently against the shared Postgres test database — every reported failure passes in isolation, and the suite is back to 1406 at 150s once the other run finished. Nothing to do with this change.

apps/betterangels-backend/tasks/tests/test_queries.py
```diff
@@ -241,6 +241,48 @@ def test_task_query_legacy_team_selection_still_resolves(self) -> None:
             {"id": str(self.org_1_team_1.pk), "slug": None, "name": self.org_1_team_1.name},
         )
 
+    def test_tasks_query_teams_filter_ignores_an_unmatchable_id(self) -> None:
+        """An id the column cannot hold matches nothing, so it changes nothing.
```
mikefeldberg: not sure what "it changes nothing" means. This is query test so nothing would have been "changed" in any case. Unless it's another of claude's "only relevant/meaningful in the immediate context" notes

apps/betterangels-backend/tasks/tests/test_queries.py
```diff
@@ -241,6 +241,48 @@ def test_task_query_legacy_team_selection_still_resolves(self) -> None:
...
+            self.get_tasks_query("id"),
+            {"filters": {"teamIds": [str(self.org_1_team_2.pk), "abc"]}},
+        )
+
+        self.assertIsNone(response.get("errors"))
+        self.assertEqual(response["data"]["tasks"]["totalCount"], 1)
+        self.assertEqual(response["data"]["tasks"]["results"][0]["id"], task_id)
+
+    def test_tasks_query_teams_filter_of_only_unmatchable_ids_matches_nothing(self) -> None:
+        """The filter must fail closed.
```
mikefeldberg: `The filter must fail closed.` 😭

mikefeldberg (approved):

PR #2357, opened 2026-08-20
feat(scalars): one boundary for the Date, Time and PhoneNumber scalars

vecchp:
## Why

Every place a `Date` crossed the GraphQL boundary was hand-rolled, in two mutually incompatible ways:

- `format(value, 'yyyy-MM-dd')` — local parts, correct — in `useShelterOccupancyMetrics`
- `date.toISOString().split('T')[0]` — UTC-based, and so the wrong day for anyone east of UTC — in three form files

Reading back had six spellings, display four more, and the `Time` scalar had three independent `HH:mm:ss` converters across two react apps.

The correct, timezone-tested implementation already existed — quarantined in `libs/expo/shared/utils`, which pulls `expo-file-system`, `expo-image-manipulator`, `react-native-mmkv` and `react-native`. Importing it from a react app would drag native deps into a web bundle, so the react side reimplemented it instead.

## What this does

Moves those primitives into a framework-free `@monorepo/shared/scalars` (git records them as renames, so the history follows) and gives each scalar one entry point:

| | |
| --- | --- |
| `Date` | `toDateString` / `fromDateString` / `formatDateString` |
| `Time` | `toTimeString` / `fromTimeString` / `formatTimeString` |
| `PhoneNumber` | `toPhoneParts` |

The branded string types come along and gain a **shared** brand key, so a `DateString` can no longer be passed where a `UUIDString` is expected. The brand stays optional, so a plain `string` still assigns in and adoption is incremental.

`formatDateStatic` and `parseToDate` are retired. All twelve call sites passed `inputFormat: 'yyyy-MM-dd'`, so the parameter was dead configurability and the functions were `formatDateString` and `fromDateString` under other names. `toLocalCalendarDate`'s never-used `inputFormat` branch goes with them, and the primitive becomes internal.

## Phone: one parse, and no validity gate

`toPhoneParts` returns `{ formatted, extension, display, dial }` from a single `libphonenumber-js` parse — `formatted` for callers that render the extension in its own element, `display` for callers that render one string, `dial` for the `tel:` href. Deriving them together is the point: display and dial can no longer disagree about where the number ends and the extension begins.

Two library APIs do the work that was previously hand-rolled:

- `format('NATIONAL', { formatExtension })` is the library's own hook for suppressing the ` ext. NN` suffix, so getting the extension-free form needs no second parse and no string surgery.
- `formatNational` already degrades to the bare digits for a number the metadata will not vouch for, so **there is no `isPossible()` gate.** That gate was actively wrong: a seven-digit local number and `911` are both `isPossible() === false`, and were being routed through the "unrecognised" branch and kept out of the dial path.

`dial` stays `digits,ext` rather than `getURI()`'s RFC 3966 `;ext=`. iOS documents `tel:` support for digits, `+`, `*`, `#`, `,` and `;` and does not parse `ext=`; Android's dialer honours the pause characters rather than RFC parameters. **This is the one claim here I could not test on a handset** — it is from documented `tel:` behaviour, so it is worth a second opinion.

## Time needed more than moving

This is the one scalar with no library to lean on, and it had three real defects:

- **`formatTimeString` anchored its `Date` to today.** On a spring-forward day a 2:00 AM curfew rendered as 3:00 AM — and rendered *identically* to a 3:00 AM one, because the hour does not exist and the runtime moves it. It now builds on a reference day with no UTC-offset change in any of the 418 IANA zones.
- **`fromTimeString` claimed to accept `HH:mm` and `HH:mm:ss` and reject the rest, but only rejected `NaN`.** `"8"`, `":30"`, `"99:99"`, `"12:60"` and `"0x10:00"` all parsed (`Number` takes hex and exponent notation), and `"99:99"` formatted back out as a plausible-looking 4:39 AM. It now matches a pattern and bounds the result to one day.
- **`toTimeString` clamped to `"24:00:00"`.** Strawberry's `Time` scalar accepts that and silently returns `time(0, 0)` while rejecting `"25:00:00"` — so a window ending at the end of the day came back ending before it started. It clamps to 23:59, which also makes the `TimeRangeField` round trip survive.
- **`fromTimeString` was too strict about seconds.** Tightening the pattern to reject junk also
  rejected `"22:00:00.500000"` — which is what Strawberry's `Time` scalar emits when the stored
  value carries microseconds, and `curfew`, `start_time` and `end_time` are all `TimeField`. That
  would have dropped an operating-hours window and rendered the raw string as a curfew. Fractional
  seconds are now accepted; every other rejection still holds.

## Call-site fixes that followed

- `PhoneNumberBtn` announced raw digits to screen readers (`"Call 2135551234,22"`), and underlined a number that opened an empty dialer when it could not be parsed.
- `GeneralInfo` and `Actions` built their `tel:` hrefs out of the *display* string — parens, spaces and `" ext. "` included. They use `dial` now, and render plain text rather than a dead link when there is nothing to dial.
- `scheduleUtils.normalizeWindow` read an unparseable time as midnight. It now drops the window it cannot place, rather than reporting "open at midnight".
- `ReportCharts` labelled unparseable dates `''`, and that label is the chart's grouping key — every bad day merged into one bar. It keeps the raw value instead.
- `libphonenumber-js` is declared in `apps/betterangels/package.json`: the library now reaches the Expo app through `PhoneNumberBtn`, and Expo autolinking wants it listed. `betterangels:expo-doctor` fails without it.

## Notes for review

- **Correction to an earlier version of this description:** it claimed `getEffectiveWindows` had no consumers. That was wrong — `OperatingHours.tsx` → `getWeeklySchedule` → `getEffectiveWindows` is live, and renders the shelter's operating-hours UI. My grep had excluded matches inside `scheduleUtils.ts` itself, which hid the internal call. Please review those hunks as live code. (The 24:00 → 23:59 clamp does not reach it: `wrapIntoDay` returns 0–1439, so the clamp never fires there.)
- **`TimeRangeField` genuinely has no consumers**, though — its only other hit in the repo is an unrelated Python class of the same name in `shelters/widgets.py`. It is converted along with everything else, but nothing exercises it. It is also one of five unused members of the ten-component kit under `components/form/`, which is why it is converted rather than deleted.
- `scheduleUtils` wraps minutes past midnight while `TimeRangeField` clamps them. That is genuinely different intent, so the wrap stays a local helper at its call site rather than becoming a flag on the shared function. (The clamp target is now 23:59, not 24:00 — see above.)
- `isoToDateSafe` stays public: `renderValue.ts` needs it for a string of unknown shape (ISO first, then a custom format), which the branded boundary does not cover. It keeps a small guard that rejects partial ISO, because `parseISO` turns `"2026"` into January 1st.
- No codegen change here, so no generated file moves. Typing the scalars themselves is #2277.

## Testing

- 98 tests in the new lib, run under `UTC`, `America/Los_Angeles`, `Asia/Kathmandu`, `Pacific/Chatham`, `Pacific/Kiritimati` and `Australia/Lord_Howe` — the +14:00, +12:45 and half-hour-DST zones are where `toISOString`-based serialization and naive `Date` construction break.
- Each of the three Time fixes was mutation-tested: reverting them fails 3, 11 and 2 tests respectively. The daylight-saving test sets its own zone, so it catches the regression even though the harness runs in UTC — the first version of it did not, and passed against the reverted code.
- `nx affected -t typecheck,lint,test` clean for 30 projects. The backend suite failed once with a Postgres deadlock in `shelters/tests/baker_recipes.py`; nx flagged it flaky, the retry was clean, and every failing test passes in isolation. This PR touches no Python.
- `nx run-many -t generate-graphql-types --all` then `git diff --exit-code`: no generated file changes.
- **Not yet done: nothing has been rendered.** The tests cover the strings; the display, the `tel:` links, the curfew and the date-formatting call sites have not been looked at in a browser or on a device. There is a checklist in the comments covering each surface.
## Summary by Sourcery

Centralize Date, Time, and PhoneNumber scalar handling in a shared, timezone-safe library and migrate application call sites to it.

New Features:
- Add a framework-independent shared scalar library with unified Date, Time, and PhoneNumber conversion and formatting APIs.
- Provide shared phone parsing outputs for display, extensions, and dial links.

Bug Fixes:
- Correct local-date serialization and parsing to prevent timezone-related day shifts.
- Make time parsing strict and timezone/DST-safe, and preserve valid end-of-day time ranges.
- Prevent invalid phone links, misleading accessibility labels, and inconsistent extension handling.
- Skip unparseable schedule windows and preserve raw unparseable chart dates instead of merging them.

Enhancements:
- Replace duplicated application-specific scalar utilities and call sites with the shared scalar APIs.
- Add mutually exclusive branded scalar string types while retaining compatibility with plain strings.
- Retire obsolete date and phone utility implementations from the Expo-specific libraries.

Build:
- Declare libphonenumber-js as an application dependency and add the shared scalars library project configuration.

Tests:
- Add comprehensive Date, Time, and PhoneNumber scalar tests covering timezone, daylight-saving, validation, extension, and round-trip behavior.


vecchp:
## Manual render check

Everything here is covered by unit tests at the string level, but nothing has been rendered.
Listing it out because the surface is wider than the phone screens — the date helpers changed
under eight Expo cards too.

### Shelter web (`libs/react/shelter`)

Needs a shelter with a plain 10-digit number, one with an extension, and one with something
malformed stored in `phone`.

- [ ] **`Actions`** (the Call button under the header) — label reads as a formatted number, and
      the link is `tel:2135551234,22`, *not* `tel:(213) 555-1234 ext. 22`. This is the one that
      changed: the href used to be built from the display string.
- [ ] **`GeneralInfo`** (phone row) — same href change. With a malformed number the row should
      render as plain text with no link, rather than a link to `tel:`.
- [ ] **`Restrictions`** (curfew) — a curfew stored as `22:00:00` shows `10:00 PM`. Worth also
      trying one stored as `22:00` (no seconds), which the old `parse(…, 'HH:mm:ss')` rejected
      and showed raw.

### Expo (`libs/expo/betterangels`)

- [ ] **`PhoneNumberRows`** (Contact Info card) — display string and the dial action.
- [ ] **`PhoneNumberBtn`** via **`RelevantContactCard`** — extension renders in its own element
      beside the number; with an unparseable number it should render as plain unlinked text with
      no underline, where it previously underlined it and opened an empty dialer. Also worth a
      pass with VoiceOver/TalkBack: the label should announce a grouped number, not a raw digit
      run.
- [ ] **`ClientSummaryContact`** and **`ClientSummaryGeneral`** — both moved to a single parse.
- [ ] **Date display on the client cards** — `PersonalInfoCard` (both the regular and HMIS one),
      `HouseholdMemberCard`, `ClientCardBase`, `ClientCardHmis`,
      `ClientSummaryLivingSituation`, `NoteEditHmis`. These should be behaviour-identical
      (`formatDateStatic` → `formatDateString`), but DOB is exactly where the UTC-vs-local
      off-by-one lived, so a client with a known date of birth is worth one look. Check a
      profile with **no** DOB too — it should render empty, not `Invalid Date`.

### Shelter operator (`libs/react/shelter-operator`)

- [ ] **`ReportCharts`** — x-axis labels still read `Jun 1` etc. across a multi-month range.

`TimeRangeField` also changed, but it has no consumers, so there is nothing to look at.

### Not checkable here

The dial string keeps the `digits,ext` form rather than RFC 3966's `;ext=`. The reasoning is
from documented `tel:` behaviour on iOS and Android, not from a handset — so if anyone is
testing on a real phone, an extension number actually pausing and sending the extension is the
one claim in this PR I could not verify.

libs/shared/scalars/src/lib/date/dateString.ts
```diff
@@ -0,0 +1,36 @@
...
+ * `new Date('2026-05-05')` would anchor to UTC midnight, which renders as the
+ * previous day anywhere west of UTC.
+ */
+export function fromDateString(value?: DateString | null): Date | undefined {
+  return toLocalCalendarDate(value ?? undefined);
+}
+
+/** Format a `Date` or `DateTime` scalar for display. `''` when unparseable. */
+export function formatDateString(
+  value: DateString | DateTimeString | null | undefined,
```
tglaz, line 30: If we’re moving toward stricter scalar typing, the name `formatDateString` may get confusing since this accepts both `DateString` and `DateTimeString`.

Not a huge deal I suppose, and we don't necessarily need/want separate formatting functions here. but maybe rename to something like `formatScalarDate`?

libs/expo/shared/ui-components/src/lib/PhoneNumberBtn/PhoneNumberBtn.tsx
```diff
@@ -1,48 +1,51 @@
...
+  const content = (pressed: boolean) => (
+    <View style={{ flex: 1, flexDirection: 'row' }}>
+      <TextBold
+        textDecorationLine={dial ? 'underline' : 'none'}
+        color={pressed ? Colors.PRIMARY_LIGHT : Colors.PRIMARY_EXTRA_DARK}
+        size="sm"
+      >
+        {label || formatted}
+      </TextBold>
+      {extension && (
```
tglaz, line 23: is this part redundant if you use `display` above instead of `formatted`?

tglaz (approved):

libs/shared/scalars/src/lib/date/dateString.ts
vecchp: Good catch — renamed to your suggestion, `formatScalarDate`, in 1dfbbdb.

I went looking for a reason to narrow the signature to `DateString` instead and keep the old
name, since every current call site does pass a `Date`: `dateOfBirth`, `unhousedStartDate`,
`birthDate` and both metrics `date` fields are all `Date` in the schema. But narrowing is the
wrong call — there are seven `DateTime` display sites still on raw date-fns that are the obvious
next callers:

```
FileCard.tsx            format(new Date(createdAt), 'MM/dd/yyyy')
TaskCardBody.tsx        format(new Date(createdAt), …)
NoteCardFooter.tsx      format(new Date(interactedAt), …)
NoteCardFooterHmis.tsx  format(new Date(interactedAt), …)
FileScreenComponent.tsx
TaskUpdatedAt.tsx
FileInfoHmis.tsx
```

So the union earns its place and the name was the thing to fix.

One note on naming: I first went with `formatDate`, then backed it out — `date-fns` exports
`formatDate` itself, and these files already `import { format } from 'date-fns'`, so it would
have been the same shadowing trap `parsePhoneNumber` was against libphonenumber-js. Yours avoids
that.

Those seven sites are worth their own PR, incidentally. `interactedAt` is nullable, and
`new Date(null)` is the epoch — so a note without one renders **01/01/1970** today, and an
undefined throws `RangeError: Invalid time value`. `formatScalarDate` returns `''`.

libs/expo/shared/ui-components/src/lib/PhoneNumberBtn/PhoneNumberBtn.tsx
vecchp: Not redundant, though it took me a minute to be sure why. The two elements differ in exactly one
prop: the first carries `textDecorationLine`, the second doesn't — so the underline covers the
number and stops before the extension. `display` in a single element would underline
`" ext. 22"` too. That split is pre-existing; `main` has the identical structure.

It's also the dominant pattern rather than a quirk of this component. Three of the four places
that render a phone number keep them separate, and one of those bakes it into another
component's props:

| call site | renders |
| --- | --- |
| `PhoneNumberRows` | one string, via `display` |
| `PhoneNumberBtn` | number underlined + extension not |
| `ClientSummaryContact` | same |
| `ClientSummaryGeneral` → `PressablePanelContainer` | `title` and `extension` as separate props |

So collapsing this one would make it the odd one out, and `PressablePanelContainer` couldn't
follow without an interface change.

**Your question did find something real, just not redundancy.** There were three spellings of the
separator: `PhoneNumberRows` renders `display`'s `" ext. "`, while this component and
`ClientSummaryContact` had a literal `" ext."` with no space — so `PhoneNumberBtn` disagreed with
its own `accessibilityLabel`, which uses `display`. That's my fault: routing `PhoneNumberRows`
through `display` changed its spacing without my propagating it. Settled on `" ext. "`
everywhere in 1dfbbdb.

One thing I've left alone, since it predates this PR and is a visual change: the
`PressablePanelContainer` path renders the extension **bare** — no `ext.` label — and its `View`
has no `flexDirection`, so it defaults to column. A client summary panel shows

```
CONTACT
(213) 555-1234
22
```

A bare `22` on its own line. Happy to fix it here if you'd rather not leave it, otherwise it
wants its own PR.

vecchp: Correcting myself on the table above — I said `ClientSummaryGeneral` → `PressablePanelContainer`
keeps them separate, which is right, but I also told you elsewhere in this PR that
`getEffectiveWindows` had no consumers. **That was wrong.**
`OperatingHours.tsx` → `getWeeklySchedule` → `getEffectiveWindows` is live and renders the
operating-hours UI. My grep had excluded matches inside `scheduleUtils.ts` itself, which hid the
internal call, so please review those hunks as live code rather than mechanical. Fixed in the
description too.

Chasing that down turned up a real bug in my own change, now fixed in 62f3a07: the tightened
`fromTimeString` pattern rejected `"22:00:00.500000"`, and that is precisely what Strawberry's
`Time` scalar emits when the stored value has microseconds — `curfew`, `start_time` and
`end_time` are all `TimeField`, so it is representable. On that live path it would have dropped
an operating-hours window silently, and rendered the raw string as a curfew. Fractional seconds
are accepted now; every other rejection still holds.

Two other things went with it, both prompted by your question here:

- **`PressablePanelContainer` now renders `ext. 22`** instead of a bare `22`. Label only — I left
  the column layout alone, so it is still its own line, just no longer contextless.
- **`formatPhoneNumber` and `toPhoneDialString` are gone.** Once every call site had moved to
  `toPhoneParts` they had no callers left outside their own tests — every real caller needs two
  or more parts, so the single-value accessors had no natural user. `toPhoneParts` is the only
  phone export now.

`PhoneNumberBtn`'s `label` prop went too. Nothing passed it, and it was what allowed
`accessibilityLabel` to omit the number; the label is now unconditionally `Call ${display}`.

PR #2360, opened 2026-08-24
feat(shelter-operator): shelter filters backend (SDB-218)

mikefeldberg:
SDB-218 backend

1. adds endpoint to fetch shelter operator orgs
2. splits `ShelterFilter` into public and operator filters 
3. implements filters that are available in django admin

## Summary by Sourcery

Expand shelter discovery and operator management with scoped filters, organization lookup, and richer shelter ordering.

New Features:
- Add a public shelter organization query for discovering organizations associated with shelter operators.
- Expose additional public and operator shelter filters, including search, lifecycle status, service-related attributes, districts, and ratings.
- Support shelter ordering by lifecycle status, organization name, and bed count.

Bug Fixes:
- Improve many-to-many filtering so filtered shelter results remain correctly scoped without distorting related bed-count ordering.

Enhancements:
- Separate public and operator GraphQL shelter filter types and apply operator-specific access scoping.
- Extend shelter property filtering to include funders and expand public filtering to cover accessibility, storage, programs, locations, services, and other shelter attributes.

Tests:
- Add coverage for shelter organization visibility, ordering, anonymous access, expanded public filters, operator-only filters, and shelter ordering behavior.

Chores:
- Update generated GraphQL clients and public shelter frontend queries to use the renamed public filter and query types.


vecchp (commented):
Nice piece of work. The public/operator split is the right shape, the custom `ShelterOrder` fields follow strawberry-django's documented `order_field` pattern exactly (annotate + `value.resolve()`, no direct `order_by()`), and the filter coverage in `test_shelter_filters.py` is thorough.

**Two blockers:**

1. CI is red on `test_spas_served_filter` — seed-data collision, see inline.
2. The organizations filter can't work against `operatorShelters` as it's currently scoped. That needs a product decision, not just a code change.

**One correctness issue worth fixing here:** the three new M2M filters — and the shared `make_m2m_in_filter` helper they sit next to — modify the queryset instead of returning a `Q`, which silently breaks the `NOT` operator. Both fixes are small and mechanical, and landing them together also makes `accessibility`/`storage`/`shelterPrograms` and the Bed/Room filters correct. The heavier queryset-shaped filters in that file (`properties`, `spa`, `openNow`, `mapBounds`, `geolocation`) have the same shape but need real queryset work — those will follow in a separate PR, not yours to carry here.

### `apps/betterangels-backend/common/graphql/types.py:46` — `make_m2m_in_filter`

(GitHub won't let me comment inline on a file that isn't in the diff, so it's here.)

This helper returns `(queryset.filter(...), Q())`, so the predicate never reaches the `Q` that `AND`/`OR`/`NOT` operate on. `NOT: { accessibility: [ADA_ROOMS] }` currently returns the shelters that *do* have ADA rooms.

The subquery is already built, so it's mostly a return-shape change:

```python
@strawberry_django.filter_field
def _filter(self: Any, queryset: QuerySet[Any], value: Optional[list[value_type]], prefix: str) -> Q:
    if not value:
        return Q()

    normalized_value = [value_type[v.name] if not isinstance(v, str) else v for v in value]
    lookup = f"{prefix}{related_object}__{field_name}__in"
    matching_pks = queryset.model._default_manager.filter(**{lookup: normalized_value}).values("pk")
    return Q(**{f"{prefix}pk__in": Subquery(matching_pks)})
```

Note the subquery now builds off the base manager rather than the incoming `queryset`. That matters once the result is negatable — building it from an already-narrowed queryset would make `NOT` depend on whichever sibling filters happened to run first.

Blast radius: 8 call sites, all in `shelters/types/filters.py` — the 3 you're adding to `PublicShelterFilter`, plus 5 in `CommonBedRoomFilterMixin` feeding `BedFilter` and `RoomFilter`. So this fixes bed and room filtering too.

**Two things for the description:**

- This is a **breaking schema change**. `ShelterFilter` is gone and `organizations` is no longer on the public filter. I checked consumers — the only in-repo document naming the type is `shelters.graphql` (updated here), and the Expo `Shelters.graphql` passes no filters, so nothing in the monorepo breaks. Any client outside it naming `ShelterFilter` in a variable definition fails validation on deploy.
- The `make_m2m_in_filter` change reaches `BedFilter` and `RoomFilter`, which this PR doesn't otherwise touch. Worth a line so it isn't a surprise.

**Test gaps:** no coverage for the new `funders`/`fundersIncludeNull` property filter; none for `filters` on `shelterOperatorOrganizations` despite the description listing filtering; and `ShelterOrder.organization` is only exercised through the public `shelters` query.

apps/betterangels-backend/shelters/schema.py
```diff
@@ -105,6 +108,33 @@ class Query:
         permission_classes=[IsAuthenticated],
     )
 
+    @strawberry_django.offset_paginated(
+        OffsetPaginated[OrganizationType],
+        permission_classes=[IsAuthenticated],
+        extensions=[HasOrgPerm(Shelter.perms.VIEW)],
+    )
+    def shelter_operator_organizations(
```
vecchp: **Blocking — needs a product decision.**

This endpoint returns every org with a `SHELTER_OPERATOR` group, and the frontend feeds that list into a multi-select that writes into `operatorShelters(filters: { organizations: ... })`.

But `OperatorShelterType.get_queryset` calls `shelter_queryset(...)` → `permissioned_queryset(..., organization_id=get_current_organization(info))`, which hard-filters to the single org in the `X-Organization-ID` header. So the dropdown's default path (`[selectedOrganizationId]`) works, and **every other selection returns zero rows**. The `organizations` filter at `filters.py:328` then intersects again with the user's own orgs, on top of scoping `get_queryset` already did.

The tests here document it — `test_operator_queries.py:440` carries the docstring *"Exercised via the public `shelters` query because `operatorShelters` is scoped to the active org"*, so `ShelterOrder.organization` is untested and unusable on the endpoint it was built for, for the same reason.

Two ways out: make `operatorShelters` genuinely multi-org for privileged users (and scope this endpoint to what the caller may actually query), or drop the org filter, the org ordering, and this endpoint. Worth settling before more frontend is built on it.

apps/betterangels-backend/shelters/schema.py
```diff
@@ -105,6 +108,33 @@ class Query:
...
+        self,
+        info: Info,
+        ordering: Optional[list[OrganizationOrder]] = None,
```
vecchp: These two params can go. `OrganizationType` already declares `ordering=OrganizationOrder, filters=OrganizationFilter`, and strawberry-django injects and applies both for a resolver-backed field. I ran both shapes — `ordering` and `filters` execute identically. The only difference is the generated signature:

```
current:  shelterOperatorOrganizations(ordering: [OrganizationOrder!] = null, filters: OrganizationFilter = null, pagination: ...)
standard: shelterOperatorOrganizations(filters: OrganizationFilter, ordering: [OrganizationOrder!]! = [], pagination: ...)
```

Removing them gives the shape every other paginated query in the schema has.

apps/betterangels-backend/shelters/schema.py
```diff
@@ -105,6 +108,33 @@ class Query:
...
+        filters: Optional[OrganizationFilter] = None,
+    ) -> QuerySet[Organization]:
+        """Return all organizations that have a Shelter Operator permission group."""
+        from accounts.models import PermissionGroup
```
vecchp: Query logic belongs in a selector — `docs/styleguides/python.md` keeps read queries in selectors and resolvers thin. This fits naturally next to `shelter_list` in `shelters/selectors/operator.py`, and both function-local imports disappear with the move.

While you're there: this returns every shelter-operator org regardless of the caller's membership, so any user with `view_shelter` in their active org can enumerate them all. That's plausibly what you want for the dropdown, but it should be a stated decision with a test asserting it rather than a side effect.

apps/betterangels-backend/shelters/types/filters.py
```diff
@@ -260,12 +255,188 @@ def spa(self, queryset: QuerySet, value: Optional[List[ID]], prefix: str) -> Tup
 
         return queryset.filter(spa_id__in=value).select_related("spa"), Q()
 
+    @strawberry_django.filter_field
+    def on_site_security(self, info: Info, value: Optional[bool], prefix: str) -> Q:
```
vecchp, line 243: Nit: `if value is None` is unreachable. `process_filters` skips both `UNSET` and explicit `null` unless the field opts into `filter_none`; only `false` reaches the method. (The `if not value` guards on the list filters *do* matter — they catch `[]`.)

apps/betterangels-backend/shelters/types/filters.py
```diff
@@ -260,12 +255,188 @@ def spa(self, queryset: QuerySet, value: Optional[List[ID]], prefix: str) -> Tup
...
+        return Q(**{f"{prefix}on_site_security": value})
+
+    @strawberry_django.filter_field
+    def city(self, info: Info, value: Optional[List[ID]], prefix: str) -> Q:
+        if not value:
+            return Q()
+        return Q(**{f"{prefix}city__in": value})
+
+    @strawberry_django.filter_field
+    def cities_served(
```
vecchp: `cities_served`, `spas_served` and `services` filter the queryset directly, which breaks the `NOT` operator the filter input advertises. Measured against 20 shelters, 9 of which serve the target city:

```
plain: totalCount=9
NOT:   totalCount=9     ← should be 11
```

The method narrows the queryset unconditionally; `NOT` only negates the empty `Q` it returns. Strawberry-django's filter docs call this out: *"It is discouraged to use `queryset.filter()` directly. When using more complex filtering via `NOT`, `OR` & `AND` this might lead to undesired behaviour."*

An `Exists` predicate fixes it and drops the `.distinct()` — no JOIN, so no duplicate rows to dedupe:

```python
@strawberry_django.filter_field
def cities_served(self, info: Info, value: Optional[List[ID]], prefix: str) -> Q:
    if not value:
        return Q()
    through = models.Shelter.cities_served.through
    return Q(Exists(through.objects.filter(shelter_id=OuterRef("pk"), city_id__in=value)))
```

Also worth knowing `.distinct()` is a query-wide flag no filter can scope or undo, and the input already exposes `DISTINCT: Boolean` so the client can ask for it.

**Scope:** these three plus the `make_m2m_in_filter` fix (see the summary comment) are small and mechanical — please take them in this PR. `properties`, `spa`, `openNow`, `mapBounds` and `geolocation` have the same shape but need real queryset work, and some may not be expressible as a pure `Q` at all. Those come as a fast-follow, not your problem here.

apps/betterangels-backend/shelters/types/filters.py
```diff
@@ -260,12 +255,188 @@ def spa(self, queryset: QuerySet, value: Optional[List[ID]], prefix: str) -> Tup
...
+        Uses an isolated correlated subquery rather than a JOIN-based
+        ``Count('beds', distinct=True)``.  The subquery is unaffected by any
+        M2M JOINs that active filters (cities_served, services, spasServed)
+        may have added to the outer queryset, so no DISTINCT deduplication is
+        needed and the count is always correct in a single pass.
+        Uses a distinct annotation name (``_order_bed_total``) to avoid
+        conflicting with the ``_bed_total`` annotation added by the bedCounts
+        resolver when both are requested in the same query.
+        """
+        bed_count_subq = (
```
vecchp: This subquery is `shelter_count_subquery(models.Bed)` from `shelters/selectors/computed_status.py:157` — the same helper `bedCounts` uses for `_bed_total`.

```python
Coalesce(shelter_count_subquery(models.Bed), Value(0))
```

The `Coalesce` is a good call — keeping it means the sort agrees with `bedCounts.total` rather than relying on null placement. The docstring explaining why a JOIN-based `Count` would be wrong can go once the shared helper is doing the work.

apps/betterangels-backend/shelters/tests/test_operator_queries.py
```diff
@@ -407,3 +587,86 @@ def test_combined_properties_filter(self) -> None:
...
+        organization_recipe.make(name="Zebra", preset_names=["shelter"], owner_roles=(SHELTER_OPERATOR,))
+        organization_recipe.make(name="Alpha", preset_names=["shelter"], owner_roles=(SHELTER_OPERATOR,))
+        organization_recipe.make(name="Middle", preset_names=["shelter"], owner_roles=(SHELTER_OPERATOR,))
+
+        self.graphql_client.force_login(self.org_1_case_manager_1)
+        response = self.execute_graphql(self.QUERY)
+
+        self.assertIsNone(response.get("errors"))
+        names = [r["name"] for r in response["data"]["shelterOperatorOrganizations"]["results"]]
+        self.assertEqual(names, ["Alpha", "Middle", "test_org", "Zebra"])
```
vecchp: This test can't fail. `Organization._meta.ordering == ["name"]`, so results come back name-ASC whether or not the `ordering` argument is applied. Hardcoding `"test_org"` from the base fixture is also what forced the `fix test` commit — Python's `sorted()` puts `Zebra` before `test_org`, Postgres's collation doesn't.

Assert a non-default ordering (`name: DESC`) over only the orgs the test creates, and it starts earning its keep.

apps/betterangels-backend/shelters/tests/test_shelter_filters.py
```diff
@@ -2026,3 +1809,193 @@ def test_shelter_has_available_beds_filter_null(self) -> None:
...
+        shelter_recipe.make(status=StatusChoices.APPROVED, cities_served=[city_b])
+
+        query = self.get_shelters_query("id")
+        response = self.execute_graphql(query, variables={"filters": {"citiesServed": [str(city_a.id)]}})
+        result_ids = {r["id"] for r in response["data"]["shelters"]["results"]}
+        self.assertEqual(result_ids, {str(match.id)})
+
+    def test_spas_served_filter(self) -> None:
+        spa_a, _ = SPA.objects.get_or_create(short_name="1", defaults={"long_name": "1 - Antelope Valley"})
+        spa_b, _ = SPA.objects.get_or_create(short_name="2", defaults={"long_name": "2 - San Fernando Valley"})
```
vecchp: **Blocking — this is the CI failure.**

```
django.db.utils.IntegrityError: duplicate key value violates unique constraint "shelters_spa_short_name_key"
```

`shelters/apps.py` seeds the lookup tables on `post_migrate`, so SPA 2 already exists in the test DB as `"2 - San Fernando"` (`shelters/seed.py:109`). Passing `long_name="2 - San Fernando Valley"` makes `get_or_create` miss on that field, attempt an INSERT, and collide on the unique `short_name`.

`spa_a` only survives because `"1 - Antelope Valley"` matches the seed verbatim.

```python
spa_b, _ = SPA.objects.get_or_create(short_name="2", defaults={"long_name": "2 - San Fernando Valley"})
```

— or read the seeded rows directly, since they're guaranteed present.

apps/betterangels-backend/shelters/schema.py
mikefeldberg: made this a public endpoint. The information isn't privileged or sensitive, so it doesn't need to be gated

apps/betterangels-backend/shelters/tests/test_shelter_filters.py
mikefeldberg: changed `get_or_create` to `get`

apps/betterangels-backend/shelters/schema.py
vecchp: Public is settled, this isn't about gating. The separate issue is still open: `organizations` can't return anything the header didn't already allow.

`OperatorShelterType.get_queryset` ANDs in `organization_id = <X-Organization-ID>` before filters run, so `organizations` can only narrow to the active org. A user in both orgs with `view_shelter` in both still gets 0 rows for `[org_2]`.

Demo — save as `shelters/tests/test_org_filter_demo.py`, run `pytest shelters/tests/test_org_filter_demo.py -s`, delete after. It names no filter input type, so it runs unchanged on `main` and here, which is the point of the second test.

```python
"""Throwaway demo — delete after reading. Prints, does not assert."""

from accounts.role_manager import OrgRoleManager
from common.tests.utils import GraphQLBaseTestCase
from django.contrib.auth.models import Permission
from notes.groups import CASEWORKER

from shelters.enums import StatusChoices
from shelters.models import Shelter
from shelters.tests.baker_recipes import shelter_recipe

OPERATOR_QUERY = """
    query ($orgIds: [ID!]) {
        operatorShelters(filters: { organizations: $orgIds }) {
            totalCount
            results { name organization { name } }
        }
    }
"""

PUBLIC_QUERY = """
    query ($orgIds: [ID!]) {
        shelters(filters: { organizations: $orgIds }) { totalCount }
    }
"""


class OrgFilterDemoTestCase(GraphQLBaseTestCase):
    def setUp(self) -> None:
        super().setUp()
        user = self.org_1_case_manager_1

        # One user in BOTH orgs with view_shelter in BOTH, so an empty result
        # can't be blamed on membership or permissions.
        self.org_2.add_user(user)
        OrgRoleManager(self.org_2).add_roles(user, CASEWORKER)

        app_label, codename = Shelter.perms.VIEW.split(".")
        perm = Permission.objects.get(codename=codename, content_type__app_label=app_label)
        for org in (self.org_1, self.org_2):
            org.permission_groups.get(template__name=CASEWORKER.name).group.permissions.add(perm)

        shelter_recipe.make(
            organization=self.org_1, name="Org 1 Shelter", status=StatusChoices.APPROVED, is_private=False
        )
        shelter_recipe.make(
            organization=self.org_2, name="Org 2 Shelter", status=StatusChoices.APPROVED, is_private=False
        )

        self._set_active_org(self.org_1)
        self.graphql_client.force_login(user)

    def test_operator_shelters_org_filter(self) -> None:
        print("\nX-Organization-ID = org_1")
        print("user is a member of org_1 AND org_2, with view_shelter in both\n")
        print(f"  {'filters.organizations':<28} {'totalCount':>10}   returned")

        cases = [
            ("(omitted)", None),
            ("[org_1]", [str(self.org_1.id)]),
            ("[org_2]", [str(self.org_2.id)]),
            ("[org_1, org_2]", [str(self.org_1.id), str(self.org_2.id)]),
        ]
        for label, org_ids in cases:
            payload = self.execute_graphql(OPERATOR_QUERY, variables={"orgIds": org_ids})
            data = payload["data"]["operatorShelters"]
            names = ", ".join(r["name"] for r in data["results"]) or "-"
            print(f"  {label:<28} {data['totalCount']:>10}   {names}")

    def test_public_shelters_org_filter(self) -> None:
        payload = self.execute_graphql(PUBLIC_QUERY, variables={"orgIds": [str(self.org_1.id)]})
        print("\npublic `shelters(filters: { organizations: ... })`")
        print(f"  errors: {payload.get('errors')}")
        print(f"  data:   {payload.get('data')}")
```

Not introduced here — `get_queryset` is byte-identical on `main`. What's new is #2361 building on it: `orgOptions` (`ShelterFilterPanel.tsx:99`) feeds the `organizations` control at `:158`, which feeds `Dashboard.tsx:185` → `operatorShelters`. Every option except the active org renders empty.

The second test also shows `organizations` is gone from the public filter, so `shelters` can no longer be filtered by org at all — including anonymous callers, who could before. Probably unintended.

Suggest putting `organizations` back on `PublicShelterFilter` as plain `Q(organization__in=value)` and letting `OperatorShelterFilter` inherit it. That fixes the public query. It doesn't fix `operatorShelters` — that needs its own call: drop the filter and switch orgs via the NavBar picker, or make `operatorShelters` multi-org.

apps/betterangels-backend/shelters/types/filters.py
```diff
@@ -260,12 +239,172 @@ def spa(self, queryset: QuerySet, value: Optional[List[ID]], prefix: str) -> Tup
...
+
+        return Q(Exists(through.objects.filter(shelter_id=OuterRef("pk"), spa_id__in=value)))
+
+    @strawberry_django.filter_field
+    def services(
+        self, queryset: QuerySet, value: Optional[List[ID]], prefix: str
+    ) -> Tuple[QuerySet[models.Shelter], Q]:
+        if not value:
+            return queryset, Q()
+        return queryset.filter(**{f"{prefix}services__in": value}).distinct(), Q()
```
vecchp, line 275: `citiesServed` and `spasServed` moved to `Q(Exists(...))`; this one still narrows the queryset, so `NOT` stays broken — `process_filters` keeps queryset mutations made inside the `NOT` branch and negates only the empty `Q`.

```python
through = models.Shelter.services.through
return Q(Exists(through.objects.filter(shelter_id=OuterRef("pk"), service_id__in=value)))
```

`.distinct()` goes with it.

apps/betterangels-backend/shelters/selectors/operator.py
```diff
@@ -38,6 +41,22 @@ def shelter_list(
...
+    if queryset is None:
+        queryset = Organization.objects.all()
+
+    has_shelter_operator_group = Exists(
+        PermissionGroup.objects.filter(
+            organization=OuterRef("pk"),
+            template__name=SHELTER_OPERATOR.name,
+        )
+    )
+    return queryset.filter(has_shelter_operator_group)
```
vecchp, line 57: Correct today — reconcile derives groups from `profile.org_types`, so outreach-only orgs never get a Shelter Operator group.

But this keys off the derived artifact rather than the declared one. `OrganizationProfile.org_types` is the source of truth and is GIN-indexed:

```python
return queryset.filter(profile__org_types__contains=[OrgTypeChoices.SHELTER])
```

One query instead of an `Exists` subquery, and immune to drift: `reconcile_org_groups` skips orgs with no profile entirely (`if org_types is not None`), so a profile-less org holding a Shelter Operator row appears here but wouldn't there. I haven't checked whether prod has any — happy to send a read-only query.

Either way, dual-type orgs (`["shelter", "outreach"]`, settable in the Django admin) match under both approaches. Probably right, but worth stating.

vecchp:
Re-reviewed at `28b8a792`. Most of the last round is resolved: `make_m2m_in_filter` returns a `Q` (fixing `NOT` for the 3 new filters and 5 Bed/Room ones), SPA collision fixed across three test files, resolver params dropped, query moved to a selector, public/anonymous access tested, dead `value is None` guards gone, `shelter_count_subquery` reused, ordering test asserts a non-default order. Thanks.

Correction to myself: the function-local `from shelters.groups import SHELTER_OPERATOR` **is** required — `shelters/groups.py` imports `shelters.models` at module level. Ignore that note.

Still open: the `organizations` scoping question and the public query losing `organizations` (threaded reply), and `services` (inline).

For the description:
- Breaking schema change — `ShelterFilter` is gone, `organizations` off the public filter. Nothing in-repo breaks, but external clients naming `ShelterFilter` fail validation on deploy.
- `make_m2m_in_filter` also changes `BedFilter` and `RoomFilter`.

Minor: `test_operator_shelters_returns_all_accessible_orgs_when_no_filter` doesn't test what its name says (passes only because that user has one org); no `funders` coverage; `filters` on `shelterOrganizations` declared but never exercised; `SHELTER_OPERATOR` imported function-locally in 5 tests; `bed_count` docstring repeats the comment below it.

vecchp (approved):

PR #2361, opened 2026-08-24
feat(shelter-operator): shelter filter frontend (SDB-218)

mikefeldberg:
SDB-218

adds sort & filter drawer with all filters available in django admin

## Summary by Sourcery

Enable comprehensive shelter operator search, sorting, filtering, pagination, and organization-aware shelter management.

New Features:
- Add a shelter operator sort and filter drawer covering shelter properties, operational attributes, organizations, locations, services, and boolean criteria.
- Support server-side shelter search, filtering, sorting, pagination, organization display, and shelter-switch confirmation in the dashboard.
- Allow global operators to select an organization when creating a shelter while retaining active-organization behavior for scoped operators.

Bug Fixes:
- Prevent client-side permission checks from granting access based on unknown backend permission strings.
- Keep global and organization-scoped permissions separate when evaluating user access.

Enhancements:
- Add reusable pagination controls and controlled table sorting support.
- Warm filter option data from organizations, cities, SPAs, and service categories for the shelter filter drawer.
- Improve shelter status labels and expand shelter list data with organization information.

Tests:
- Add coverage for global permission evaluation, permission filtering, and organization-aware shelter creation behavior.


libs/react/shelter-operator/src/lib/atoms/shelterFiltersAtom.ts
```diff
@@ -1,15 +1,61 @@
 import { atomWithReset } from 'jotai/utils';
 
 export type TOperatorShelterFilters = {
+  // ShelterPropertyInput fields
```
tglaz, line 4: main suggestion would be to have this be strongly typed. 
we have to convert them anyway to send filters with the query, so we can just store them as our gql types in the core `operatorShelterFiltersAtom`. 
Right now we just cast them, which kind of works since the options in the forms have integrity, but that's a lot of casting.

- if we need them as strings for the forms to use, we can just define a separate derived `getter` atom, or perhaps using `useAtomValue` would work
- this is maybe out of scope, and there are many ways to architect this with atoms. But we can also define a separate utils such as setFilterValue, toggleFilterValue, clearFilter etc... so we don't need to do this everywhere: 
```() => setFilters((prev) => ({ ...prev, [filterKey]: [] }))```

but the typing would be the primary concern. whether we want to do it now is a separate question

File: libs/react/shelter-operator/src/lib/pages/dashboard/Dashboard.tsx [hunk 2]
```diff
@@ -152,26 +305,25 @@ export function Dashboard() {
         className="my-1 flex w-full flex-wrap items-center gap-3 bg-white px-3"
         style={{ fontFamily: 'Poppins, sans-serif' }}
       >
-        <label className="flex h-11 w-full max-w-[380px] items-center gap-2 rounded-full border border-[#D3D9E3] bg-white px-2">
-          <span className="flex h-8 w-9 items-center justify-center rounded-full bg-[#FCF500] text-[#1E3342]">
-            <Search size={20} />
+        <div className="flex w-full max-w-[380px] flex-col gap-1">
+          <label className="flex h-11 w-full items-center gap-2 rounded-full border border-[#D3D9E3] bg-white px-2 mb-2">
+            <span className="flex h-8 w-9 items-center justify-center rounded-full bg-[#FCF500] text-[#1E3342]">
+              <Search size={20} />
```
tglaz, line 311: I wonder if we'd want to keep `search` and `filter` functionality inside the Dashboard or move it out to a separate view. Shelter operators will typically have 1-2 shelters, and at max ~10? So they probably won't really need these, and moving them out would keep the Dashboard simpler

The `search` and `filter` features seem more for BA admins, or users that have special permissions for this. I'd imagine we can just have some `admin` button show up at bottom the SideBar if the user has such permissions. And that would open a view to search/select cross-org shetlers.

But that's a separate UX questions probably

tglaz (approved):

vecchp:
#2360 renames `shelterOperatorOrganizations` → `shelterOrganizations`. `useShelterOperatorOrganizations.graphql` and its generated types still use the old name, so this needs a codegen pass once that lands.

Also: `operatorShelters` is scoped to the `X-Organization-ID` org, so the multi-select at `ShelterFilterPanel.tsx:158` → `Dashboard.tsx:185` renders empty for every option except the active org.

PR #2372, opened 2026-08-24
feat(accounts): transfer organization ownership, and confirm before revoking roles (DEV-2552)

vecchp:
## What

The two follow-ups #2352 and #2354 declared as "Known, not addressed", now that both have merged.

## 1. Ownership can be handed to another member

`Organization.add_user` makes the first member the owner, and `organization_remove_member` refuses to remove an owner. Together those meant **the first person invited to a new organization could never be removed from it** — the refusal said "transfer ownership first" while offering no way to do so except editing the `OrganizationOwner` row by hand.

`organization_transfer_ownership` is that way, and the organization page offers it beside **Add member**. Ownership is one row, so the previous owner simply stops owning: they keep their membership and their roles, and their row becomes deletable. Only members are offered, because the service refuses anyone else — the choices and the rule agree rather than the form promising something the service rejects.

The move delegates to `Organization.change_owner`, so django-organizations' `owner_changed` signal still fires.

## 2. A save that revokes roles asks first

Two edits on the organization page take a role away from everyone holding it, and neither of them shows that on the form:

- **unchecking an org type** — reconcile deletes its permission groups, and `delete_orphaned_group` takes the `auth.Group` with them.
- **ticking Delete on a Permission groups row** — the same teardown, except `save_related`'s reconcile then recreates a derived row on the same save with a fresh, empty group. The page comes back looking untouched while everyone who held the role has lost it. The quieter of the two, and the reason `can_delete = False` on the inline was not the fix: the same delete stays available on `PermissionGroupAdmin`, whose confirmation page lists neither the group nor its members, because that teardown is a `post_delete` receiver rather than an FK cascade.

The save is now re-offered first, naming each role and how many members hold it, in the shape `delete_selected` already uses: the submitted POST is preserved in hidden fields and replayed with a confirmation flag.

A change **nobody holds** goes straight through. A prompt that always fires stops being read, and the point of this one is that it means something. Neither is a role the remaining org types still grant — `Organization Admin` and `Organization Superuser` belong to both types, so dropping one of them revokes nothing.

Intercepted in `change_view` rather than the form — a form cannot re-render the whole change view, and a `clean()` error would be the wrong shape, since this is a confirmation rather than a rejection. It defers to `has_change_permission`, so the prompt never answers a question Django is about to refuse.

## Testing

- `1486 passed, 31 skipped` — full suite. `mypy` clean on 387 files with the cache cleared; `ruff check` and `ruff format` clean.
- Every behaviour mutation-tested: never intercepting, always intercepting, intercepting only one of the two routes, prompting when nobody holds the role, naming a role the remaining type keeps, skipping the permission check, transferring to a non-member, transferring in an organization with no owner row, and a transfer that does nothing.
- The un-removable-owner case is pinned end to end: removal raises, transfer, removal succeeds.
- The no-owner transfer is the branch that matters in production: the Shelter CSV importer's `get_or_create` never called `add_user`, so 97 of 99 organizations have no `OrganizationOwner` row at all.

Two things caught by mutation-testing rather than by review. The first version of the confirmation ran `super().change_view()` before deciding whether to ask, so it saved first and asked afterwards; the check now precedes the save. And the no-owner transfer test survived a mutant that deleted the branch entirely — deleting the `OrganizationOwner` row left it cached on the in-memory organization, so `change_owner` found it anyway. The test refetches.

## Not included

The confirmation says that putting a role back creates it again but does not give it back to anyone. Restoring the previous holders is a bigger question and is not attempted here.

`PermissionGroupAdmin`'s own delete page has the same hole with a weaker warning, and `add_member_view` / `change_member_roles_view` do not check `has_change_permission` either. Both are pre-existing and belong in their own PR.

## Summary by Sourcery

Enable organization ownership transfers and clearly confirm saves that revoke member roles.

New Features:
- Add an admin workflow for transferring organization ownership to another member, including organizations without an existing owner record.

Bug Fixes:
- Allow former organization owners to be removed after ownership is transferred.
- Require confirmation before organization changes revoke roles held by members, while allowing unaffected saves to proceed directly.

Enhancements:
- Report the affected roles and member counts while preserving the original organization save for confirmation and replay.

Tests:
- Add coverage for ownership transfer, owner removal, ownerless organizations, invalid non-member transfers, and role-loss confirmation paths.


vecchp:
Reconciliation note from #2378, which makes `PermissionGroup` a subclass of `auth.Group`.

Two things here need adjusting whichever way the two land:

1. `_roles_lost_to_org_type_removal`'s sibling for inline deletions reads the row's primary key as `permission_groups-N-id`. Under inheritance the inline's pk field is `group_ptr`, so that lookup finds nothing and the confirmation silently stops firing for row deletions. Reading `PermissionGroup._meta.pk.name` instead survives both shapes.
2. `OrganizationTypeRemovalConfirmationTestCase._payload` posts `permission_groups-N-id` and `permission_groups-N-name`. Those become `-group_ptr` and `-label`. Worth noting that with (1) unfixed the test still passes — it supplies the key the code looks for — so the pair has to be fixed together or the coverage is illusory.

Nothing to do until one of them merges.

mikefeldberg (approved):

vecchp:
Updating the earlier reconciliation note: the #2378 overlap is larger than "read the pk as `-id`", found by merging #2378 into this branch locally and running `accounts`.

## Done here, ahead of the merge (`89046c26`)

`_roles_lost_to_row_deletion` and the test's `_payload` now derive the inline's pk field from `PermissionGroup._meta.pk.name` — `id` today, `group_ptr` after #2378. This is the half that **fails silently**: a hardcoded `-id` matches nothing after #2378, so the page reports no losses and the confirmation never fires, with its tests still passing. Unobservable on this branch, which is exactly why it goes in first.

## Still to do, once #2378 is in

The rest fails loudly, so it is safe to leave until the merge:

- `_held_by` — `permission_group.name` → `.label`, `permission_group.group.user_set.count()` → `permission_group.user_set.count()`, and `select_related("group", "template")` → `select_related("template")`.
- `_payload` — the inline's label field is `label`, so the POST key becomes `permission_groups-{i}-label` carrying `row.label`.
- `_payload` also posts `"is_active": "on"`, which #2377 removed from the form. This branch predates #2377, so it is still valid here; drop it in the same pass.

With only the first two of those ported, the role-loss POSTs come back as **"Please correct the errors below"** rather than the confirmation page — three `OrganizationRoleLossConfirmationTestCase` tests fail. So the merge needs the form payload and `_held_by` done together, not one at a time, and the confirmation should be mutation-tested afterwards: a lookup that silently matches nothing would otherwise leave those tests green.

Order that avoids all of this being guesswork: #2378 → #2380 → merge `main` here → fix the three items above → this.

PR #2374, opened 2026-08-24
Create custom Header to use with modal-screen (DEV-2526 DEV-2487 DEV-2524)

tglaz:
### Create custom Header to use with modal-screen
DEV-2526

- adds custom `ScreenHeader` that we can render in any screen
  - initial target is `modal-screen`
     - this will allow us to to render things like `BottomSheet` in full-screen (assuming the screen is full screen height)
     - also an escape hatch for any iOS glass issues - can just switch to this
- does not implement it in any screen yet
  - but does update some `config` logic as to nav background colors etc...

## Summary by Sourcery

Introduce a unified configurable header system for modal and screen navigation surfaces.

New Features:
- Add configurable in-app screen headers with screen, modal, and minimal variants, custom slots, safe-area handling, and accessible back and close controls.
- Allow modal screens to select native, custom, or no headers and customize modal close labels and controls.

Bug Fixes:
- Prevent incorrect top spacing for custom headers in page-sheet modal presentations.

Enhancements:
- Centralize header styling and native-header options so custom and native navigation headers share consistent configuration.
- Replace modal header hiding with the unified header configuration API and support text-labelled native close buttons.

Tests:
- Add coverage for ScreenHeader titles, default and custom controls, safe-area insets, variants, and navigation behavior.

Chores:
- Disable the React hooks refs lint warning and extend shared close-button support with test identifiers and child content.

## Summary by Sourcery

Introduce a configurable shared header system for native and in-app navigation surfaces, including custom modal headers.

New Features:
- Add a reusable, safe-area-aware ScreenHeader with configurable palettes, variants, slots, and accessible back/close controls.
- Allow modal screens to choose native, custom, or no headers and configure modal close labels and custom header controls.

Bug Fixes:
- Prevent excess top spacing when custom headers are used with iOS page-sheet presentations.

Enhancements:
- Unify native and custom header styling and navigation options through shared configuration and palettes.
- Standardize stack navigation controls and support configurable pressed states and test identifiers for shared buttons.

Tests:
- Add coverage for ScreenHeader rendering, variants, safe-area behavior, controls, labels, and navigation actions.

Chores:
- Replace the legacy modal header-hiding configuration with the unified header API.


eslint.config.mjs
```diff
@@ -143,7 +143,7 @@ export default [
     rules: {
       ...reactHooksPlugin.configs.recommended.rules,
       'react-hooks/set-state-in-effect': 'warn',
-      'react-hooks/refs': 'warn',
+      'react-hooks/refs': 'off',
```
vecchp: Why are we turning this off repo wide? Seem unrelated to this PR and it's needs?

libs/expo/betterangels/src/lib/navigation/modalOptions/getDefaultStackModalOptions.tsx
```diff
@@ -1,48 +1,52 @@
...
 type TProps = {
   title?: string;
   presentation?: TModalPresentationType;
-  hideHeader?: boolean;
   onClose?: null | (() => void);
+  /** Header configuration, as passed to `showModalScreen`. */
+  header?: TModalHeaderConfig;
 };
 
 export function getDefaultStackModalOptions(props?: TProps) {
```
vecchp: Doesn't this impact live code?  So no users pass the new header prop so i think this changes the colors?

main uses `BRAND_STEEL_BLUE` and the default falls through to `BRAND_DARK_BLUE`? Is this correct?

apps/betterangels/src/app/modal-screen.tsx
```diff
@@ -22,6 +24,23 @@ export default function BaseModalScreen() {
...
           }}
         >
+          {header?.mode === 'custom' && (
+            <ScreenHeader
+              variant={header.variant ?? 'modal'}
+              title={title}
+              // A page-sheet 'modal' starts below the notch already, so the
+              // window's top inset would add a wrong gap there. Every other
+              // presentation fills the window.
+              topInset={presentation === 'modal' ? 0 : undefined}
```
vecchp: Don't we have iOS specific reasoning applied to both platforms here?

I think when set to `modal` ios uses `UIModalPresentationStyleAutomatic` where as on android it's effectively equivalent to `push` full window (can check react-native-screens `types.d.ts:465`. 

With edge-to-edge mandatory on SDK 57, `insets.top > 0` , so on Android the bar gets `paddingTop: 0` and the title and close button draw under the status bar. 'modal' is also `DEFAULT_PRESENTATION`, so it's the path you get by not specifying one. Same root cause makes variant: 'minimal' a zero-height header, since its height is the inset.

We could do 
```
Platform.OS === 'ios' && presentation === 'modal' ? 0 : undefined
```
 
or move this decision into `ScreenHeader` instead of here?

libs/expo/betterangels/src/lib/navigation/ScreenHeader/ScreenHeader.tsx
```diff
@@ -0,0 +1,154 @@
...
+        {
+          backgroundColor: bgColor,
+          paddingTop,
+          height: paddingTop + HEADER_BAR_HEIGHT,
+        },
+      ]}
+    >
+      <View style={[styles.slot, styles.slotLeft]}>{left}</View>
+
+      {!!title && (
```
vecchp: With no title, the right-hand button isn't right-aligned? I think?

libs/expo/betterangels/src/lib/providers/modalScreen/types.ts
```diff
@@ -8,24 +9,50 @@ export type TRenderContentApi = {
...
+   * Which header this modal gets. Defaults to `'native'` — the platform bar.
+   * `'custom'` draws the in-app `ScreenHeader` instead (styled by `variant`);
+   * `'none'` draws no bar at all.
+   */
+  mode?: THeaderMode;
+  /** Which style from `headerVariants`. Only used when `mode` is `'custom'`. */
+  variant?: THeaderVariant;
+  /** Left slot. Same tri-state as `ScreenHeader`'s `buttonLeft`: omit for the
+   *  variant's default, `null` for no button, a node for custom content. */
+  buttonLeft?: ReactNode | null;
```
vecchp: No button right? Is that intentional?

eslint.config.mjs
tglaz: yeah, I forgot to undo it. i'll reset

libs/expo/betterangels/src/lib/navigation/modalOptions/getDefaultStackModalOptions.tsx
tglaz: can you point me to where you see it?
I think the default is still `BRAND_DARK_BLUE` here (as per main).
We typically use `BRAND_STEEL_BLUE` for modal-screen headers though.
it's driven by `headerVariants` defined in the config.ts

the default variant is `screen`, as set here:
```
export function ScreenHeader(props: TProps) {
  const {
    variant = 'screen',
    title,
```

and backgroundColor:
```
export const headerVariants: Record<THeaderVariant, THeaderVariantConfig> = {
  screen: {
    backgroundColor: Colors.BRAND_DARK_BLUE,
    textColor: Colors.WHITE,
```

libs/expo/betterangels/src/lib/providers/modalScreen/types.ts
tglaz: yeah. right now it has a `closeButton` that occupies the right slot

apps/betterangels/src/app/modal-screen.tsx
tglaz: think better to keep it here for now. 
the `presentation` prop doesn't get passed and probably shouldn't (not yet anyway)

libs/expo/betterangels/src/lib/navigation/modalOptions/getDefaultStackModalOptions.tsx
vecchp: The non-`card` branch here uses `getNativeHeaderOptions('screen')`, so `presentation: 'modal'` and `'fullScreenModal'` now resolve to `BRAND_DARK_BLUE`. Main built those two from `defaultModalNavOpts.modal` / `.fullScreenModal`, which were `BRAND_STEEL_BLUE` — the steel-blue-for-modal-headers convention. 22 of the 23 `showModalScreen` call sites use one of those two presentations, and none of them pass `header`, so they all land on this branch.

`getNativeHeaderOptions('modal')` restores it — `headerVariants.modal` already carries the steel blue.

Drop this at `libs/expo/betterangels/src/lib/navigation/modalOptions/getDefaultStackModalOptions.spec.tsx` to see it. It passes on `main` and fails here:

```tsx
import { Colors } from '@monorepo/expo/shared/static';
import { TModalPresentationType } from '../../providers';
import { getDefaultStackModalOptions } from './getDefaultStackModalOptions';

// Same reason as ScreenHeader.spec.tsx: the shared barrels pull in native
// modules that cannot initialise here. Nothing below renders a button.
vi.mock('@monorepo/expo/shared/ui-components', () => ({
  CloseButton: () => null,
  TextBold: () => null,
  TextButton: () => null,
  TextRegular: () => null,
}));

vi.mock('@monorepo/expo/shared/icons', () => ({ PlusIcon: () => null }));

vi.mock('expo-router', () => ({ useRouter: () => ({ back: () => undefined }) }));

function headerBackgroundFor(presentation: TModalPresentationType) {
  const options = getDefaultStackModalOptions({
    presentation,
    title: 'Upload Files',
    onClose: () => undefined,
  }) as { headerStyle?: { backgroundColor?: Colors } };

  return options.headerStyle?.backgroundColor;
}

describe('getDefaultStackModalOptions native header', () => {
  it('gives both modal presentations the steel blue bar', () => {
    expect(headerBackgroundFor('modal')).toBe(Colors.BRAND_STEEL_BLUE);
    expect(headerBackgroundFor('fullScreenModal')).toBe(
      Colors.BRAND_STEEL_BLUE,
    );
  });

  it('gives a pushed card the dark blue bar', () => {
    expect(headerBackgroundFor('card')).toBe(Colors.BRAND_DARK_BLUE);
  });
});
```

```
cd libs/expo/betterangels
npx vitest run src/lib/navigation/modalOptions/getDefaultStackModalOptions.spec.tsx
```

On this branch:

```
FAIL  getDefaultStackModalOptions native header > gives both modal presentations the steel blue bar
AssertionError: expected '#1E3342' to be '#375C76'
```

Worth keeping once it's fixed — nothing currently pins the modal header colour.

vecchp (approved):

tglaz:
`run-e2e` passed on CI

shiozakiyuko (commented):
Hi @tglaz 

So far, on iOS looks good. However, the other modal screens are using X to close the modal. This one is using "Done". Shall we be consistent and use X here too? 
As for Android, I'm still waiting for loading as usual.

shiozakiyuko (commented):
I was finally able to test Android side and it looks good as well. Thanks.

PR #2377, opened 2026-08-25
fix(accounts): warn before a permission group delete revokes roles, and drop the inert Active switch (DEV-2552)

vecchp:
## What

Two things the organization admin promises and does not do. Both fall out of reviewing #2372; neither is caused by it.

## 1. A permission group delete says who loses the role

Deleting a `PermissionGroup` from its own admin page revokes that role from every member holding it, and the confirmation page never says so.

Django cannot work it out: `delete_orphaned_group` tears the `auth.Group` down from `post_delete` rather than by cascade, and the foreign key points from the row *to* the group — so nothing appears downstream of what is being deleted, least of all the people about to lose their access.

`get_deleted_objects` is the hook for exactly this. `ModelAdmin.delete_view` and the `delete_selected` action both call it, so one override covers a single delete and a bulk one alike, with no custom template. Each row now contributes a line naming its group and how many members lose the role.

This is the sibling of the confirmation #2372 adds on the organization page — and this is the page that fix sends you to when you actually want a row gone.

## 2. `is_active` comes off the organization form

The page has an **Active** checkbox. Unchecking it changes nothing: members keep signing in, keep their org-scoped permissions, and keep seeing everything.

`is_active` is django-organizations', and it ships an `ActiveOrgManager` that filters on it. This codebase never uses that manager — every query goes through the unfiltered default, `OrganizationType` exposes only `id` and `name`, and `OrganizationFilter` has no `is_active`. The column has never gated anything here.

Taking it off the form rather than wiring it up. What deactivation should mean is a real question with a real blast radius — header org resolution, the org-scoped permission checks, the shelter operator selectors, everyone in the org at once — and it is better answered when someone needs it than guessed at now. The column stays, so nothing is lost and nothing is migrated.

## Testing

- `1474 passed, 31 skipped` — full suite. `ruff check`, `ruff format --check` and `mypy` clean on 387 files with the cache cleared.
- Mutation-tested: removing the `get_deleted_objects` override fails all three of its tests, putting `is_active` back on the form fails its test, and always pluralising fails the one-member case.
- Both delete routes are covered, because `delete_selected` is the likelier one.

## Not included

`add_member_view`, `change_member_roles_view` and `transfer_ownership_view` reach `has_change_permission` through nothing but `admin_site.admin_view`, which only asks for `is_staff`. Every staff user in this codebase is a superuser today — the only `is_staff=True` grants are the local-dev admin and a seed command — so it is a gap rather than a live hole, and it belongs in its own PR covering all three.

## Note for whoever merges second

#2372 touches `accounts/tests/test_admin.py` as well, and its `_payload` helper still posts `is_active`. Harmless either way — an extra POST key for a field the form does not render — but worth dropping when the two meet.

## Summary by Sourcery

Make permission-group deletion consequences explicit and remove the organization form control that does not affect application behavior.

Bug Fixes:
- Warn administrators which permission group and how many members will lose the role when deleting permission groups through either single or bulk deletion.
- Remove the non-functional Active field from the organization administration form.

Tests:
- Add coverage for single-delete and bulk-delete warnings, including correct singular wording and removal of the inactive organization field.

### DIFF
--- a/apps/betterangels-backend/accounts/admin.py
+++ b/apps/betterangels-backend/accounts/admin.py
@@ -77,6 +77,40 @@ class PermissionGroupAdmin(admin.ModelAdmin):
     # group's members with it.
     fields = ("organization", "template", "name")
 
+    def get_deleted_objects(
+        self, objs: Any, request: HttpRequest
+    ) -> tuple[list[Any], dict[str, int], set[str], list[str]]:
+        """Name the ``auth.Group`` going with each row, and who loses the role.
+
+        Django cannot work this out on its own: the group is torn down by
+        :func:`accounts.signals.delete_orphaned_group` from ``post_delete`` rather
+        than by cascade, and the foreign key points the other way, so nothing
+        appears downstream of the row being deleted — least of all the people
+        about to lose their access.
+
+        Hooked here because it is what both the delete view and the
+        ``delete_selected`` action call, so one override covers a single delete
+        and a bulk one alike.
+        """
+        deletable, model_count, perms_needed, protected = super().get_deleted_objects(objs, request)
+
+        losses = []
+        for permission_group in objs:
+            holders = permission_group.group.user_set.count()
+            losses.append(
+                format_html(
+                    "Group: {} — revoked from {} member{}",
+                    permission_group.group.name,
+                    holders,
+                    "" if holders == 1 else "s",
+                )
+            )
+
+        if losses:
+            deletable = [*deletable, *losses]
+            model_count = {**model_count, "groups": len(losses)}
+        return deletable, model_count, perms_needed, protected
+
 
 @admin.register(PermissionGroupTemplate)
 class PermissionGroupTemplateAdmin(admin.ModelAdmin):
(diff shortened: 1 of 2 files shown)


mikefeldberg (approved):
in a similar vein, we should probably make these read-only. This is in the org admin form and looks like it's touching a permission group, but it's actually touching a template

Image

vecchp:
@mikefeldberg you were right about the template field, and it is worse than read-only would suggest.

`PermissionGroupTemplate` is **one row shared by every organization holding the role**. The inline's `template` field is wrapped in Django's `RelatedFieldWidgetWrapper`, so the organization page renders add / change / delete links straight into `/admin/accounts/permissiongrouptemplate/…`:

```
can_add_related: True   can_change_related: True   can_delete_related: True
<a class="related-widget-wrapper-link change-related" id="change_id_template" …
```

The form already disables the *select* for an existing row, but `disabled` does not touch those links — so from one organization's page you could edit or delete a role every organization shares.

And what they open is either inert or global. For a role the code names (`Organization Admin`, `Caseworker`, …), `sync_group_permissions` re-applies its `TemplateConfig` on the next `migrate` or reconcile, so an edit there is undone without saying so — the same thing as the Active switch this PR dropped. For a role created in the admin, its permissions *are* the definition, so editing it changes every organization holding it.

#2380 has the fix: those three links off (view stays, since what a role grants is worth seeing), code-owned roles read-only in the template admin, hand-defined ones still editable, and a code-owned role can no longer be deleted — that delete used to look successful and then break the next reconcile.

Your naming comments on #2378 are applied too, including dropping the relative docstring. The "likelier than what?" point was fair and I've made it a standing rule rather than a one-off fix.

PR #2378, opened 2026-08-25
refactor(accounts)!: make PermissionGroup a subclass of auth.Group (DEV-2552)

vecchp:
> Stacked on #2377 — review that one first. Base changes to `main` once it merges.

## Why

`PermissionGroup` and its `auth.Group` have always been one thing with two identities, held together by a `post_delete` receiver. That receiver existed only because the foreign key pointed from the row *to* the group, so the teardown ran against the cascade rather than with it. Its own docstring explains why it could not be a `delete()` override: the paths that matter — queryset deletes from reconciliation, and organization cascades — skip the override entirely.

**The teardown is load-bearing, not tidiness.** Object-level permissions for notes, tasks, referrals and client documents are assigned to the group, and `BigGroupObjectPermission.group` is `FK(Group, on_delete=CASCADE)`. A group outliving its row keeps granting every object-level permission ever assigned to it, to every member, with no `PermissionGroup` left to revoke through. Nothing tested that; something does now.

Inheritance makes it structural. `CASCADE` calls `collector.collect()` with `keep_parents` defaulting to `False`, and `collect()` walks `_meta.parents` — so the group goes with the row on a direct delete, a queryset delete, and an organization cascade alike. All three paths the receiver was written to cover, handled by machinery Django already runs.

## What goes away

- `delete_orphaned_group` and its `post_delete` wiring in `apps.py`
- the `save()` hook that created the backing group
- the `("organization", "group")` unique constraint
- `PermissionGroupAdmin`'s `group` column
- `select_related("group")`, `permission_groups__group__user`, `permission_group.group` — 37 files, **208 insertions against 245 deletions**

`assign_object_permissions(permission_group, obj, perms)` needed nothing beyond dropping `.group`: guardian branches on `isinstance(…, Group)`, an MTI child is one, and its pk *is* the group pk. One GraphQL mutation lost a query — the role lookup and the group it needed are one row now, so `test_add_organization_member` went from 21 to 20.

## The rename

`name` is the group's now — the unique `Organization [pk] · Role` key `group_name()` builds — so the human role label moves to `label`.

**No data moves.** `auth_group.name` already held the unique key and `accounts_permissiongroup.name` already held the label; they were in the two right tables all along. The migration renames one column and promotes a foreign key the table already carried as `NOT NULL UNIQUE` to primary key.

The autodetector cannot express a change of bases — it emits `DeleteModel` + `CreateModel`, which drops every row — so the swap is hand-written under `SeparateDatabaseAndState`.

## Deployment — this migration is breaking, do not merge it into a routine deploy

`0007` is **not backward-compatible**, and migrations run before the tasks roll. Old containers query `accounts_permissiongroup.id`, `.name` and `.group_id`; the moment the migration applies, none of those columns exist and every org-scoped permission query fails until the new task set is serving.

What that costs, stated plainly:

- authenticated GraphQL fails for the length of one rollout, then heals by itself
- no data is at risk — `0007` is a single transaction on Postgres, so there is no half-applied state and nothing to repair
- unauthenticated reads are unaffected
- the migration itself is milliseconds on ~300 rows; the exposure is entirely the rollout gap

**Expand/contract does not help here.** The usual pattern — add the new column, ship code that reads both, drop the old one — works for columns. This is a primary key swap, and the new code cannot run against the old schema at all, because multi-table inheritance is a model-level fact rather than a column. The only genuinely zero-downtime route is a table copy: build the new-shaped table alongside, backfill, keep both in sync with triggers, repoint the code, drop the old one in a later deploy. Two tables, bidirectional triggers and two deploys for a 300-row table — several times the work and risk of this change.

So this wants a chosen window rather than a mitigation. Merging it is a decision about when to deploy, not just about the code.

## Testing

- `1474 passed, 31 skipped` — full suite. `ruff check`, `ruff format --check`, and `mypy` clean on 388 files with the cache cleared.
- **The deletion property was already pinned.** `test_permission_group.py` covered all three paths — direct, queryset, organization cascade — before this change and is untouched apart from saying `permission_group` where it said `permission_group.group`. Those passing with the receiver deleted is the evidence.
- Mutated to prove they are not vacuous: making `delete()` pass `keep_parents=True` fails both the direct-delete test and the new object-permission test. The queryset and cascade tests are unaffected by a `delete()` override — which is exactly the point the old docstring made.
- **Migration round-trip** on a scratch database: forwards, backwards, forwards. Reverse restores every original column and constraint, `django_migrations` drops the row, and every seeded `PermissionGroup` came back out the far side pointing at the same `auth.Group` with the same label.

## Two things worth a reviewer's attention

**A removed test.** `RestoreGroupNamesTestCase.test_reverse_survives_two_organizations_sharing_a_name` called `0005`'s reverse function with the live `apps` registry, which only worked while the live model matched `0005`'s state. After this, reversing `0005` necessarily goes through `0007` first, which restores exactly the shape that function expects — so the function is unchanged and still correct, but the shortcut the test took no longer exists. The forward property it guarded is still covered by `test_two_organizations_may_share_a_name`.

**`PermissionGroup.objects` is redeclared** with a `type: ignore[assignment]`. django-stubs types `Group.objects` as `GroupManager`, i.e. `Manager[Group]`; inheriting it resolved every query on this model against the parent's fields and produced 100 spurious errors. The override is what keeps `PermissionGroup.objects.filter(organization=…)` type-checked.

## Note for the #2372 rebase

#2372's `_roles_lost_to_row_deletion` reads the inline's primary-key field as `-id`. The inline's pk is `group_ptr` now, so that lookup silently stops matching — it needs `PermissionGroup._meta.pk.name`, and its `_payload` helper needs to post `-group_ptr`.

## Summary by Sourcery

Make PermissionGroup a first-class auth group with inherited lifecycle and permissions while preserving existing role data and access behavior.

Bug Fixes:
- Ensure deleting permission groups also removes inherited object-level permissions and associated auth groups across all deletion paths.
- Prevent code-managed permission templates from being edited or deleted in the admin, and clarify role revocation warnings.

Enhancements:
- Make PermissionGroup inherit directly from auth.Group, consolidating group identity, membership, permissions, and lifecycle management.
- Rename the human-readable role field to label while retaining generated group names and migrate the existing schema without moving data.
- Simplify organization role and permission queries to use PermissionGroup directly and reduce redundant database access.

Deployment:
- Add a breaking migration for the PermissionGroup primary-key and inheritance change; deploy it in a coordinated rollout because older containers cannot query the migrated schema.

Tests:
- Update coverage for inherited group deletion, object-permission cleanup, and admin behavior while preserving direct, queryset, and cascade deletion guarantees.

Chores:
- Remove the orphaned-group post-delete signal, backing-group creation hook, group relationship fields, and obsolete migration test.


apps/betterangels-backend/accounts/tests/test_permission_group.py
```diff
@@ -80,16 +81,33 @@ def test_group_without_a_template_has_no_permissions(self) -> None:
...
     def test_deleting_permission_group_also_deletes_associated_group(self) -> None:
         permission_group = permission_group_recipe.make()
-        group_id = permission_group.group_id
+        group_id = permission_group.pk
 
         permission_group.delete()
 
         self.assertFalse(Group.objects.filter(id=group_id).exists())
 
+    def test_deleting_a_permission_group_takes_its_object_permissions(self) -> None:
```
mikefeldberg: `takes` -> `deletes`/`removes`/`revokes`. Based on the test just above, `deletes` is best

apps/betterangels-backend/accounts/tests/test_admin.py
mikefeldberg: `takes` -> `deletes`

mikefeldberg: Change name and docstring to be specific and descriptive. e.g., `test_the_bulk_delete_page_names_the_group_and_who_loses_it`
I don't think the docstring is even necessary with the updated name.

`"""``delete_selected`` is the likelier route, and goes through the same hook."""`
 Pretty surprised at this relative-naming. `likelier` than what? `same hook` as what? Mine did a similar thing. I added filters and it wrote a test called `test_new_filters`. Just weird

mikefeldberg: nit, but `test_is_active_not_available_in_form` or `test_form_hides_is_active` seem so much more readable

mikefeldberg (approved):

vecchp:
Three things found while merging `main` in, for the record rather than as blockers. Mike's four naming comments are applied.

## 1. This removes a database-level refusal, quietly

Deleting a code-owned `PermissionGroupTemplate` behaves differently before and after this PR:

| | Deleting e.g. the `Organization Admin` template |
| --- | --- |
| `main` | `reconcile_org_groups` creates derived rows with `get_or_create(organization=…, template=…)` and no name, so `name=""`. `SET_NULL` then violates `permission_group_has_template_or_name` — **the database refuses the delete**. |
| here | Derived rows carry `label`, so the constraint is satisfied. The delete **succeeds**; each row survives holding its label and its members with `template_id = NULL`, which reconciliation no longer treats as derived. `post_migrate` re-seeds the template, and the next reconcile's `get_or_create` collides with the orphan: `IntegrityError: duplicate key value violates unique constraint "auth_group_name_key"`. |

So the failure moves from the destructive action to a later, unrelated-looking reconcile. Verified by probing both branches, not inferred.

#2380 refuses that delete in the admin, which closes the route a human takes. Nothing backs it up for a queryset delete, a shell, or a data migration — worth knowing, and arguably worth a constraint, but not something this PR has to solve.

## 2. `test_the_permission_group_inline_does_not_offer_the_group_field` is now vacuous

Its assertion is `assertNotIn("group", …empty_form.fields)`. Inheritance removes the `group` field, so that cannot fail regardless of the inline's configuration — it passes for a reason unrelated to what it was written to guard. Its docstring also still explains that `group` "is created and torn down by code", describing `delete_orphaned_group`, which this PR deletes.

Either drop it, or repoint it at the property that still needs guarding.

## 3. mypy: local and CI disagree

`uv run mypy --config-file=../../mypy.ini .` from `apps/betterangels-backend`, cache cleared, django-stubs 5.2.9 as `uv.lock` pins — **30 errors**, all `Cannot resolve keyword 'user' into field. Choices are: group, group_id, id, name, organization, …` on `PermissionGroup.objects.filter(…, user=…)`. Four of those are mine, from translating `main`'s new `group__user=` queries onto this model; the other 26 predate the merge.

CI's run of the identical command on `355bc3d0` was green in 17.4s. Both cannot be right, and the choices list mypy prints is the pre-inheritance field set. Worth re-running that target with the nx cache skipped before trusting the green.

## The merge itself

`main` had #2377 squashed, so its commits appear on both sides. The auto-merge reported clean while leaving `get_deleted_objects` **defined twice** in `PermissionGroupAdmin` — this branch's version, then `main`'s pre-inheritance one, which wins. It calls `permission_group.group.user_set`, so the permission-group delete page 500s with `AttributeError: 'PermissionGroup' object has no attribute 'group'`. Confirmed by restoring the duplicate: three delete-warning tests fail. Removed in the merge commit.

Also: `test_add_organization_member` drops to 19 queries, because #2371 replaced `Site.objects.get(pk=settings.SITE_ID)` with `base_url_for()`.

Full suite after the merge: **1486 passed, 31 skipped** on a freshly created database; `ruff` clean; `makemigrations --check` reports no changes.

PR #2379, opened 2026-08-25
fix(teams): refuse to delete a team that notes or tasks reference (DEV-2556)

vecchp:
## Why

The admin Teams page works on `main` — an org admin can create, rename, deactivate and delete teams. Three things made it unsafe to hand to an organization.

## 1. It reported "Sorry, something went wrong" for errors it had the words for

`teams/services.py` writes a message naming the duplicate — `A team named "Drop-in Center" already exists in this organization.` — and the drawer *extracted it, threw it, and then replaced it with the generic string in the `catch`*. Renaming a team into a collision is the likeliest error on this page and it said nothing.

Both handlers now show the `OperationInfo` message where it is extracted and return there. The `catch` keeps the generic string, so the copy written in `teams/services.py` reaches the user and an Apollo network failure does not.

## 2. It held the first 100 teams while claiming to search all of them

The `AdminTeams` query declares `$filters` and `$pagination`; the page passed neither, so `PAGINATION_DEFAULT_LIMIT` applied and search, sort and the inactive toggle all ran client-side over that one page. Now it asks for the whole list, as `useOrgTeams` already does for the app.

Kept client-side rather than moving the toggle server-side: once the full list is present, toggling costs nothing, and a refetch per toggle would be a regression.

## 3. Deleting a team in use silently detached history

`Note.team` and `Task.team` were `SET_NULL` — the only two FKs to `Team` — so a delete detached the team from every note and task that used it, behind a bare `window.confirm`, with `BaseModel` carrying only `created_at`/`updated_at`.

**The refusal is now the foreign key, not a check in the service.** Both become `on_delete=RESTRICT`, so Django's deletion collector refuses it — which covers the writers a service check cannot reach:

- `TeamAdmin` is a plain `ModelAdmin` with delete enabled and no override, so a staff user deleting there never touched `team_delete`
- `queryset.delete()` and a shell session reach no service at all

`team_delete` translates `RestrictedError` into the `ValidationError` the mutation layer already turns into `OperationInfo`, and counts the referencing records **only once the delete has been refused**:

> `Cannot delete "Drop-in Center": it is used by 12 notes and 3 tasks. Deactivate it instead — an inactive team is hidden in the app but keeps its history.`

So the success path no longer pays for two aggregates, and `deleteTeam` drops from **9 queries to 7** — fewer than `main`, because `SET_NULL` was issuing UPDATEs the `RESTRICT` path does not need. An unused team — the typo case — still deletes outright.

### Why `RESTRICT` and not `PROTECT`

Checked against Django's collector rather than assumed. `PROTECT` raises unconditionally, so deleting an **Organization** would fail: Teams cascade, and then its notes hit the protected FK. `RESTRICT` is the precise tool, per the comment at [`deletion.py:408`](https://github.com/django/django/blob/main/django/db/models/deletion.py):

> Raise an error if collected restricted objects (RESTRICT) aren't candidates for deletion also collected via CASCADE.

`Note.organization` is `CASCADE`, so an organization delete still removes its notes and teams together. Switching to `PROTECT` fails the service and GraphQL refusal tests, which expect the `RestrictedError` that `team_delete` translates.

### Both migrations are what `makemigrations` wrote

A plain `AlterField`, unedited. Worth knowing that it is not a no-op: Django 6.0 dropped `on_delete` from `Field.non_db_attrs` when it added `DB_CASCADE`/`DB_SET_NULL`, so `sqlmigrate` shows an identical foreign key dropped and re-added, and the re-add revalidates the table.

```sql
SET CONSTRAINTS "notes_note_team_id_4f22a287_fk_teams_team_id" IMMEDIATE;
ALTER TABLE "notes_note" DROP CONSTRAINT "notes_note_team_id_4f22a287_fk_teams_team_id";
ALTER TABLE "notes_note" ADD CONSTRAINT "notes_note_team_id_4f22a287_fk_teams_team_id"
  FOREIGN KEY ("team_id") REFERENCES "teams_team" ("id") DEFERRABLE INITIALLY DEFERRED;
```

Skipping it with `SeparateDatabaseAndState` is possible and was tried, but that trades an obvious migration for an unusual one to save a cost nobody has measured on these tables. If `notes_note` turns out to be large enough for the revalidation to matter at deploy time, that is worth doing deliberately and with a number attached.

## One behaviour change worth calling out

**Deleting an organization is now refused while a task still holds one of its teams.** `Task.organization` is `SET_NULL`, so the task outlives the cascade and is therefore not a candidate for deletion — so `RESTRICT` fires. Notes are unaffected, because `Note.organization` is `CASCADE`.

No existing test hits this, and being made to deal with live task history before deleting an organization is defensible, so it is left as it stands. The alternative is a `pre_delete` receiver nulling those `team_id`s — happy to add it if you would rather the org delete went through.

## Also here

- The row's action button is icon-only and had no accessible name; it gets an `aria-label` naming the team it acts on.
- `libs/react/betterangels-admin/vite.config.ts` gains the jsdom + `svgTestResolver` block, byte-identical to `libs/react/shelter-operator/vite.config.ts`. That library had **no test environment and zero component tests**; these are its first.
- `delete_team_fixture` selects `OperationInfo` as well as `DeletedObjectType`, so a refusal is assertable.
- `team_delete` loses its `@transaction.atomic`. `Collector.collect()` raises `RestrictedError` before any SQL runs, and `Model.delete()` opens its own atomic block, so the decorator was not doing anything the delete did not already do.

## Testing

- **1488 passed, 31 skipped** — full backend suite on an isolated database. `mypy` clean over 390 files with the cache cleared; `ruff check` and `ruff format --check` clean.
- `makemigrations --check --dry-run` → no changes detected.
- `nx test react-betterangels-admin` — 3 passed (new). typecheck and lint clean for `react-betterangels-admin`, `betterangels-admin`, `ba-platform`.
- `export_schema` reproduces `schema.graphql` — **no schema change**.

Every behaviour mutation-tested:

| Mutation | Fails |
| --- | --- |
| `Note.team` / `Task.team` back to `SET_NULL` | 4 — three service tests and the GraphQL refusal test |
| `PROTECT` instead of `RESTRICT` | the same 4 — `team_delete` catches `RestrictedError`, not `ProtectedError` |
| always pluralise the noun (`f"{count} {noun}s"`) | the singular cases, which is why they assert `used by 1 note.` with the period |
| the drawer shows the generic string instead of the `OperationInfo` message | the duplicate-name test |
| `TeamsPage` shows the generic string instead of the `OperationInfo` message | the delete-refusal test |
| drop the explicit `pagination` variable | the "asks for every team" test |

## Stacked on top of this

- **`feat/teams-restrict-delete`** — `TeamType.isInUse`, so the page offers *Deactivate* instead of a Delete this PR would refuse. That is the proactive half; this PR is the invariant and the backstop.
- **`test/teams-cross-org-mutations`** — pins cross-org refusal at the mutation boundary.

## Not included

- `UserTeamPreferenceSelect` and `FilterTeamsOptions` neither filter nor label inactive teams, and `useOrgTeams`'s docstring claims an active-only default it does not implement. Separate PR — and the docstring is what should change, since the forms need the full list to keep an already-selected inactive team.
- No record of *who* created or deleted a team; `BaseModel` has timestamps only. Worth its own decision.

## Summary by Sourcery

Protect team history by refusing deletion of referenced teams while improving team administration feedback and complete-list behavior.

New Features:
- Prevent deletion of teams referenced by notes or tasks and provide actionable validation messages recommending deactivation.
- Load the complete team list so client-side search, sorting, and inactive filtering apply across all teams.

Bug Fixes:
- Display server-provided operation errors for team creation, updates, and deletion instead of replacing them with generic error messages.

Enhancements:
- Add accessible labels to team action buttons.

Build:
- Configure the admin package for jsdom component testing and SVG resolution.

Tests:
- Add backend and frontend coverage for deletion refusals, organization deletion behavior, server error display, and full-list team queries.

Chores:
- Add state-only migrations aligning note and task team relationships with restricted deletion behavior.


apps/betterangels-backend/tasks/migrations/0005_alter_task_team.py
```diff
@@ -0,0 +1,34 @@
+import django.db.models.deletion
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+    """Point ``Task.team`` at ``RESTRICT`` so a team in use cannot be deleted.
+
+    See ``notes.0004_alter_note_team`` -- ``on_delete`` has no database
+    representation, so re-creating the identical foreign key would lock
+    ``tasks_task`` to prove nothing.
```
mikefeldberg: man... these comments

apps/betterangels-backend/notes/migrations/0004_alter_note_team.py
```diff
@@ -0,0 +1,37 @@
...
+
+class Migration(migrations.Migration):
+    """Point ``Note.team`` at ``RESTRICT`` so a team in use cannot be deleted.
+
+    ``on_delete`` is enforced by Django's deletion collector, not by the database:
+    both ``SET_NULL`` and ``RESTRICT`` compile to a plain foreign key with no
+    ``ON DELETE`` clause.  A plain ``AlterField`` would still drop and re-add that
+    identical constraint, and re-adding it validates every row in ``notes_note`` --
+    one of the two largest tables -- under a lock, for no schema change.  So the
+    database half is deliberately empty.
```
mikefeldberg: `makemigrations` could have avoided this bloviating lol

apps/betterangels-backend/teams/tests/test_models.py
```diff
@@ -0,0 +1,76 @@
...
+from django.db.models import RestrictedError
+from django.test import TestCase
+from model_bakery import baker
+from notes.models import Note
+from tasks.models import Task
+
+from teams.models import Team
+
+
+class TeamDeletionTestCase(TestCase):
```
mikefeldberg: this seems unnecessary. We didn't add a model method, we just updated the `on_delete`; this is basically testing django

apps/betterangels-backend/teams/services.py
```diff
@@ -63,10 +65,28 @@ def team_update(
...
-    """Hard-delete a Team. FK references are SET_NULL by the database."""
-    team.delete()
+    """Delete a Team that nothing references.
+
+    The refusal is the ``RESTRICT`` on ``Note.team`` and ``Task.team``, not a check
+    here -- so it also covers the Django admin, ``queryset.delete()`` and a shell
+    session, none of which reach this function.  ``is_active`` retires a team
+    reversibly, which is what the message points at.  The counts are gathered only
+    once the delete has already been refused.
+    """
```
mikefeldberg: Is Claude practicing its poetry in these docstrings? I don't think this fn needs a docstring at all, since the body is clear in its intent. If we really want one, I think something much simpler would suffice. e.g., 

```suggestion
    """Delete a Team. Deletion of Teams associated with a Note or Task is restricted by the field's ``on_delete``"""
```

libs/react/betterangels-admin/src/lib/pages/teams/TeamsPage.tsx
```diff
@@ -120,10 +127,13 @@ export function TeamsPage(props: IProps) {
...
     } catch (err) {
-      console.error(err);
+      const error = toError(err);
+
+      console.error(`error deleting team: ${error.message}`);
+
       showAlert({
         type: 'error',
-        content: 'Sorry, something went wrong. Please try again.',
+        content: error.message,
```
mikefeldberg: I know these errors are unlikely to reveal anything sensitive but we typically display something generic here. Relevant to what the user was doing, but not the actual error text.

libs/react/betterangels-admin/src/lib/pages/teams/TeamsPage.tsx
```diff
@@ -120,10 +127,13 @@ export function TeamsPage(props: IProps) {
       });
       refetch();
     } catch (err) {
-      console.error(err);
+      const error = toError(err);
+
+      console.error(`error deleting team: ${error.message}`);
```
mikefeldberg: we use this pattern elsewhere
```suggestion
      console.error(`[deleteTeam error]: ${error.message}`);
```

apps/betterangels-backend/teams/services.py
mikefeldberg: the pr title can also use a rephrase. If I read `report what the server said, and refuse at the FK a delete that strands history` in a log I don't think I'd have a good idea of what was done here

mikefeldberg (approved):

PR #2381, opened 2026-08-26
camera in bottomsheet with e2e tests (DEV-2487, DEV-2524)

tglaz:
### e2e: upload and delete client document
DEV-2487
### Fix expo Doc MediaPicker freezing issues
DEV-2524

- mocks expo-camera instance for e2e tests (iOS only)
  - expo-comera is a mock in iOS anyway, and it would freeze the simulator which made e2e testing almost impossilbe
- replace `CameraModal.tsx` with `CameraSheet.tsx`
  - camera now rendered via BottomSheet
- remove redundant header + close btn from UploadModal

## Upload Modal with inner header removed (title + Done btn on top of screen only)

image

## Summary by Sourcery

Replace the camera modal with a full-screen bottom-sheet experience and strengthen document upload coverage across platforms.

New Features:
- Present camera capture as a full-screen bottom sheet with permission handling, capture controls, and simulator-safe camera behavior.
- Add end-to-end coverage for capturing, uploading, and deleting client documents.

Bug Fixes:
- Prevent camera capture and navigation from leaving an unresponsive or transparent modal view on iOS.
- Avoid camera initialization delays in iOS simulators by using a placeholder preview and generated test photos.

Enhancements:
- Improve bottom-sheet defaults and full-screen behavior, including accessibility, stacking, sizing, and over-drag handling.
- Add stable test identifiers across client documents, media picker, camera controls, and document actions.
- Move camera controls into dedicated header and footer components and stop the camera preview after capture or cancellation.

CI:
- Make camera permission handling reliable across iOS and Android Maestro runs.

Tests:
- Add Maestro coverage for the client document upload and deletion flow.

Chores:
- Remove the obsolete upload modal completion test and replace the camera modal implementation with the bottom-sheet implementation.

## Summary by Sourcery

Replace the camera modal with a simulator-safe full-screen bottom sheet and expand end-to-end document upload coverage.

New Features:
- Present document capture through a full-screen camera bottom sheet with permission handling and capture controls.
- Add end-to-end coverage for capturing, uploading, and deleting client documents.

Bug Fixes:
- Prevent iOS camera and navigation flows from becoming unresponsive during capture and dismissal.
- Avoid iOS simulator camera initialization freezes by using a placeholder preview and generated photos.

Enhancements:
- Improve bottom-sheet defaults and full-screen behavior for camera presentation.
- Add stable test identifiers across document, media picker, camera, and file actions.

CI:
- Make camera permission handling reliable across iOS and Android Maestro runs.

Tests:
- Add Maestro coverage for the client document upload and deletion flow.

Chores:
- Replace the legacy camera modal with the bottom-sheet implementation and remove the obsolete upload completion test.


tglaz:
e2e tests passed in CI

vecchp:
The sheet-instead-of-`Modal` reasoning here holds up, and the fiddly parts are each right — `overDragResistanceFactor: 0`, `contentStyle: { height: '100%' }` (gorhom really does absolutely-position the content view, so `flex: 1` is inert), `enableDynamicSizing: false` alongside a fixed snap point. Two things I think need addressing before this merges.

**1. The e2e flow can pass with capture, upload and delete all broken.**

- The "upload landed" check is `extendedWaitUntil visible: client-docs-accordion-forms`, but that accordion only renders when the client already *has* consent-form documents (`hasConsentFormDocuments`, `Docs/index.tsx:93`). It proves the upload worked on the first run and passes before the capture on every run after.
- `id: 'file-card-.*jpg'` matches any jpg in the folder, and the delete half asserts nothing at all — tap `index: 0`, confirm, `waitForAnimationToEnd`, end. A delete that silently fails looks identical to one that works.
- The fixture client persists across runs and this flow leaves a jpg behind whenever it deletes someone else's, so it decays; `retries: 2` compounds it (a run that fails after uploading hands the retry a pre-existing jpg).

Cheapest fix that doesn't wait on DEV-2498: `assertNotVisible id: 'file-card-.*jpg'` inside the Forms accordion *before* capturing, and again after confirming the delete. That turns "there is a jpg here" into "we put one here and removed it".

Separately, the gate at line 23 can be satisfied by the loading state. `tapOn: 'See Details'` → `waitForAnimationToEnd` → `assertVisible: id: 'client-screen'` doesn't wait on `ClientProfileQuery` (every other wait in the file uses `extendedWaitUntil`), so on a cold backend the flow clears its gate against the spinner and then scrolls for `client-tab-docs` on an empty screen. You already added `client-screen-loading` for exactly this — worth wiring it up: `extendedWaitUntil notVisible: id: 'client-screen-loading'` before the `client-screen` assert. Worth renaming it too, since it's currently a prefix extension of the id being matched.

**2. The camera sheet isn't full-screen on the HMIS upload path.**

`CameraSheet`'s `snapPoints: ['100%']` resolves against whichever container `BottomSheetLayoutProvider` registered, which inside the modal-screen route is the route's content view — gorhom's hosting container is `absoluteFill` inside the nested provider at `modal-screen.tsx:19`. `mode: 'custom'` is what makes that the whole window, and only the BA Docs path passes it. `ClientDocsViewHmis.tsx:70` calls `showModalScreen` with no header config, so there the camera stops below the native "Upload Files" bar and that bar's close button stays live above a supposedly full-screen camera. The new flow only covers the BA path, so CI won't see it.

Either give `ClientDocsViewHmis` the same `header: { mode: 'custom', closeLabel: 'Done' }`, or have `CameraSheet` measure the window rather than the sheet container so the full-screen claim holds on every host. Worth a look at `ProfilePhotoUploaderHmis`, `ClientProfilePhotoUploader` and `RestDebugHmis` too — the `MediaPicker` swap lands in all three untested.

---

I have more (accessibility, a contrast regression in the new footer, the mocked camera reaching device previews) but it's smaller — happy to hold it until you've looked at these two.

vecchp (approved):

tglaz:
regarding:

> 1. The e2e flow can pass with capture, upload and delete all broken.

Yeah, these tests are not as concrete as I wanted, but here's the background:
- the `upload landed` assertion is pretty soft
  - i'm using this `file-card-.*jpg'` which can match any jpg, but with current "generic" user that's just a limitation. Without a dedicated user that's guaranteed to have 0 existing uploads, not sure of a better approach that's not expensive. If we add UI to enable naming the uploaded file, we'd be able to validate that.
  - I don't validate the `flash message` as it's temporal, and even today we show it for longer than we probably should. Validating it would open the test up to more random flakiness, though I can try and see how it goes
  - i don't want to check for `client-screen-loading` because we're likely to miss it and test would fail (it's super small file so upload should be fast)

> 2. The camera sheet isn't full-screen on the HMIS upload path.

I ignored it on purpose, as it still should work. And not sure if I should spend the time. But I could I suppose.

shiozakiyuko (commented):
Hi @tglaz 

Tested on both iOS and Android. Once logged into the Android app, everything ran smoothly overall.

However, on Android, uploading a 202KB PDF got stuck in the "Queued" state for a few minutes before completing (unclear if this is related to the login issue). On iOS, a similar-sized PDF uploaded smoothly without delay.

Image

PR #2385, opened 2026-08-26
update OpenNowInput to accept includeUnknown (SDB-259)

tglaz:
[Filter - Open Now] Add Include Unknown option - BE
SDB-259

* updates `OpenNowInput` to accept `includeUnknown`
  * implements in `open_now` filter query
 

## Summary by Sourcery

Allow the Open Now shelter filter to optionally include shelters with unknown schedule status.

New Features:
- Add an `includeUnknown` option to the Open Now shelter filter, allowing shelters without a defined schedule for a requested type to be included.

Enhancements:
- Preserve existing Open Now behavior while unioning open and unknown shelters per requested schedule type.

Tests:
- Add coverage for unknown shelters, exception-only schedules, multiple schedule types, and no-op filter cases.


apps/betterangels-backend/shelters/types/filters.py
```diff
@@ -196,15 +197,31 @@ def open_now(
...
+        open_qs = shelters_open_at(
+            queryset,
+            dt=get_current_shelter_schedule_datetime(),
+            schedule_types=value.schedule_type,
         )
 
+        if not value.include_unknown:
+            return open_qs, Q()
+
+        # "Unknown" for a schedule type = the shelter has no non-exception
```
mikefeldberg, line 209: this can return false positives. We allow users to create exceptions even if there's no "open" schedule, so it's possible for a shelter to have _only_ exceptions. This filter would return that shelter, even if an exception covered the time of search. 

We can either update the filter to account for exceptions or prevent users from creating exceptions without accompanying "open" schedules. I think the latter is more "correct" since it doesn't really make sense to have only exceptions. Up to you, tho. Adding the exception restriction would be oos so we can make a ticket for that.

mikefeldberg (approved):

apps/betterangels-backend/shelters/types/filters.py
tglaz: I assumed that we couldn't create an `exception` without a regular schedule, but regardless, something can update later?
i'd think that if they don't have a regular schedule than it's pretty safe to assume it's "unknown" - and the consequence of being wrong is negligible.

PR #2386, opened 2026-08-27
add unknown filter to shelter openNow options (SDB-259)

tglaz:
### [Filter - Open Now] Add Include Unknown option
SDB-259

- FE implementation (BE already done)

## Summary by Sourcery

Add support for including shelters with unknown open-now status in shelter search filters.

New Features:
- Add an “Include unknown” option to the shelter “Open Now” filter, enabled when at least one schedule type is selected.
- Include the unknown-status choice in active filter pills and allow it to be cleared independently.

Bug Fixes:
- Ensure the include-unknown setting is cleared when no open-now schedule types remain selected.

Enhancements:
- Rename and align shelter open-now filter state and result-source handling with the updated filter model.
- Improve checkbox styling for enabled and disabled states.


mikefeldberg (approved):

shiozakiyuko (commented):
Hi @tglaz 

I noticed the Include Unknown is disabled. Does it mean that every shelters we have in our database has some kind of Open Now data?

tglaz:
@shiozakiyuko - a user shouldn't be able to select `unknown` until they select a type - otherwise it makes no sense

shiozakiyuko:
@tglaz Ok. I see. Can you include "Select a filter option first" next to it so it'll be clearer? e.g. Include unknown: Select a filter option first

shiozakiyuko (approved):

PR #2387, opened 2026-08-27
fix(permissions): stop a successful delete reporting a permission denial (DEV-2510)

vecchp:
### DEV-2510 — a successful delete reports a permission denial

`deleteClientDocument` **deletes the row and then returns a permission error**, and `deleteNote` does the same. The two are indistinguishable from a genuine denial except for `field`, an internal of strawberry-django's `handle_no_permission`:

```
permitted user  → row deleted, OperationInfo{PERMISSION, field: "deleteClientDocument"}
real denial     → row kept,    OperationInfo{PERMISSION, field: null}
```

This is what #2347 is trying to work around from the client. It can't be — which is why that PR's `persist: true` would turn today's transient error into a permanent "You don't have permission to access this app." on a delete that succeeded.

### Cause

Both fields paired `mutations.delete` with `HasRetvalPerm`, which runs the object-permission check twice — once inside the resolver via `get_with_perms` before anything is written, and again on the returned instance afterwards. The returned instance still carries its pk, so the identical check that passed before the delete fails after it, and `fail_silently=True` (the default) surfaces that as an `OperationInfo` rather than an error.

```
success path:  get_with_perms OK pk=3242              ← pre-write check passes
               [row deleted]
               _resolve_obj obj=Attachment pk=3242  → DjangoNoPermission
               handle_no_permission fail_silently=True → OperationInfo{PERMISSION, field:"deleteClientDocument"}

denial path:   get_with_perms RAISED PermissionDenied → row kept, field: null
```

Nothing was unprotected — the pre-write check is what stops a genuine denial.

### Fix

Both fields now follow `delete_service_request`: `PermissionedQuerySet` + `get_object_or_permission_error`, which checks once before the write and lets a denial carry a written message instead of the extension's default. `PermissionedQuerySet`'s own docstring already explains that it skips the parent retval check — same problem, other guise.

The delete itself goes through strawberry-django's own `resolvers.delete`, which restores the pk that `delete()` clears. Without that the success payload fails with `Cannot return null for non-nullable field ClientDocumentType.id`.

Return types are unchanged (`ClientDocumentType`, `NoteType`), so deployed clients selecting `... on ClientDocumentType { id }` keep working. The only `schema.graphql` movement is the directive name.

### Why only these two fields

`clients/schema.py` has four more `mutations.delete` + `HasRetvalPerm` fields, and they're fine. The Caseworker template grants **global** DELETE on `ClientContact`, `ClientHouseholdMember`, `HmisProfile` and `SocialMediaProfile`, so their retval check resolves from the group permission and never touches the row. `Attachment` and `Note` get ADD + VIEW only — their DELETE is granted per-object at creation, so it dies with the row. `Task` and `ServiceRequest` are also object-permissioned, but `delete_task` and `delete_service_request` already use the correct pattern.

### Tests

All four existing tests passed against the bug. `test_delete_client_document` asserted only that the row was gone; `test_delete_note_mutation` asserted `assertIsNotNone(...["deleteNote"])`, which an `OperationInfo` satisfies; both permission tests only counted rows.

The delete tests now assert the payload, and the permission tests are split by outcome — permitted / denied / anonymous — rather than branching on a `should_succeed` flag. That matters here specifically: in the branched form an assertion lives behind an `if`, so a parameter table that stops reaching a branch silently stops testing it. That is the same shape of dead coverage that let this bug through.

`deleteSocialMediaProfile` had no permission test at all, unlike its three siblings; it gets one.

Verified by reverting both schema files to `main` with the new tests in place: 10 failures across the permitted and denied paths.

Dropping the duplicated check also takes the document delete from 16 queries to 8, and the note from 21 to 13.

### Notes

Full backend suite passes (1496 passed, 31 skipped), `mypy` clean, `ruff` clean, `schema.graphql` regenerated.

@tglaz — #2347 can rebase on this; the `persist: true` blocker goes away once this lands.

## Summary by Sourcery

Correct GraphQL deletion authorization handling so permitted document and note deletions report success while denied operations remain protected.

Bug Fixes:
- Fix client document and note deletions so successful operations return the deleted object instead of a misleading permission error.

Enhancements:
- Apply single pre-delete permission checks to preserve correct denial behavior and reduce deletion query counts.
- Strengthen deletion permission coverage for client documents, notes, and social media profiles, including successful, denied, and unauthenticated cases.

Tests:
- Update deletion tests to verify returned payloads, object removal, permission errors, and unauthenticated access.


apps/betterangels-backend/clients/schema.py
```diff
@@ -621,13 +623,19 @@ def create_client_document(self, info: Info, data: CreateClientDocumentInput) ->
...
-        DeleteDjangoObjectInput,
+    @strawberry_django.mutation(
         permission_classes=[IsAuthenticated],
         extensions=[
-            HasRetvalPerm(perms=Attachment.perms.DELETE),
+            PermissionedQuerySet(model=Attachment, perms=[Attachment.perms.DELETE]),
         ],
     )
+    def delete_client_document(self, info: Info, data: DeleteDjangoObjectInput) -> ClientDocumentType:
+        qs: QuerySet[Attachment] = info.context.qs
```
tglaz, line 633: The current implementation has a **domain-scoping hole**.

## What the code does

```python
qs: QuerySet[Attachment] = info.context.qs
client_document = get_object_or_permission_error(qs, data.id, ...)
return cast(ClientDocumentType, resolvers.delete(info, client_document))
```

`info.context.qs` comes from `PermissionedQuerySet(model=Attachment, perms=[Attachment.perms.DELETE])`. That only filters by **whether the user can delete an `Attachment` row** — it does **not** check that the attachment is actually a client document.

Client documents are stored as `Attachment` rows with `content_type=ClientProfile`, but `Attachment` is also used for **note attachments** and potentially other generic relations.

## The problematic scenario

Your commit message already reveals the key detail: *"the Caseworker template grants them global DELETE"* on `Attachment`. So for a caseworker:

1. `info.context.qs` contains **every** `Attachment` in the system (global DELETE passes for all rows).
2. A caseworker views a case note and copies the ID of one of its attachments (those are plain `Note` → `Attachment` generic relations).
3. The caseworker calls `deleteClientDocument(data: { id: "<note-attachment-id>" })`.
4. `get_object_or_permission_error(qs, data.id)` succeeds because global DELETE covers that row.
5. The note attachment is deleted.

The note itself is left untouched, so none of the note domain's permission checks ever run — `NotePermissions.CHANGE`, the `PermissionedQuerySet(model=Note, ...)`, or the note's org-scoped filters are all bypassed.

## Why this is a real problem

- **Cross-domain deletion** — the "delete client document" endpoint can destroy data belonging to the notes feature (or any other future feature using `Attachment`).
- **Permission bypass** — the caller never needed `Note`-level permissions, only `Attachment` DELETE. If per-object DELETE is granted broadly (e.g., on creation, like `create_note_attachment` does), even non-caseworkers could delete arbitrary attachments.
- **Broken contract** — the mutation's name and type (`ClientDocumentType`) promise client documents; the implementation deletes anything deletable.

## The fix (as suggested earlier)

Filter the queryset to the client-document content type before the lookup:

````python
qs = qs.filter(content_type=ContentType.objects.get_for_model(ClientProfile))
````

You could tighten it further with a `namespace` filter if client documents are a subset of `ClientProfile` attachments. That keeps the pre-delete permission check but makes it impossible to target an attachment outside the client-document domain.

tglaz (approved):

apps/betterangels-backend/clients/schema.py
vecchp: Good catch — the endpoint does delete attachments outside the client-document domain, and it's now fixed in #2394.

Two corrections to the mechanism, because they change how urgent this is.

The Caseworker template doesn't grant global DELETE on `Attachment` — `notes/groups.py` grants `ADD` and `VIEW` only. The "global DELETE" line in my description is about `ClientContact`, `ClientHouseholdMember`, `HmisProfile` and `SocialMediaProfile`; the sentence after it says Attachment and Note get ADD + VIEW. So `info.context.qs` isn't every `Attachment` in the system, and non-caseworkers can't reach any of them — `filter_for_user` restricts to rows whose object permissions the caller's groups hold, and object DELETE is only ever assigned at creation, to the org's CASEWORKER group.

The real path is narrower: `resolve_note_file_uploads` assigns DELETE on each note attachment to that same group, so those rows *are* in the queryset. No permission is bypassed — the caller does hold DELETE on the row — but the domain boundary is missing, which is your point.

It's also not new here. `main` uses `mutations.delete(...) + HasRetvalPerm(perms=Attachment.perms.DELETE)`, which resolves the `Attachment` by pk with the same absent scoping, so the reach is identical before and after this PR. That's why the fix is its own PR rather than folded into this one.

On the fix itself — rather than filtering in the resolver, #2394 scopes the `ClientDocument` proxy that already exists for the admin, so the model means "an `Attachment` on a `ClientProfile`" and every caller inherits it. `ClientDocumentAdmin.get_queryset`, which was hand-filtering `content_type` for the same reason, gets deleted.

PR #2392, opened 2026-08-27
fix(permissions): org-scope every write PermissionedQuerySet guards (DEV-2567)

vecchp:
> Stacked on #2390. Base retargets to `main` when that lands.

Rebuilt replacement for #2310. **No longer breaking**, and 12 files instead of 29.

## The hole

`PermissionedQuerySet` filtered by object permissions only. Guardian grants those to the permission group that *created* a record, and that grant says nothing about which organization the caller is currently acting as. So a multi-org user who legitimately reached a record through org A could act on it while their header said org B — and **four of the nine call sites had no compensating `.filter(organization_id=...)` in the resolver body at all**: `generateNoteFileUploads`, `resolveNoteFileUploads`, `deleteServiceRequest`, `updateReferral`.

## What this does

`organization_field` becomes a **required** argument on `PermissionedQuerySet` — a field path for an org-owned endpoint, or `None` to declare it deliberately platform-shared.

Making it required is what makes this reviewable: adding a mutation forces an answer, and mypy rejects a call site that omits one. Nine sites declare it. `ServiceRequest` has no organization foreign key, so it names the two paths it reaches one through:

```python
organization_field=["provided_notes__organization_id", "requested_notes__organization_id"]
```

A multi-valued traversal is filtered via `pk__in` on a subquery, so join duplicates stay inside it and a later `.get()` cannot raise `MultipleObjectsReturned`.

`require_organization_id` raises `DjangoNoPermission`, which strawberry-django catches around `resolve_for_user` — resolver *bodies* keep `get_current_organization`. It also validates that a client-supplied header is a usable primary key; unvalidated, a junk header raised `ValueError` from inside the query rather than reading as a denial.

`docs/permissions.md` gains the org-owned vs platform-shared model.

## Why this is a rebuild, not a rebase

#2310 carried 36 commits that are largely earlier forms of work since merged — `git rebase --onto` conflicts on commit 1 of 36. So this compares **trees, not histories**, the same way #2312 was rebuilt. An in-memory `git merge-tree` against #2390 showed 14 of 29 files conflicting, all clustered where the four teams PRs already changed things.

Comparing trees showed what was actually left, and two things dropped out:

- **`allow_implicit_org` — 30 mentions.** The entire compatibility shim. Pointless now: no pre-header builds are live, and #2390 made the last headerless fallback strict. #2310's own body called `grep -rn allow_implicit_org` "the removal list" — so this deletes deliberately temporary code before it ever shipped. **Dropping it is why this PR is no longer breaking.**
- **`active_organization` and `permission_group_for_request`** — both would have shipped unused here. The latter belongs with the de-guessing change below.

Verified that taking whole files reverted nothing: `git diff 9e1ea0ced origin/main` is empty for every file taken wholesale, so `main` has not touched them since #2310 forked.

## Deliberately not included

**The nine call sites that still resolve the *attributed* organization by first match.** The org filter here confines which records a write may *reach*; it does not decide which organization the write is *recorded against*. `docs/permissions.md` now names all nine — #2310's version listed only three, because it assumed its own notes/tasks fix had landed. The follow-up makes `organization_id` required on `resolve_permission_group` so the guessing branch cannot be reached at all.

**Moving `createNote` / `createTask` / `importNote` to `HasOrgPerm`.** Tried, then backed out: `HasOrgPerm` denies with a **raw GraphQL error** where `HasPerm` returned a structured `OperationInfo` with a `PERMISSION` message. The outcome is right either way (denied, nothing created), but that is an error-surface regression on three mutations and deserves its own decision rather than riding along in a rebuild. It cost 10 test updates when attempted.

## Verification

- **1507 passed, 31 skipped** (#2390's branch is at 1491, so +16 — this PR's own new tests). `mypy` clean over 391 files with the cache cleared; `ruff check` and `ruff format --check` clean.
- No schema change.

Mutation-tested:

| Mutation | Fails |
| --- | --- |
| the extension ignores its declared `organization_field` | **5** |
| `organization_field=[]` instead of `None` | the empty-list test — an empty list would silently scope to nothing |

## Summary by Sourcery

Enforce active-organization boundaries across permissioned mutation querysets while preserving deliberate platform-shared access.

Bug Fixes:
- Scope permission-filtered mutation querysets to the caller’s active organization, preventing cross-organization writes through object-permission access.
- Treat missing or invalid organization headers as structured permission denials instead of unhandled query errors.

Enhancements:
- Require every PermissionedQuerySet to explicitly declare an organization field or opt into platform-shared behavior, including support for indirect multi-path organization ownership.
- Add queryset/model validation and document the organization-owned versus platform-shared permission model.

Documentation:
- Document org-scoped mutation behavior, remaining organization-attribution risks, and known cross-organization client deletion risk.

Tests:
- Add coverage for organization scoping, shared-resource opt-out, invalid headers, multi-path relationships, duplicate prevention, and queryset/model mismatches.

### DIFF
--- a/apps/betterangels-backend/accounts/extensions.py
+++ b/apps/betterangels-backend/accounts/extensions.py
@@ -24,7 +24,7 @@ from collections.abc import Callable
 from typing import Any
 
 from accounts.models import Organization
-from common.permissions.utils import permissioned_queryset
+from common.permissions.utils import permissioned_queryset, require_organization_id
 from strawberry.types import Info
 from strawberry_django.permissions import (
     DjangoNoPermission,
@@ -75,11 +75,7 @@ class HasOrgPerm(HasPerm):
         if not user or not user.is_authenticated:
             raise DjangoNoPermission("Authentication required.")
 
-        org_id_raw = info.context.request.organization_id
-
-        if org_id_raw is None:
-            raise DjangoNoPermission("Organization ID (X-Organization-ID header) is required.")
-        org_id = str(org_id_raw)
+        org_id = require_organization_id(info)
 
         if not self.perms:
             raise DjangoNoPermission("No permissions specified for this operation.")
--- a/apps/betterangels-backend/common/graphql/extensions.py
+++ b/apps/betterangels-backend/common/graphql/extensions.py
@@ -1,7 +1,10 @@
-import inspect
-from typing import Any, Callable, Optional, Type, Union
+from functools import reduce
+from operator import or_
+from typing import Any, Callable, Optional, Sequence, Type, Union
 
-from django.db.models import Model, QuerySet
+from common.permissions.utils import require_organization_id
+from django.db.models import Model, Q, QuerySet
+from django.db.models.constants import LOOKUP_SEP
 from strawberry.types.info import Info
 from strawberry_django.auth.utils import get_current_user
 from strawberry_django.permissions import (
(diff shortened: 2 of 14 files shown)


vecchp:
DEV-2567 triage (from the grant-cutover cascade side):

**Current state:** CONFLICTING — needs a rebase either way, and its design predates two things that changed on main/are landing:

1. **`get_current_organization` + the header are going away** (#2450 retirement in the cascade — the org travels in payloads/filters now). This PR's `require_organization_id` reads the header, so a rebase has to re-anchor on the payload org.
2. **Two of the four unguarded sites it fixes are being rewritten by the cutover chain**: the notes slice (RFC 0003 slice 2 — role-back + `can_obj` writes) removes `generateNoteFileUploads` / `resolveNoteFileUploads` / `deleteServiceRequest` from `PermissionedQuerySet` entirely; referrals keep theirs until the referral slice (needs `own_org_or`). Client documents keep theirs until the documents tier (RFC 0002 follow-up).

**Recommendation:** hold until the cascade lands, then either
- **rebase + narrow** to the remaining legacy sites (referrals, HMIS, client documents) re-anchored on the payload org — if those domains are still months out, this is still worth landing for them; or
- **close as absorbed** if the referral/document slices are close behind.

Either way the org-owned vs platform-shared analysis here is valuable — it should survive the rebase even if the `PermissionedQuerySet.organization_field` mechanics get deleted incrementally.

PR #2395, opened 2026-08-31
feat(shelter-operator): PDF export for the Operational Summary report

emilyqqian:
## Summary
- Adds a fixed-Letter-size, multi-page `ShelterReportPrint` component that renders the Operational Summary report (stats, bed status, daily occupancy) off-screen, matching the Figma design.
- `useExportPdf` captures each report page with `html2canvas-pro` (chosen over plain `html2canvas`/`react-to-pdf` for Tailwind v4 `oklch()` support) and assembles them 1:1 onto a jsPDF document, returning a blob per the ticket's acceptance criteria.
- Wires PDF in as a fourth format in Charles's real, backend-integrated `ExportDataModal`/`useShelterMetricsExport` system (`ModalExportFormat = ExportFormat | 'pdf'`), so PDF shares the same file-type picker and "Included in Export" metric checkboxes as CSV/XLSX/JSON, generating client-side instead of hitting the export endpoint.
- Page layout adapts to the selected metrics: Daily Occupancy moves up to page 1 when Bed Status isn't included, instead of leaving page 1 mostly empty.

## Test plan
- [x] `yarn nx run react-shelter-operator:typecheck` — clean
- [x] `yarn nx run react-shelter-operator:test` — 67/67 passing
- [x] `yarn nx run react-shelter-operator:lint` — 0 errors
- [x] `yarn nx run shelter-web:build` — succeeds
- [ ] Manual click-through: open Export Data modal on the Reports tab, select PDF + a few metrics, confirm the download and page layout match Figma

## Summary by Sourcery

Enable selectable client-side PDF export for the shelter Operational Summary report through the existing Reports export workflow.

New Features:
- Add client-side PDF export for the Operational Summary report with selectable metrics and fixed Letter-sized pagination.
- Integrate PDF into the existing Reports export modal alongside CSV, XLSX, and JSON formats.

Bug Fixes:
- Prevent exports while the report data required for PDF generation is loading or unavailable.
- Adapt report pagination so Daily Occupancy uses page one when Bed Status is excluded.

Enhancements:
- Replace the standalone printable-report route and page with an off-screen report renderer integrated into the Reports view.
- Support static chart rendering and metric-specific operational summary layouts for exported reports.
- Return generated PDF blobs for download handling and user feedback.

Tests:
- Add coverage for PDF export wiring, failure states, data readiness, report pagination, metric selection, and multi-page PDF generation.

Chores:
- Add report and modal motion styles with reduced-motion support.


apps/shelter-web/src/assets/styles/global.css
```diff
@@ -42,6 +42,73 @@
   }
 }
 
+@keyframes popIn {
```
frank731, line 45: do we know where we decided we wanted to animate the modal pop in from? or is this a change from another commit that ended up here

libs/react/shelter-operator/src/lib/components/reports/ReportCharts.tsx
```diff
@@ -86,7 +87,17 @@ const STATUS_COLORS = ['#008CEE', '#05B428', '#FF7B00', '#F64949', '#8B5CF6'];
  *
  * Renders empty (no data) until `data` is provided.
  */
-export function BedStatusChart({ data }: { data?: DailyBedStatusMetrics[] }) {
+export function BedStatusChart({
```
frank731, line 90: Image

Image

doesn't seem to be rendering the bed status export correctly sometimes (sometimes it works sometimes it doesn't), is it possible that we need to wait a bit before rendering since it does the animation where the bars grow

frank731: we can also see if we can disable animations on the charts in the export

libs/react/shelter-operator/src/lib/components/reports/ReportFilterBar.tsx
```diff
@@ -52,9 +62,115 @@ function rangeLabel({ from, to }: DateRange): string {
...
+    // the newly-selected metrics to the DOM before exportPdf reads it —
+    // without it, the state update would still be batched when the capture
+    // starts, and it would rasterize last render's content instead. This is
+    // one of the narrow cases React's own docs call out flushSync for:
+    // synchronously flushing before something that reads the DOM directly.
+    // eslint-disable-next-line @eslint-react/dom-no-flush-sync
+    flushSync(() => setPdfMetrics(selectedMetrics));
+
+    try {
+      const filename = `shelter-${shelterId}-report.pdf`;
```
frank731, line 129: i think we should unify all the filenaming with a helper across all file types, realizing id isn't visible to the users so having something like (start date)_(end date)_(shelter name)_report as the naming scheme for all should be good

libs/react/shelter-operator/src/lib/components/reports/ReportFilterBar.tsx
```diff
@@ -76,15 +192,35 @@ export function ReportFilterBar({ shelterId }: { shelterId?: string }) {
       {shelterId && (
         <ExportDataModal
           isOpen={isExportOpen}
-          isExporting={isExporting}
+          isExporting={isBackendExporting || isPdfExporting}
+          disableExport={isPdfDataUnavailable}
```
frank731, line 196: is it possible for this to interfere with the other export types, ex if pdf data fails for some reason and prevents exports to other types?

libs/react/shelter-operator/src/lib/components/reports/ReportFilterBar.tsx
```diff
@@ -52,9 +62,115 @@ function rangeLabel({ from, to }: DateRange): string {
...
+    error: shelterDataError,
+  } = useQuery(GetShelterSummaryDocument, {
+    variables: { id: shelterId ?? '' },
+    skip: !shelterId || !isExportOpen,
+  });
+  const {
+    metrics,
+    loading: isMetricsLoading,
+    error: metricsError,
+  } = useShelterOccupancyMetrics({
```
frank731, line 95: instead of calling this again, is it possible to pass through the data from the outer ReportsView that also ran useShelterOccupancyMetrics?

frank731 (changes requested):
Tried it and its looking clean, some code cleanups and design parity issues remain but nice job getting this working

libs/react/shelter-operator/src/lib/components/reports/ReportCharts.tsx
frank731: also the fonts seem to be a bit off of the designs for the header portion

PR #2401, opened 2026-08-31
fix(accounts): migrate StringAgg off the deprecated PostgreSQL aggregate, upgrade django-stubs to 6.1

vecchp:
Three commits. The first is the small aggregate fix; the second is the stubs upgrade it exposed, plus the errors that upgrade surfaced — **including three live admin bugs**. The third removes the suppressions the second one reached for.

---

## 1. `StringAgg` migration

`annotate_permission_templates` used `django.contrib.postgres.aggregates.StringAgg` with a bare-string delimiter and the `ordering` kwarg. All three are deprecated and warn on every query.

The **delimiter** is the one with teeth: on Django 7.0 a bare `", "` is resolved as a field reference rather than a string literal, which will break this query.

Being explicit that **this is not a behaviour fix** — `ordering` was already being silently discarded into `Func.extra`, but the emitted SQL is unchanged either way, because `STRING_AGG(DISTINCT ...)` already returns sorted output as a side effect of deduplication:

```sql
SELECT string_agg(DISTINCT v, ', ') FROM (VALUES ('Shelter Operator'),('Caseworker'),('Zebra'),('Alpha')) t(v);
--  Alpha, Caseworker, Shelter Operator, Zebra
```

Two existing tests select `permissionTemplates` but never assert its value, so the field had no value coverage. Added one that pins the delimiter contract `UserType.permission_templates` relies on when it does `raw.split(", ")`.

## 2. django-stubs 6.1

The stubs were pinned `>=5.2.9,<6` — targeting Django **5.2**, two majors behind. That is why the new import needed a `type: ignore`, and it would have recurred on every Django 6.x API we adopt.

Also reconciles a three-way version drift: the lock pinned Django 6.0.7 while local environments had installed 6.1. Lock now moves to Django 6.1 and DRF 3.18.

### Three live bugs the old stubs were hiding

**`LocationAdmin` crashed on every Location change form.** Its readonly `notes` and `tasks` columns called `format_html("<br>".join(...))`. `format_html` requires format arguments and raises `TypeError` without them. All three call sites (including `attachments`) now use `format_html_join` — which also stops row contents being interpolated into HTML unescaped.

**`LocationTaskAdminMixin` looked up the wrong model.** `apps.get_model("notes", "Task")` — `Task` lives in `tasks`, and has no `location` field. It reaches a location through its note, so the lookup is now `note__location`. The reverse URL was `admin:notes_task_change` too.

**Shelter CSV import broke on contacts.** `ShelterResource.after_save_instance` passed `email=` and `title=` to `ContactInfo`, whose fields are `contact_email` and `contact_title`. Any import with `additional_contacts` raised.

Each is covered by a regression test that fails when the fix is reverted.

### Also

`HmisProfile.display_gender` appended `gender_identity_text` unconditionally even though it's nullable, putting a `None` into a `list[str]`. Now only included when set — worth a look, since it changes what the field returns.

The remaining 16 errors were `admin_interface` missing stubs (now in `mypy.ini`'s ignore list alongside the other untyped libraries), plus `.values()` row typing, unannotated pghistory `tracks()` calls, and `Optional` narrowing.

## 3. No `cast`, no `type: ignore`

The second commit bought mypy silence with four `cast(list[dict[str, Any]], ...)` calls in `reports/selectors.py` and two `# type: ignore[no-untyped-call]` comments in `shelters/admin.py`. Neither is necessary.

**The casts were hiding a stale annotation, not a stubs gap.** django-stubs 6.1 infers a precise anonymous TypedDict for `.values()` rows instead of `dict[str, Any]`:

```
QuerySet[Note@AnnotatedWith[TypedDict({'trunc_date': date, 'count': int})],
         TypedDict({'date': Any, 'count': Any})]
```

A TypedDict is not assignable to `dict[str, Any]`, so `list(rows)` fails `arg-type`. The stubs are right. Building the row dicts in the comprehension — the shape `note_count_by_team` was already using, uncast, in the same file — makes the declared return type true rather than asserted, and lets a rename-only trailing `.values()` go from three of the queries. Verified against the query compiler that the SQL is unchanged apart from the column alias.

**pghistory ships `py.typed` but leaves `EventQuerySet.tracks` unannotated**, which is what `disallow_untyped_calls` was tripping on. `untyped_calls_exclude = pghistory` is mypy's option for exactly that case. Worth being clear that this buys no safety either way — `tracks` is `def (*objs: Any) -> Any`, so the chain after it is unchecked regardless; the argument for config over comments is that `warn_unused_ignores` is off here, so the comments would never be reported as stale if pghistory ever annotates it.

**`topProvidedServices` had no key-name coverage.** The existing GraphQL test creates notes with no services, so the list comes back empty and a renamed key never reaches `NameCountType(**row)`. Added a test that gives it real services and asserts the labels and counts; it fails when the key is renamed.

`[mypy-admin_interface.*] ignore_missing_imports` stays. It is not a workaround for anything here — the django-stubs plugin pulls every `INSTALLED_APPS` models module into the build, `django-admin-interface` has no `py.typed` and no published stubs, and removing the entry puts 16 `import-untyped` errors back across `*/apps.py`, `common/admin.py` and others.

## Verification

- `pytest` — 1528 passed, 31 skipped, on Django 6.1
- `mypy` — clean across 390 files
- `ruff check` / `ruff format` — clean

## Note

This is the first of a series clearing the backend suite's 1185 warnings. Next up is the `Attachment.save()` deprecation, which is ~1050 of them on its own.

## Summary by Sourcery

Modernize Django aggregation and typing dependencies while fixing admin, import, reporting, and nullable-data issues uncovered by the upgrade.

Bug Fixes:
- Fix admin rendering for location notes, tasks, and attachments, including task relationships and safe HTML output.
- Fix shelter CSV contact imports to persist email and title fields correctly.
- Prevent nullable gender identity values from appearing in HMIS gender output.
- Handle empty attachment filenames and nullable report data safely.

Enhancements:
- Migrate permission-template aggregation to Django's current StringAgg API and add coverage for delimiter handling and template ordering.
- Update Django and Django stubs dependencies to the 6.1 release line and align Django REST Framework versions.
- Simplify report aggregation result handling and remove redundant type casts.

Build:
- Upgrade the project runtime and development dependency constraints for Django, Django stubs, and Django REST Framework.

Tests:
- Add regression coverage for permission-template aggregation, location admin columns, attachment links, shelter contact imports, and report service summaries.

Chores:
- Adjust mypy configuration for untyped third-party libraries and updated type-checking APIs.


apps/betterangels-backend/common/tests/test_models.py
```diff
@@ -461,3 +468,60 @@ def test_save_direct_upload_requires_mime_type(self) -> None:
 
         with self.assertRaises(ValueError):
             attachment.save(direct_upload=True)
+
+
+class LocationAdminMixinTestCase(TestCase):
```
mikefeldberg, line 473: was this just a drive-by update?

apps/betterangels-backend/common/admin.py
```diff
@@ -19,44 +19,38 @@ def attachments(self, obj: Model) -> SafeString:
...
+        return format_html_join(
+            mark_safe("<br>"),
+            '<a href="{}">Note {}: {}</a>',
+            ((reverse("admin:notes_note_change", args=(note.id,)), note.id, note) for note in notes),
+        )
 
 
 class LocationTaskAdminMixin:
     def tasks(self, obj: Location) -> SafeString:
-        Task = apps.get_model("notes", "Task")
```
mikefeldberg, line 49: nice one

mikefeldberg (approved):

PR #2407, opened 2026-09-01
global perm template org-bypass (SDB-218)

mikefeldberg:
SDB-218

## Summary by Sourcery

Introduce permission-scoped organization bypass roles with tightly controlled administration and cross-organization shelter management.

New Features:
- Add permission-template support for organization-scoped bypass roles, including cross-organization shelter operations for the Global Shelter Operator.

Bug Fixes:
- Prevent org-bypassing roles from being granted through member management, invitations, organization creation, or non-admin role APIs.
- Apply permission-aware authorization consistently across shelter, room, bed, reservation, and shelter-photo operations.

Enhancements:
- Make organization bypass access permission-scoped and preserve organization boundaries for membership-only and empty-permission queries.
- Allow authorized bypass users to target a specific organization when creating shelters while rejecting cross-organization targets for ordinary users.
- Synchronize managed templates’ organization-bypass flags from code configuration while preserving admin-defined template values.

Tests:
- Add coverage for bypass permissions, role-assignment restrictions, cross-organization shelter operations, and regression behavior for ordinary operators.


vecchp:
## Defects

**`shelters/groups.py:86-89` — the TODO is inverted.** `strawberry_django/utils/query.py:145-155` ORs an uncorrelated `Exists(user.groups.filter(permissions__codename=…))` into every `filter_for_user`. Generated SQL for `HasRetvalPerm(ClientProfile.perms.VIEW)`:

```sql
OR EXISTS(SELECT 1 FROM "auth_group" "U0"
  JOIN "accounts_user_groups" "U1" ON ("U0"."id" = "U1"."group_id")
  JOIN "auth_group_permissions" "U3" ON ("U0"."id" = "U3"."group_id")
  JOIN "auth_permission" "U4" ON ("U3"."permission_id" = "U4"."id")
  WHERE ("U1"."user_id" = 1 AND "U4"."codename" = view_clientprofile) LIMIT 1)
```

No reference to `clients_clientprofile`. `PermissionGroup` subclasses `auth.Group`, so an org role satisfies it. Every Shelter Operator reads every ClientProfile, Note, ClientDocument, Task and Referral in every org — on `main`, list path and by-id. `filter_with_perms` sets `is_safe`, so `_resolve_obj` skips the per-object check too.

**`common/permissions/utils.py:394` — every user loses the org index.**

```
main:  Index Scan using shelters_shelter_organization_id_d9064f34
       Index Cond: (organization_id = 1)

this:  Seq Scan on shelters_shelter
       Filter: ((organization_id = 1) OR (InitPlan 1).col1)
```

Uncorrelated `EXISTS` becomes an `InitPlan`; the `OR` defeats the index for all callers.

**`accounts/extensions.py:88` — nonexistent org passes.** `Organization.objects.filter((pk=<header> AND perm) | bypass).exists()` with the bypass term uncorrelated → true whenever any Organization row exists. `X-Organization-ID: 99999999` passes every gate; `get_current_organization` then returns `"99999999"` to services.

**`accounts/models.py:141` — `bypasses_org_scoping` is admin-editable.** `accounts/admin.py:117-130` makes `("name", "permissions")` readonly on code-owned templates. The flag is the only writable field. `accounts.change_permissiongrouptemplate` + tick it on `Caseworker` = cross-org read/write for every Caseworker in every org until the next `migrate` reverts it.

**`accounts/role_manager.py:49-61` — both guards unreachable.** `PermissionTemplateEnum` and `get_template_or_raise` are built from `invitable_template_names()`; GSO is `is_invitable=False`. `create_organization_with_presets` is the only other caller and GSO has no `PermissionGroup` to resolve. The DB-row guard's unselected `template` FK access is the `assertNumQueriesWithoutCache` 19→20 bump — one query per role on every member add.

**`shelters/schema.py:156` — 500 on a non-numeric id.** `Organization.objects.filter(pk="abc")` raises `ValueError`; `strawberry_django/mutations/fields.py:99-105` converts only `ValidationError | PermissionDenied | ObjectDoesNotExist`. Response leaks `Field 'id' expected a number but got 'abc'`. `accounts/selectors.py:31` swallows the same error.

**`shelters/services/bed.py:130`, `room.py:156`, `reservation.py:105-110` — gate/selector mismatch.** `cloneBed`/`cloneRoom` gated ADD, scoped VIEW. `createReservation` gated `Reservation.perms.ADD`, resolves with `Bed.perms.VIEW`/`Room.perms.VIEW`.

**`shelters/tests/test_global_shelter_operator.py`** — 9 tests pass with the feature reverted. 3 negative assertions vacuous (no GSO `PermissionGroup` exists for those orgs either way). Docstring cites `docs/global-shelter-operator-org-bypass.md`, not in the repo.

**Stale artifacts.** `schema.graphql` plus 4 frontend generated files. CI gates only the backend SDL — `nx affected -t generate-graphql-types` cannot select the seven frontend codegen projects from a backend change (no `implicitDependencies`, `sharedGlobals` omits the schema).

**Import reordering, ~9 files.** Root `pyproject.toml` selects `E,F,B` only; isort rules are off and `ruff check`/`format --check` pass either way. `shelters/schema.py:69-77` splits one `shelters.selectors` import into two.

## Blocked on the frontend

`accounts/types.py:100-127` — `CurrentUserOrganizationType.get_queryset` is `queryset.filter(users=user)`; `permissions` aggregates `permission_groups__permissions` filtered `permission_groups__user=user`. A holder with no `OrganizationUser` row contributes to neither → `useActiveOrgState` commits no active org → `libs/fetch/src/lib/fetchInterceptors.ts:135` sends no `X-Organization-ID`.

| file | result |
|---|---|
| `apps/betterangels-admin/src/app/pages/home.tsx:17` | permanent `Loading...` |
| `apps/betterangels-admin/src/app/routes/PermissionGuard.tsx:13` | every guarded route redirects |
| `libs/react/shelter-operator/.../Dashboard.tsx:97` | `skip: !selectedOrganizationId` |
| `.../CreateShelterProfile.tsx:61-63` | `if (!activeOrg?.id) return;` — silent no-op |

Needs `CurrentUserType.globalPermissions`, an assignable-organizations query, and those four call sites.

## Design

**`GlobalRole(Group)`** — MTI, no `organization` field. Not `permissiongroup__isnull=True`: `accounts/migrations/0003:59-63` discards a `get_or_create`d Group inside `defaults={"group": new_auth_group, …}`, `0005:66-70` prints the orphan inventory, and `auth.Group` uses the stock unrestricted `GroupAdmin` — absence-based makes `auth.add_group` a network grant.

**`common/permissions/grants.py`**

```python
@dataclass(frozen=True)
class Grant:
    network: bool = False
    organizations: frozenset[str] = frozenset()   # str(pk)
    def covers(self, organization) -> bool: ...

def grants_for(user) -> Mapping[str, Grant]       # 2 queries, memoized on user._ba_grants
```

```python
Group.objects.filter(user=user).values_list(
    "permissions__content_type__app_label", "permissions__codename",
    "permissiongroup__organization_id", "globalrole__group_ptr_id")
```

One statement, two LEFT JOINs, no row multiplication. `globalrole__group_ptr_id` non-null → network; `permissiongroup__organization_id` non-null → that org; both null → nothing. Keys stay `app_label.codename` — `accounts/seed.py:117-152` derives ContentTypes from `codename.rsplit("_", 1)[-1]`, producing duplicate `(app_label, codename)` rows and synthetic types (`organizations.portal`, `reports.reports`), so ContentType-exact matching silently denies Org Admin.

**`common/permissions/scopes.py`** — the model declares its tenant paths, replacing the `organization_field`/`organization_fields` kwargs and the four `*_queryset` wrappers.

```python
class ScopeSet:
    ORGANIZATION: tuple[str, ...] = ()
    OBJECT: bool = False
    UNSCOPED_REASON: str = ""

class Reservation(BaseModel):
    class scopes(ScopeSet):
        ORGANIZATION = ("bed__shelter__organization_id", "room__shelter__organization_id")

class Note(BaseModel):
    class scopes(ScopeSet):
        ORGANIZATION = ("organization_id",)
        OBJECT = True

class ClientProfile(BaseModel):
    class scopes(ScopeSet):
        UNSCOPED_REASON = "No organization FK and no guardian rows: shared network index."
```

Ordered precedence, first match wins:

1. `grant.network` → all rows
2. `ORGANIZATION` ∧ `grant.organizations` → `reduce(or_, Q(**{f"{p}__in": ids}))`
3. `UNSCOPED_REASON` ∧ `bool(grant)` → all rows
4. `OBJECT` → `Q(pk__in=get_objects_for_user(..., accept_global_perms=False))`
5. → `.none()`

Never `ORGANIZATION | OBJECT`. With `enable_seqscan=off`:

```
organization_id IN (1,2,3)                          → Bitmap Index Scan on notes_note_organization_id_15f23c82
... OR id IN (<guardian subquery>)                  → Seq Scan  (cost=10000000017.33..)
```

Postgres takes the seq scan under a 10⁷ penalty. The union is also empty: `notes.view_note` and `tasks.view_task` are never granted at object level (`notes/services.py:275-281` grants CHANGE+DELETE+PrivateDetails.VIEW; `tasks/services.py:52-57` CHANGE+DELETE), and `assign_object_permissions` always targets a `PermissionGroup` of the row's own org, so rung 2 is a superset of rung 4.

**`common/permissions/checks.py`** — `manage.py check`, not docstrings.

- `E002` — declared path unresolvable **or multi-valued**. `User` reaches Organization only by reverse M2M; `filter(organizations_organization__id__in=…)` compiles to an INNER JOIN that duplicates rows (the bug recorded at `notes/types.py:336-348`).
- `E005` — `Group not in PermissionGroup._meta.parents` or `not in GlobalRole._meta.parents`. Reverting either MTI silently reclassifies every org role as network.

**`common/middleware/organization.py`** — `request.organization = SimpleLazyObject(...)`; `None` when absent, `PermissionDenied` when unresolvable, no `int()` coercion. `PermissionDenied` not `BadRequest`: `teams/tests/test_queries.py:139` asserts a GraphQL error on a 200, and Apollo reads a 400 as a network `ServerError`. `active_organization(info)` replaces all 43 `get_current_organization` uses; the same validator serves `CreateShelterInput.organization_id`.

**GraphQL seam** — `RequiresPerm(perm, scope=ANY|ACTIVE|ROW)` subclassing `DjangoPermissionExtension`, plus a `ScopedType.get_queryset` mixin. `get_queryset_hook` applies the type hook on all four field shapes (connection, list, `qs.first()`, `qs.get()`) and on the optimizer's prefetch path; `filter_with_perms` covers neither. Drops every undocumented import (`PermTarget`, `PermDefinition`, `resolve_for_user_with_perms`, `filter_for_user`, `UserType`) — relevant under the `>=0.86.1,<1` pin.

Two exclusions: `ShelterType` keeps delegating to `shelter_list` (anonymous public directory; `grant_for(AnonymousUser, …)` is falsy → `.none()`, pinned by `PublicShelterQueryTestCase`), and `_room_beds_prefetch` (`shelters/types/outputs.py:299-306`) keeps an explicit `scoped()` call because `run_type_get_queryset` builds from `_base_manager`.

`ShelterFilter.organizations` (`shelters/types/filters.py:123-137`) authorizes on `current_user.organizations_organization.all()` — `OrganizationUser` membership, permission-blind — and ANDs after the type hook, so it must be fixed in the same change. Its unauthenticated branch (127-130) is looser than its authenticated one.

**`user.has_perm`** — do not narrow `get_group_permissions` here. Measured on a migrated database: a staff user holding only org roles goes 44 permissions → 0, `has_module_perms` True→False on shelters/clients/notes/tasks, usable ModelAdmins 12/50 → 0/50, while `AdminSite.has_permission` still admits them. Additive half is safe now: `ModelBackend._get_permissions` returns `set()` when `obj is not None` (`django/contrib/auth/backends.py:115-116`), so a backend implementing `has_perm(user, perm, obj)` fixes `HasRetvalPerm`'s single path, `HasSourcePerm` and `ModelAdmin.has_change_permission` with no loss of access. Narrowing goes last, on both `ModelBackend` and allauth's backend (shared `_perm_cache` attribute names), after global grants are issued to affected staff.

**Net** ≈ −790 production lines deleted (`permissioned_queryset`, `_org_perm_exists_across_fields`, `get_current_organization`, `perm_filter`, `accounts/extensions.py`, `common/graphql/extensions.py`, `get_object_or_permission_error` +12 sites, `get_user_permitted_org` +3, the four wrappers, `user_shelter_list`/`operator_shelter_list`, 14 `filter_for_user` sites, `REGISTRY.unscoped`) against ≈ +475. Ten mechanisms → three.

## Stages

| | change | notes |
|---|---|---|
| S0 | production audit, no code | orphan Groups with perms; `user_permissions` rows; `is_staff` users whose only groups are `PermissionGroup`s; null-`organization_id` counts on Shelter/Referral/Task; duplicate template names. Blocks S2, S6, S9 |
| S1 | `request.organization`, `active_organization` | ~45 sites. A nonexistent-but-numeric header stops passing |
| S2 | `grants.py`, `scopes.py`, `checks.py`, `GlobalRole`, shelters domain | ~40 sites. No global role yet, so no access changes. Verify by SQL equality against a hand-built org-only `Q` + `EXPLAIN`. `assertNotIn("EXISTS", …)` is insufficient — Django renders `pk__in` as `IN (SELECT …)` |
| S3 | `RequiresPerm` + `ScopedType`, 33 `HasOrgPerm` conversions | delete `accounts/extensions.py`, `get_user_permitted_org`; keep `?org_id=` as deprecated alias (`useCSVDownload.ts:16`). Regenerate all 5 artifacts |
| S4 | frontend half | ships before S5; mobile goes through app stores |
| S5 | `GLOBAL_ROLES` outside `REGISTRY`, `seed_global_roles()`, data migration, `organizationId` | exclusion from invitations becomes structural — `PermissionTemplateEnum` is built from `invitable_template_names()` |
| S6 | outreach domain: 26 `HasPerm` + 32 `HasRetvalPerm`, delete `common/graphql/extensions.py` and the 14 `filter_for_user` sites | closes the live leak. Requires note/task/referral selectors that don't exist yet. Re-derive `assertNumQueries`, don't bump |
| S7 | object-aware backend | additive |
| S8 | validated org at every `resolve_permission_group` call site (~10) | closes two in-tree TODOs |
| S9 | narrowing mixin, `name unique=True`, admin scoping | audit-gated; only stage that can lock a human out |

Data migration in S5: move users from any `Global Shelter Operator` `PermissionGroup` into the `GlobalRole`, delete those groups, delete the template row. The last step matters — once the name leaves `REGISTRY.template_names()`, `accounts/admin.py:117-118` reclassifies the row as hand-defined and `sync_group_permissions` treats its 126 permissions as the definition.

## Land now, separately

`perms=` / `permission=` threading through the shelter services, and the `baker_recipes.py` website fix (`shelter_update` calls `full_clean()`; `URLField` rejects a schemeless value). Both are S2 prerequisites. `test_bypass_never_grants_permissions_not_on_the_template` should survive.

vecchp:
# Critique

The permission system is already tiered, and `strawberry_django.utils.query.filter_for_user_q`
is the shape of a tier — a term in one predicate asserting *this permission is held at this
level*:

```python
q = Q()
q |= Q(Exists(user.user_permissions ...))      # direct user permissions
q |= Q(Exists(user.groups ... permissions))    # group permissions
q |= Q(pk__in=<guardian object rows>)          # object permissions
return qs & qs.filter(q)
```

This PR doesn't add a tier. It adds a boolean on the role that means *ignore the level*.
That's an exception rather than a member, and three things follow from it directly.

It has to be implemented once per check path, and there are three: `_bypass_perm_exists`
for the queryset, `user_holds_org_bypass_perms` for resolvers, and `get_user_permitted_org`
— which deliberately doesn't honour it, with a 16-line docstring standing in for the
invariant. `CreateShelterInput.organization_id` and the two `add_roles` guards are the same
instruction reappearing at two more sites.

It's role-wide rather than per-permission, so all ~200 permissions on the template become
cross-org together. There's no way to say cross-org for shelters and org-scoped for
everything else — which is what the TODO at `shelters/groups.py:86-89` is reaching for in
prose.

And it can't compose with the existing terms, because it isn't one. A tier would be another
`Q` OR'd into the same predicate. A bypass is a second predicate wrapped around the first,
which is why `HasOrgPerm` ends up as `(pk = <header> AND perm) OR <uncorrelated exists>` —
true whenever any Organization row exists, so a holder passes with any header value
including a nonexistent org id, which `get_current_organization` then hands to services.

# Target state, scoped to shelters

No new model and no flag — the tier is implied by whether the role has an organization.
`PermissionGroup.organization` becomes nullable: set means org-scoped, unset means global.

That keeps a global role inside the machinery that already exists — seeded from its
`TemplateConfig`, permissions maintained by `sync_group_permissions`, same admin — which a
code-owned ~200-permission role wants. It also means `PermissionGroup.objects.filter(organization=…)`
excludes global roles for free, so `reconcile_org_groups`, the admin inlines and
`OrgRoleManager` structurally can't grant or reconcile one; the PR's two `add_roles` guards
become unnecessary rather than enforced.

Two small consequences. `organization` is non-null today, so this is an `AlterField`, and
`unique_together = (("organization", "template"))` stops protecting global rows since
Postgres treats NULLs as distinct — it needs
`UniqueConstraint(fields=["template"], condition=Q(organization__isnull=True))` or
`nulls_distinct=False`. And `group_name()` interpolates `self.organization_id` and reads
`self.organization.name`, so it needs a branch for the global case.

The missing tier is the organization. `filter_for_user_q` has no term for "held in these
organizations", so adding one extends its shape rather than wrapping it:

```python
q = Q()
q |= Q(Exists(<held with no organization>))        # global: user perms, or a role with org unset
q |= Q(**{f"{org_field}__in": orgs_with_perm})     # organization
q |= Q(pk__in=object_perm_rows)                    # object
```

One query, one term per tier, the same OR'd predicate the library already builds. Adding a
tier is additive by construction, and no call site learns that a global tier exists. The
test for the first term is a single column — org-scoped iff held through a `PermissionGroup`
with an organization, global otherwise.

Two details the library settles for us. The global term is row-invariant, so leaving it in
the predicate puts a constant inside a row filter — the same shape as the bypass.
`filter_for_user_q` already handles that: lines 126-131 short-circuit and return the
queryset whole when the permission is in Django's `user._perm_cache`. We do it
deterministically rather than cache-dependently — hoist the global tier, keep the
row-varying tiers in the Q.

And the global test reads `user_permissions` and org-less roles directly rather than calling
`user.has_perm`, because an org role's permissions are currently also group permissions, so
`has_perm` answers True for any Caseworker. Caching follows Django's convention — memoized
on the user instance the way `ModelBackend` populates `_perm_cache` / `_group_perm_cache` —
rather than a new registry.

The organization term needs to know how each model reaches an organization. That's declared
on the model, by inheriting a base that marks participation and carries one overridable
attribute — the same way `BaseModel` already declares a capability with
`class perms(PermissionSet)`:

```python
class OrgScoped(models.Model):
    """Participates in organization scoping. Empty org_via = my own organization FK."""
    org_via: ClassVar[tuple[str, ...]] = ()

    class Meta:
        abstract = True
```

```python
class Shelter(OrgScoped, BaseModel):     ...                      # owns organization
class Room(OrgScoped, BaseModel):        org_via = ("shelter",)
class Bed(OrgScoped, BaseModel):         org_via = ("shelter",)
class Reservation(OrgScoped, BaseModel): org_via = ("bed", "room")
```

This replaces `organization_field="shelter__organization_id"` and Reservation's two-path
lookup — the same knowledge moved from the call site onto the model, so it's a refactor
rather than a schema change. `org_via` names *relations*, not lookup paths, so the base
resolves and validates each hop from `_meta` rather than trusting a hand-written string, and
`manage.py check` can enumerate subclasses from the app registry to flag a permissioned
model that scopes nothing.

The base deliberately contributes no field. Django adds abstract-base fields
unconditionally, so a base that sometimes supplies `organization` and sometimes doesn't
would need two bases or metaclass work; `Shelter` already has the FK.

Having the base add an `organization` column to every model instead would make the term a
single column, but it's the wrong trade here: it needs a migration and backfill on live
tables plus a composite FK (`UNIQUE (id, organization_id)` on the parent) purely to stop the
copy drifting — a problem traversal doesn't have, since there's one source of truth. It also
doesn't generalize, because `Reservation.bed` and `.room` are both nullable, so there's no
parent to enforce against; traversal lets a reservation with neither simply have no
organization. The parent FKs are all indexed already (`shelters_bed_shelter_id_af909cf8`,
`shelters_room_shelter_id_51a61c33`), so the hop is an indexed FK→PK lookup. Worth
revisiting for sharding, where a tenant column on every table is a requirement.

Each hop must be single-valued — a reverse-FK or M2M hop makes the filter duplicate rows,
which is the bug recorded at `notes/types.py:336-348`. That's what the `check` validates.

Photos, schedules and contact info stay transitively scoped through the shelter queryset.

There's no object term for shelters, since the app grants no object-level permissions. When
a domain does need one, extend `filter_for_user_q`'s existing term rather than calling
`guardian.shortcuts.get_objects_for_user`: it resolves the object-permission models per
target model (handling both the generic and direct guardian layouts), and keeps everything
in SQL via `Cast("object_pk", model._meta.pk)` plus a `union()` of the user-object and
group-object querysets. The shortcut materializes a pk list in Python and defaults
`accept_global_perms=True`, which would put the leak back.

# Getting there

Dual path. A new extension delegating to the same check, applied only to the shelter
endpoints:

```python
class RequiresPerm(DjangoPermissionExtension):
    """Field gate. Delegates to the tiered check; the field states only the permission."""
```

`HasOrgPerm` and `PermissionedQuerySet` stay untouched everywhere else and migrate when
their domain needs it.

`createShelter` takes the target organization as an input validated by the same helper the
header uses — which also removes the current 500: `Organization.objects.filter(pk="abc")`
raises `ValueError`, and `strawberry_django/mutations/fields.py:99-105` only converts
`ValidationError | PermissionDenied | ObjectDoesNotExist`, so `organizationId: "abc"` leaks
`Field 'id' expected a number`.

The frontend goes first, or none of it is reachable. `currentUser.organizations` and its
`permissions` are both derived from membership (`accounts/types.py:100-127`), so a role
granting access without membership yields an empty org list, no `X-Organization-ID`, and
`hasPermission` false everywhere:

| file | result |
|---|---|
| `apps/betterangels-admin/.../home.tsx:17` | permanent `Loading...` |
| `apps/betterangels-admin/.../PermissionGuard.tsx:13` | guarded routes redirect |
| `libs/react/shelter-operator/.../Dashboard.tsx:97` | `skip: !selectedOrganizationId` |
| `.../CreateShelterProfile.tsx:61-63` | `if (!activeOrg?.id) return;` — silent no-op |

# What this gets us

Shelter data-entry staff hold shelter permissions at the global tier — a real level in the
permission system, not an exemption from it. The shelter endpoints resolve tiers through one
check; every other endpoint keeps the check it has and gains the organization tier when its
domain needs it. Because a tier is per-permission, a role can be global for shelters and
org-scoped for anything else, which is the distinction the flag can't express.

It also means the next cross-org role costs a row rather than a design: create a
`PermissionGroup` with no organization and point it at a template. And the tier terms are the place the remaining gaps get fixed — the
group term in `filter_for_user_q` currently satisfies any org role, which is why reads
aren't org-confined for `HasPerm`/`HasRetvalPerm` today and why the `ClientProfile` TODO
reads backwards. Same predicate, one term to correct, whenever that domain comes up.

Two things worth fixing on this PR regardless of shape: that TODO, and the fact that
`bypasses_org_scoping` would be the only editable field on every code-owned template
(`accounts/admin.py:117-130` makes `("name", "permissions")` readonly), so
`accounts.change_permissiongrouptemplate` plus a tick on `Caseworker` grants cross-org
access until the next `migrate` reverts it.

vecchp:
# Why the flag keeps recurring (1/4)

The flag has to exist because there is no level to put a global role at. That is a property of `PermissionGroup`, not something this PR introduced, and it is worth tracing because the same property produces most of the other awkwardness in the permission system.

`PermissionGroup` is an `auth.Group` subclass, unique on `(organization, template)`. Holding a role is `user.groups.add(...)`. That is genuinely economical — it makes `user.has_perm` work with no code, and it makes guardian's `GroupObjectPermission` rows cascade correctly, which is what #2378 was for.

It has one property that everything else follows from: **`auth.Group.permissions` is unconditionally global.** `ModelBackend.get_all_permissions` unions `user.groups__permissions` with no notion of scope, and the backend signature — `has_perm(user, perm, obj=None)` — gives it nowhere to learn one even if it wanted to. So the moment `sync_group_permissions` writes `notes.view_note` onto the SELAH-Caseworker group:

- `user.has_perm("notes.view_note")` is `True` everywhere, for a caseworker at any org.
- `strawberry_django.utils.query.filter_for_user_q` ORs in `Exists(user.groups…permissions…)`. That subquery does not reference the outer row, so it is row-invariant: it **matches every row in `notes_note`**. Object permissions stop being a row filter at all.

That is not a bug in the group term — it is the group term working correctly for a tier that has no scope. The error is upstream.

Five things follow, and all five are in the codebase today:

**Permissions had to be withheld from roles.** `CASEWORKER` (`notes/groups.py:12`) carries `Note.perms.ADD` and `VIEW` but deliberately not `CHANGE` or `DELETE`, because a model-level CHANGE would let any caseworker edit any note in the system. So CHANGE/DELETE were pushed onto per-object guardian rows written at creation (`notes/services.py:176`). django-guardian is load-bearing for a job it was never needed for: "the creating organization may edit this" is already carried by `Note.organization`.

**A parallel filter had to be built.** Since `has_perm` could not be trusted, `permissioned_queryset` re-derives the answer by joining `Organization → PermissionGroup → Permission` and ignores the Django path entirely. There are now two answers to "does this user hold `notes.change_note`", and they disagree.

**Seven checking mechanisms.** `HasPerm` (24 sites) asks the leaky question, `HasRetvalPerm` (29) asks guardian, `HasOrgPerm` (36) asks the join, `PermissionedQuerySet` (11) asks guardian and then deliberately does not enforce, `IsAuthenticated` (105) asks nothing, plus one DRF class and two raw `has_perm`. You cannot tell from a resolver which question is being asked.

**The organization had to come from somewhere.** The join needs an org id, so the `X-Organization-ID` header appeared. Where a mutation has no header plumbing, `resolve_permission_group` picks *the first organization where the user holds the template* (`accounts/selectors.py:105`) — 11 call sites, three `TODO(org-scoping)` comments, and non-deterministic attribution of new notes for any multi-org caseworker.

**Cross-org is structurally impossible and "global" is inexpressible.** The join clause `permission_groups__user=user` proves membership *and* role-holding at once, so a non-member matches nothing. And `PermissionGroup.organization` is non-null, so a role without an organization cannot exist — which is why `GLOBAL_SHELTER_OPERATOR` is a 126-permission template pinned to one arbitrary organization with `reconcile_org_groups` taught to skip it, and why this PR needs a boolean that makes the join skip itself.

So the rule the design has to follow is:

> **The role must be organization-independent. The grant must carry the organization.**

Once that holds, every consequence above unwinds. Roles can carry CHANGE/DELETE, so guardian becomes unnecessary. There is one filter, so there is one mechanism. The organization is an explicit argument, so nothing is guessed. And a grant whose principal sits outside the scope is just a row — which is what SDB-218 needs.

The next three comments work that out: the model, the resolver and the surfaces, and the path. The short version is that SDB-218 becomes one row in one table, so `bypasses_org_scoping` is not needed and this PR can close once the direction is agreed. Two things on it are worth fixing regardless of shape, and comment 4 comes back to them.

vecchp:
# The target model (2/4)

## Roles are plain `auth.Group`, one row per role, no organization

```python
Group(name="Caseworker")                      # permissions = the bundle
Group(name="Organization Admin")
Group(name="Shelter Operator")
Group(name="Shelter Reference Data Admin")
```

`Group.permissions` is precisely the M2M needed, with an admin UI already built. A new `Role` model would need its own M2M, its own admin and its own migration, and would gain nothing.

More importantly, keeping `Group` means **`user.groups` becomes a real tier rather than a leak**: adding a user to a role's group grants that role *globally*, read by `ModelBackend`, the admin, `PermissionRequiredMixin` and templates. The property that broke the old model is the one we now use deliberately.

That is what dissolves SDB-218. `SHELTER_OPERATOR` is 18 permissions; `GLOBAL_SHELTER_OPERATOR` is 126 and contains **all 18 of them**, because "global" had to be baked into the role. In the target model there is one `Shelter Operator` role and one `Shelter Reference Data Admin` role for the lookup tables. BA data-entry staff are added to both groups (global); a partner org's staff get a `Grant` for the first. The duplication was an artefact of the role carrying its own scope.

Making the role organization-independent also deletes the machinery that exists only to materialise it per org:

| | today | target |
|---|---|---|
| role definitions | 5 `PermissionGroupTemplate` | — |
| materialised roles | 5 × N_orgs `PermissionGroup`, and the same in `auth_group` | 5 `auth_group` |
| permission rows | 5 × N_orgs × \|perms\| in `auth_group_permissions`; the five templates total 190 permissions, so ~190 rows *per organization* | 190 rows, once |
| provisioning | `post_migrate` → `reconcile_org_groups` + `sync_group_permissions` + `group_name()` | one `get_or_create` + `permissions.set` per role |
| membership | `auth_user_groups` | `accounts_grant`, same cardinality |

Nothing is provisioned per organization any more. Which roles an org may *grant* stays a validation question in the invite/assign service reading `OrganizationProfile.org_types` — a rule about who may be given what, not a reason to create rows in advance.

## `Grant` — the scoped binding

```python
class Grant(models.Model):
    """Who holds which role, in which organization. The only scoped authorization input."""

    # ── principal: exactly one ──
    user = models.ForeignKey(
        User, null=True, blank=True, on_delete=models.CASCADE, related_name="grants")
    organization = models.ForeignKey(
        Organization, null=True, blank=True, on_delete=models.CASCADE,
        related_name="delegated_grants")

    # ── what ──
    role = models.ForeignKey(Group, on_delete=models.CASCADE, related_name="grants")

    # ── where ──
    scope = models.ForeignKey(Organization, on_delete=models.CASCADE, related_name="grants")

    class Meta:
        constraints = [
            models.CheckConstraint(
                name="grant_has_exactly_one_principal",
                condition=Q(user__isnull=False) ^ Q(organization__isnull=False)),
            models.UniqueConstraint(fields=["user", "role", "scope"], name="unique_user_grant"),
            models.UniqueConstraint(
                fields=["organization", "role", "scope"], name="unique_org_grant"),
            models.CheckConstraint(
                name="grant_org_principal_is_not_scope",
                condition=Q(organization__isnull=True) | ~Q(organization=F("scope"))),
        ]
```

Wrapped in `@pghistory.track` like the rest of the domain models, which is the audit trail for "who let this partner in" — so no `granted_by` column.

**`user` / `organization`, XOR'd.** An organization as principal means *everyone who holds anything at that organization*. Two nullable FKs rather than a generic FK to a principal: only two principal kinds are asked for, a GFK costs referential integrity and turns both arms of the predicate into `(type, id)` comparisons instead of indexed FK lookups, and the XOR check makes "exactly one" a database fact rather than a convention. Nullability works in our favour here — Postgres treats NULLs as distinct in a unique index, so the two uniqueness constraints do not collide with each other and no partial indexes are needed.

Reusing `Group.user_set` as the membership instead — `Grant(group, role, scope)` — does not work, because a group's membership is global, so you would need a distinct group per organization. That is the materialisation being removed.

**`role` as an FK, not a permissions M2M on the grant.** Roles are the compression; a grant of 40 permissions would duplicate the set per (user, organization). The trade-off worth naming: granting one person exactly `reports.view_reports` at one org means creating a role rather than ticking a box. That is arguably better — the access is named, reusable and auditable rather than bespoke.

**`scope` is non-null, and this is where the design differs from the earlier comment on this PR.** That comment made `PermissionGroup.organization` nullable, with unset meaning global. That was right *given its premise*: `PermissionGroup` was the only place a role could live, so global had to fit inside it.

Once roles are plain groups, global already has a home, and a better one — `user.groups` and `user.user_permissions`, which `ModelBackend`, the admin, templates and every third-party package already read. Creating a global role is `Group.objects.create(...)` plus permissions; granting it is the group picker on the User admin page. No new column, no new UI, no new code. A nullable `scope` would therefore create a *second* way to say "global": `scopes()` would have to check both, the admin would have two places to look, and they could disagree. Non-null keeps `Grant` unambiguous — every row means scoped authority.

**Growth.** `Grant` grows as users × organizations they act in × roles they hold — the same cardinality as `auth_user_groups` today, so it is a rename rather than growth. Guardian grows as objects × permissions per object: every note creation writes three rows to `notes_notegroupobjectpermission`, every service request three, every attachment two. Guardian's tables grow with the **data**; `Grant` grows with the **org chart**. That is the strongest single argument for deriving object access rather than materialising it.

## How rows reach an organization

`OrgScoped` / `org_via` is unchanged from the earlier comment, with two additions.

```python
class OrgScoped(models.Model):
    """Declares how this model reaches the organizations that scope it.

    ``()``    my own ``organization`` FK
    ``(…)``   hop these relations; a row is in scope for *every* organization it reaches
    ``None``  deliberately unscoped (platform-shared) — reachable at the global tier only
    """

    org_via: ClassVar[tuple[str, ...] | None] = ()

    class Meta:
        abstract = True

    @classmethod
    def org_paths(cls) -> tuple[str, ...]:
        # cls.__dict__, not getattr — a subclass must not inherit its parent's paths
        if "_org_paths" not in cls.__dict__:
            cls._org_paths = tuple(_resolve(cls, seen=()))
        return cls._org_paths
```

```python
class Shelter(OrgScoped, BaseModel):       ...                              # owns the FK
class Room(OrgScoped, BaseModel):          org_via = ("shelter",)
class Bed(OrgScoped, BaseModel):           org_via = ("shelter",)
class Reservation(OrgScoped, BaseModel):   org_via = ("bed", "room")
class Referral(OrgScoped, BaseModel):      org_via = ("organization", "shelter")
class ClientProfile(OrgScoped, BaseModel): org_via = None
```

Resolution recurses, so the knowledge lives once:

```
Reservation.org_via = ("bed", "room")
  bed  → Bed.org_via = ("shelter",) → Shelter.org_via = () → ["organization_id"]
       → ["bed__shelter__organization_id"]
  room → ["room__shelter__organization_id"]
```

which is exactly the hand-written `organization_fields` list at `shelters/selectors/operator.py:160`, now derived. When `Shelter`'s own path changes, `Room` and `Bed` follow with no edit.

**Addition one: several relations mean union, and that is the general case.** A row genuinely can belong to more than one organization; single-org models are the degenerate case with one path. `Referral.org_via = ("organization", "shelter")` is then the entire relation-derived cross-org story: the receiving shelter's operators see the referral because it is in *their* organization's scope and they hold `view_referral` there. They cannot edit it because their role does not carry `change_referral`. No sharing table, no flag. It also closes the gap at `referrals/selectors.py:8`, where the receiving org cannot see a referral to its own shelter at all.

**Addition two: `None` is an explicit opt-out, not an omission.** Otherwise a model that forgot to declare scoping and a model that deliberately has none look identical. `ClientProfile.org_via = None` is greppable, and it lets a system check demand that every model a role grants a permission on has decided one way or the other:

```python
@register()
def check_org_scoping(app_configs, **kwargs):
    errors = []
    # PERM_MODEL is populated by PermissionSet.contribute_to_class, which already knows
    # the model when it builds each codename — no hand-kept list.
    for role in ROLES:
        for perm in role.permissions:
            model = PERM_MODEL.get(perm)
            if model and not issubclass(model, OrgScoped):
                errors.append(Error(
                    f"{model.__name__} is granted {perm} but does not declare org scoping. "
                    f"Set org_via, or org_via = None to declare it platform-shared.",
                    obj=model, id="permissions.E001"))

    for model in _org_scoped_models():
        for name in model.org_via or ():
            field = model._meta.get_field(name)
            if not (field.many_to_one or field.one_to_one):
                errors.append(Error(
                    f"{model.__name__}.org_via names {name!r}, which is multi-valued; "
                    f"the scope filter would duplicate rows.",
                    obj=model, id="permissions.E002"))
    return errors
```

It reads the role definitions from code, so it runs without a database. E002 is the bug recorded at `notes/types.py:336-348`.

Three models cannot satisfy a single-valued path today, and the design says so rather than working around it:

| Model | Why | Resolution |
|---|---|---|
| `ServiceRequest` | reaches an org only through the reverse M2Ms `provided_notes` / `requested_notes` (`notes/models.py:155,159`) | its own `organization` FK; it is created inside one org in one transaction (`notes/services.py:158`), so the backfill is `note.organization` |
| `Attachment` | reaches its parent by generic FK, which the ORM cannot traverse in a filter | denormalised `organization` FK, set at upload |
| `ClientProfile` + subtree | has no organization and never has — no migration in `clients/` has ever added one | `org_via = None` |

## The five tiers, as rows

| Tier | Row |
|---|---|
| user, global | `alice.user_permissions.add(view_private_shelter)` |
| group, global | `Group("Shelter Operator")`, `alice.groups.add(it)` |
| org role | `Grant(user=bob, role=Caseworker, scope=SELAH)` |
| container | *the same row* — the organization is the container; a note reaches it via `Note.organization`, a bed via `bed__shelter__organization` |
| org grants a user | `Grant(user=consultant, role=Caseworker, scope=SELAH)`, consultant not an `OrganizationUser` of SELAH |
| org grants an org | `Grant(organization=HeartForward, role=NoteViewer, scope=SELAH)` |
| shared subject matter | *no row* — `Referral.org_via = ("organization", "shelter")` |

The last two are deliberately different mechanisms because they are different facts. Delegated authority is a decision SELAH made and revokes by deleting a row. Shared subject matter is a property of the data, with nothing to grant or revoke — modelling it as a grant would mean writing rows at creation time, which is the guardian pattern this removes.

Object-level and organization-level are not separate mechanisms in this model. They are the same statement at different granularities of scope, which is why they are one table and one predicate term.

vecchp:
# The resolver and the surfaces (3/4)

Everything reduces to one question: **which organizations does this user hold `perm` in?**

```python
ALL = object()


def scopes(user, perm):
    """``ALL``, or a values-queryset of organization ids where *user* holds *perm*."""
    if not user.is_authenticated:
        return Grant.objects.none().values("scope")

    if user.has_perm(perm):                 # tiers 1 and 2 — Django's, including is_superuser
        return ALL

    cache = user.__dict__.setdefault("_scope_cache", {})
    if perm not in cache:
        app_label, codename = perm.split(".", 1)
        roles = Group.objects.filter(
            permissions__codename=codename,
            permissions__content_type__app_label=app_label,
        ).values("pk")
        at_my_orgs = Grant.objects.filter(user=user).values("scope")
        cache[perm] = Grant.objects.filter(
            Q(user=user) | Q(organization__in=Subquery(at_my_orgs)),
            role__in=Subquery(roles),
        ).values("scope")
    return cache[perm]


def visible(qs, user, perm, *, in_org=None):
    """The rows of *qs* on which *user* may exercise *perm*."""
    s = scopes(user, perm)
    if s is ALL:
        return qs                           # row-invariant: hoist, never filter by a constant

    paths = qs.model.org_paths()
    if not paths:                           # org_via = None and no global grant
        return qs.none()

    qs = qs.filter(reduce(or_, (Q(**{f"{p}__in": s}) for p in paths)))
    if in_org is not None:                  # the header confines the view, not the authority
        qs = qs.filter(reduce(or_, (Q(**{p: in_org}) for p in paths)))
    return qs


def can(user, perm, *, org):
    """Authority in an organization — the check for creates, which have no row yet."""
    s = scopes(user, perm)
    return s is ALL or s.filter(scope=org).exists()


def can_obj(user, perm, obj):
    """The single-row check *is* the row filter, applied to one row."""
    return visible(obj.__class__._base_manager.filter(pk=obj.pk), user, perm).exists()


def holds_anywhere(user, perm):
    """Menu visibility only — never a gate."""
    s = scopes(user, perm)
    return s is ALL or s.exists()
```

**`ALL` is a sentinel, not a list of organization ids.** A global grant is row-invariant. Materialising it as "every organization" would cost a query, put a large `IN` list into every filter, and be wrong: it would exclude rows whose organization is NULL, which `Shelter.organization`'s `SET_NULL` produces. Short-circuiting `visible()` to `return qs` is cheaper and more correct. `filter_for_user_q:126-131` already does this, but only opportunistically when the permission happens to be in `user._perm_cache`; here it is deterministic.

**The global test defers to `user.has_perm`,** so `is_superuser`, `user_permissions`, `user.groups` and any other installed backend are all handled and stay consistent with the admin. It cannot recurse into the new backend, which returns `False` whenever `obj is None`. During the migration this is temporarily untrue — while `PermissionGroup` still carries permissions, `has_perm` answers `True` for any caseworker — so steps 2 and 3 read `user.user_permissions` and the non-`PermissionGroup` entries of `user.groups` explicitly, and collapse to `user.has_perm` after teardown.

**The principal predicate** is where both cross-org shapes live. The first disjunct covers grants made to the person: an org role, and the "org grants a user" case, with membership never consulted. The second covers grants made to an organization the user acts in. "Acts in" is deliberately *any* grant there, because delegation means "everyone at the partner organization".

Delegation is one level, and **the query shape is what enforces it**: the inner subquery filters `user=user`, so it returns only directly-held scopes and a delegated grant can never appear in it. Transitive delegation would need a recursive CTE. It is not prevented by a flag that could be forgotten — it is prevented by the fact that the inner query does not select delegated rows.

**`can_obj` is `visible` on one row.** This is the most important line. Any other implementation of the single-row check is a second expression of the same rule, and two expressions drift — which is the current bug, where `HasPerm` and `permissioned_queryset` are two answers to one question. Deriving one from the other means multi-path models and `org_via = None` both work with no extra code, and a change to the scoping rule cannot be applied to one and forgotten in the other. Cost is one `EXISTS`, the same as guardian's object check. `_base_manager` rather than `_default_manager`, because `ClientProfileManager` hides merged profiles and a manager hiding a row must never read as "you lack permission".

`holds_anywhere` answers "should this menu item appear", which the Django admin needs because its objectless check gates whether the changelist opens. If it were used as a gate the original leak would be back; it is named to say so, and its only callers are the admin mixin and the frontend's permission list.

Caching mirrors `ModelBackend`'s `_perm_cache`: a dict on the user instance, so the lifetime is the request and there is no new registry. The cached value is a lazy queryset used as a subquery, so caching it does not evaluate it.

## The SQL

`notes` list, for a user with no global grant:

```sql
SELECT … FROM notes_note
WHERE notes_note.organization_id IN (
        SELECT g.scope_id
        FROM accounts_grant g
        WHERE g.role_id IN (
                SELECT ag.id FROM auth_group ag
                JOIN auth_group_permissions agp ON agp.group_id = ag.id
                JOIN auth_permission p          ON p.id = agp.permission_id
                JOIN django_content_type ct     ON ct.id = p.content_type_id
                WHERE p.codename = 'view_note' AND ct.app_label = 'notes')
          AND (g.user_id = 42
               OR g.organization_id IN (
                    SELECT scope_id FROM accounts_grant WHERE user_id = 42)))
```

One statement, uncorrelated, driven by the FK index on `accounts_grant.user_id` — a handful of rows per user, against a single-digit `auth_group`. Today's equivalent is a correlated `EXISTS` per row against `Organization` joined through `PermissionGroup` to `Permission` (`common/permissions/utils.py:260-272`), plus a separate `HasPerm` call, plus guardian's union for the object-level part. For a user with a global grant there is no filter at all.

## Three surfaces, no more

**Django — one auth backend.** This is why the backend seam is used rather than a private API: it makes `user.has_perm(perm, obj)` correct in templates, in the admin, and in any third-party code that asks Django.

```python
class ScopedBackend(BaseBackend):
    def authenticate(self, request, **kwargs):
        return None

    def has_perm(self, user_obj, perm, obj=None):
        return obj is not None and user_obj.is_active and can_obj(user_obj, perm, obj)
```

`ModelBackend` stays first and answers the objectless case; Django ORs them, so this is purely additive. Note the deliberate asymmetry: `has_perm(perm)` with no object means **global**, not "somewhere". That holds structurally, because scoped roles are never in `user.groups`.

**Django admin — one mixin.** `ModelAdmin.has_change_permission` ignores its `obj` by default, which is why guardian ships `GuardedModelAdmin`; this is the equivalent, and it is where the "anywhere" question is confined so it cannot leak into `has_perm`.

```python
class ScopedAdminMixin:
    def get_queryset(self, request):
        return visible(super().get_queryset(request), request.user, self._perm("view"))

    def _check(self, action, request, obj):
        perm = self._perm(action)
        return can_obj(request.user, perm, obj) if obj else holds_anywhere(request.user, perm)

    def has_view_permission(self, request, obj=None):   return self._check("view", request, obj)
    def has_change_permission(self, request, obj=None): return self._check("change", request, obj)
    def has_delete_permission(self, request, obj=None): return self._check("delete", request, obj)
```

Role editing is Django's existing Group admin. Granting a global role is the group picker on the User page. Granting a scoped role is a `Grant` inline on the Organization page. All stock.

**strawberry-django — one type mixin and one field extension.**

```python
class Scoped:
    perm: ClassVar[str]

    @classmethod
    def get_queryset(cls, queryset, info):
        return visible(queryset, get_current_user(info), cls.perm,
                       in_org=info.context.request.organization_id)


class Requires(DjangoPermissionExtension):
    """Field gate for mutations. The scope is the active organization."""

    DEFAULT_ERROR_MESSAGE = "You do not have permission to perform this action."

    def __init__(self, perm: str, **kwargs):
        kwargs.setdefault("fail_silently", False)
        super().__init__(**kwargs)
        self.perm = perm

    def resolve_for_user(self, resolver, user, *, info, source):
        org_id = info.context.request.organization_id
        if org_id is None:
            raise DjangoNoPermission("X-Organization-ID is required.")
        if not can(user, self.perm, org=org_id):
            raise DjangoNoPermission(self.message)
        return resolver()
```

For a mutation that loads a row, **the load is the check**: `visible(qs, user, perm).get(pk=…)` raising `DoesNotExist` → `PermissionDenied`, which `get_object_or_permission_error` (`common/graphql/utils.py:12`) already implements. No second extension, and the failure is a not-found rather than an oracle confirming the row exists.

Note what `in_org` separates: **grants decide authority, the header decides the view.** A user with authority in three organizations sees one at a time because the client says which. That is a UI concern, and conflating it with authorization is what makes `HasOrgPerm` reject a global role today.

## Where the existing call sites go

| Today | Sites | Becomes |
|---|---|---|
| `HasPerm(perm)` | 24 | `Requires(perm)` on mutations; nothing on reads — the type mixin covers them |
| `HasRetvalPerm(perm)` | 29 | `Scoped` type mixin |
| `PermissionedQuerySet(perm)` | 11 | `visible(...)` in the service |
| `HasOrgPerm(perm)` | 36 | `Requires(perm)` |
| `permissioned_queryset(...)` | 10 | `visible(...)` |
| `get_user_permitted_org(...)` | 8 | `can(user, perm, org=…)` |
| `resolve_permission_group(...)` | 11 | deleted — the organization is an argument |
| `assign_object_permissions(...)` | 6 | deleted — CHANGE/DELETE move onto the role |
| `user.has_perm(p)` | 2 | unchanged; now unambiguously means global |

vecchp:
# Getting there, and two fixes for this PR regardless (4/4)

## Four steps

Steps 2 and 3 run with both systems live and services dual-writing, so a rollback is a code revert.

**Step 1 — the frontend, which has to go first or none of this is reachable.** `CurrentUserOrganizationType.get_queryset` (`accounts/types.py:113`) is `queryset.filter(users=user)` — organizations the user is an `OrganizationUser` of — with permissions annotated by `ArrayAgg` filtered on `permission_groups__user=user`. A user holding a global role but a member of nothing therefore gets an empty organization list, sends no `X-Organization-ID`, and `hasPermission` is false everywhere:

| file | result |
|---|---|
| `apps/betterangels-admin/.../home.tsx:17` | permanent `Loading...` |
| `apps/betterangels-admin/.../PermissionGuard.tsx:13` | guarded routes redirect |
| `libs/react/shelter-operator/.../Dashboard.tsx:97` | `skip: !selectedOrganizationId` |
| `.../CreateShelterProfile.tsx:61-63` | `if (!activeOrg?.id) return;` — silent no-op |

Target shape: the organization list becomes `Organization.objects.filter(grants__user=user)` unioned with the delegated arm — the same query shape with one join changed — and a new `currentUser.permissions: [String!]!` carries the global tier so the UI can render for a global-role holder with no active organization.

**Migration A** creates `Grant`. Nothing reads it.

**Migration B** backfills: for each `PermissionGroup`, get-or-create the plain group named after its template carrying the template's permissions, then one `Grant(user, role, scope=pg.organization)` per member. `GLOBAL_SHELTER_OPERATOR` is the exception — its members go into `user.groups` on the plain global group, because that is now the global tier. Reversible; the reverse deletes the grants.

**Step 2 — the resolver and the two surfaces, shelters only.** Everything else keeps the check it has. **This closes SDB-218 without `bypasses_org_scoping`**, and cross-org works from here; it needs only a `Grant` inline on the Organization admin to be usable.

**Step 3 — the remaining domains.** Notes, tasks, referrals, clients, teams, reports. Add `ServiceRequest.organization` and `Attachment.organization`, move CHANGE/DELETE onto the roles, stop the guardian writes. `resolve_permission_group` and its first-match attribution die here.

**Migration C and step 4 — teardown.**

## What is deleted

| Gone | Where |
|---|---|
| **django-guardian** — package, two generic tables, three per-model table pairs, the backend, the settings, `assign_object_permissions`, its 6 call sites, `backfill_attachment_perms.py` | `accounts/models.py:97-120`, `notes/models.py:237-251`, `common/models.py:322-334` |
| `PermissionGroup`, `PermissionGroupTemplate`, `TemplateConfig.permissions`, `sync_group_permissions`, `seed_permission_templates`, `reconcile_org_groups`, `OrgRoleManager`, `REGISTRY.unscoped`, `group_name()` | `accounts/models.py:123-249`, `accounts/seed.py`, `accounts/services.py:278`, `accounts/role_manager.py` |
| `permissioned_queryset`, `_org_perm_exists_across_fields`, `_perm_q`, `perm_filter`, `get_user_permitted_org`, `resolve_permission_group` | `common/permissions/utils.py:208-342`, `accounts/permissions.py:30`, `accounts/selectors.py:97` |
| `organization_field` / `organization_fields` threaded through 10 selectors | `shelters/selectors/operator.py` |
| `HasOrgPerm`, `HasPerm`, `HasRetvalPerm`, `PermissionedQuerySet` at ~100 sites | `accounts/extensions.py`, `common/graphql/extensions.py` |

Added: `Grant` (~30 lines), `scoping.py` (~90), `access.py` (~70), `backend.py` (~12), `admin.py` (~25), `graphql.py` (~40), `roles.py` (~40), three migrations.

## Deliberately not built

None of these has a current consumer, and none changes the shape of the predicate.

| Not built | Cost to add |
|---|---|
| Per-object ACL — share one record with one person | nullable `scope_type`/`scope_id`, one `Q(pk__in=…)` disjunct |
| Team-level scope | nullable `scope_team`, a second set from `scopes()`, one more disjunct |
| Per-grant permissions with no role | nullable M2M, one more disjunct |
| Grant expiry | `expires_at`, one clause in `scopes()` |
| Transitive delegation | a recursive CTE — the only real change on this list |

**Deny rules are not on that list and should never be added.** The tiers can be a union precisely because grants only add. A denial forces an evaluation order, and then every future tier has to declare where it sits in it. The requirement usually reached for — private notes — is not a denial but a narrower permission, `notes.view_private_details` held by fewer roles. That composes.

## Two things worth fixing on this PR regardless of its shape

1. **The `ClientProfile` TODO reads backwards.** Under any of these designs it needs to say that client profiles are platform-shared by decision, not that they are pending scoping.
2. **`bypasses_org_scoping` would be the only editable field on every code-owned template.** `accounts/admin.py:117-130` makes `("name", "permissions")` readonly, so `accounts.change_permissiongrouptemplate` plus a tick on `Caseworker` grants cross-org access until the next `migrate` reverts it.

## Open decisions

- **`Shelter.organization` is nullable** (`shelters/models/shelter.py:78`, `SET_NULL`), so orphan shelters are in nobody's scope and reachable only at the global tier. Intended, or should the FK become required with a backfill?
- **`ClientProfile.org_via = None`** makes today's behaviour explicit and leaves the accepted risk from #2313 as it is: `delete_clientprofile` cascades every other organization's notes. The model makes the fix expressible; it does not make it.
- **Who administers a cross-org grant** — an organization admin self-serving in the portal, or BA staff in the Django admin. That decides whether step 2 ships a UI or only the admin inline.

PR #2408, opened 2026-09-01
fix maps current location button (OUT-223)

tglaz:
### fix maps current location button
OUT-223

## Summary by Sourcery

Fix map current-location handling and make the locate-me control reusable across map layouts.

New Features:
- Support reusable map locate-me controls with configurable positioning, styling, accessibility text, and location callbacks.

Bug Fixes:
- Fix the map current-location flow so the selected location and user-location indicator update after locating the user.

Enhancements:
- Centralize current-location behavior through the shared map locate-me button and return the resolved location to callers.


libs/expo/shared/ui-components/src/lib/Map/MapLocateMeBtn.tsx
```diff
@@ -1,18 +1,31 @@
...
 
+type TVariant = 'absolute' | 'relative';
+
 type TProps = {
   mapRef: RefObject<TMapView | null>;
   regionDelta?: TMapDeltaLatLng;
   duration?: number;
+  onLocated?: (location: LocationObject) => void;
+  variant?: TVariant;
+  style?: ViewStyle;
```
vecchp, line 19: Nit: this PR widens `LocateMeButton` to `StyleProp<ViewStyle>` (so it accepts style arrays), but `MapLocateMeBtn.style` is still `ViewStyle`. Consider `StyleProp<ViewStyle>` here too for parity, since the value is forwarded into a style array.

libs/expo/shared/ui-components/src/lib/Map/MapLocateMeBtn.tsx
```diff
@@ -27,13 +40,17 @@ export function MapLocateMeBtn(props: TProps) {
...
-      await goToUserLocation({
+      const newLocation = await goToUserLocation({
         mapRef,
         regionDelta,
         duration,
         onPermissionDenied,
       });
 
+      if (newLocation) {
+        onLocated?.(newLocation);
```
vecchp, line 55: `onLocated` is async (the picker's handler awaits `geocode`), but it's not awaited here, so `setDisabled(false)` runs while the consumer's geocoding is still in flight — a rapid second tap can fire overlapping geocodes, and an async rejection inside `onLocated` would become an unhandled promise rejection. Consider `await onLocated?.(newLocation);`.

libs/expo/shared/ui-components/src/lib/Map/MapLocationPicker/index.tsx
```diff
@@ -51,6 +48,14 @@ export function MapLocationPicker({
     );
   }, []);
 
+  async function onCurrentLocationChange(location: LocationObject) {
```
vecchp, line 51: Nit: the `location` parameter shadows the component's `location` state. Consider renaming (e.g. `userLoc`) for readability.

libs/expo/shared/ui-components/src/lib/Map/MapLocationPicker/index.tsx
```diff
@@ -90,8 +95,8 @@ export function MapLocationPicker({
   }, [initialLocation]);
 
   useEffect(() => {
-    if (propUserLocation) setUserLocation(propUserLocation);
-  }, [propUserLocation]);
+    setShowUserLocationDot(!!userLocation);
+  }, [userLocation, setShowUserLocationDot]);
```
vecchp: Minor: `setShowUserLocationDot` is a stable setter and doesn't need to be in the dependency array (exhaustive-deps doesn't require it). Harmless, just redundant.

libs/expo/shared/ui-components/src/lib/Map/MapLocationPicker/index.tsx
```diff
@@ -257,15 +247,12 @@ export function MapLocationPicker({
...
-          <IconButton
-            onPress={goToUser}
-            style={styles.shadowBtn}
-            accessibilityLabel="My location"
-            accessibilityHint="center map on your location"
-            variant="secondary"
-          >
-            <LocationArrowIcon color={Colors.PRIMARY} />
-          </IconButton>
+          <MapLocateMeBtn
```
vecchp, line 250: Minor a11y: the old button had `accessibilityLabel="My location"` / hint "center map on your location"; the picker now inherits the shared `LocateMeButton` label "userlocation button". Worth improving that label in the shared component.

vecchp (commented):
Reviewed the refactor — nice dedup: shared `MapLocateMeBtn` + `goToUserLocation` now returns the resolved location, and the picker reuses permission handling and shadow styling. Default `variant='absolute'` keeps `MapViewport` behavior unchanged, and CI is green. A few non-blocking suggestions below.

libs/expo/shared/ui-components/src/lib/Map/MapLocationPicker/index.tsx
tglaz: no

libs/expo/shared/ui-components/src/lib/Map/MapLocateMeBtn.tsx
tglaz: no

libs/expo/shared/ui-components/src/lib/Map/MapLocateMeBtn.tsx
tglaz: it's a simple callback. fire and forget. no await. unrelated to geocode. if rejectes then onLocated doesn't get called

vecchp (approved):

shiozakiyuko (approved):
Using my hotspot worked beautifully! 

iOS and Android both worked as expected including the "Allow BA to use your location?" notification.

Approved.

PR #2409, opened 2026-09-01
feat(permissions): grant-based authorization data model (ADR 0001) (DEV-2553)

vecchp:
## Summary

First PR of the grant-based authorization redesign, per **ADR 0001** (`docs/adr/0001-grant-based-authorization.md`). Schema-only — **nothing reads `Role`/`Grant` at runtime yet**.

## What's in this PR

- **`Role`** — an `auth.Group` subclass with an `is_global` flag. Global roles are held directly in `user.groups` (the global tier); scoped roles are granted via `Grant` rows.
- **`Grant`** — the only scoped authorization input: exactly one principal (user, or organization-as-delegator) and exactly one scope (org, or object once the object-grant arm is wired). DB constraints enforce the principal/scope XORs, reject self-delegation, and enforce uniqueness with `NULLS NOT DISTINCT` so org- and object-scoped rows never collide.
- **`OrgScoped`** — abstract base declaring how a model reaches organizations (`org_via`) with `org_paths()` resolving lookup paths. Applied to the core shelter data models (`Shelter`, `Bed`, `Room`, `Reservation`, `ShelterPhoto`). No schema change — the base contributes no fields.
- **System checks `permissions.E001`–`E005`** — scoped role in `user.groups`, global role in a `Grant`, object-grant whitelist, multi-valued `org_via` hops, unscoped role permissions. Data-reading checks return `[]` when the DB is unreachable or not yet migrated, so `manage.py check`/`makemigrations` keep working pre-migrate.

## Tests (TDD, 27 new)

- `accounts/tests/test_grant.py` — Role/Group semantics + every Grant DB invariant (principal XOR, scope XOR, self-delegation, per-principal uniqueness, org-vs-object rows not colliding).
- `common/tests/test_org_scoping.py` — `org_paths()` resolution for the shelter models + the two declaration errors (multi-valued hop, hop onto an unscoped model).
- `common/tests/test_permission_checks.py` — E001–E005 fire and stay quiet correctly.

All 27 pass; `ruff` + `mypy` clean; `manage.py check` reports 0 issues.

## Follow-up PRs (per ADR 0001 §4)

- PR 2: `RoleDef` provisioning + `sync_roles` + backfill (GSO → global role; shelter memberships → `Grant`s)
- PR 3: `scopes()`/`visible()`/`can()` predicate + shelter cutover (global + user arms) — closes SDB-218
- PR 4: org→org delegation + admin inline
- PR 5: frontend reachability (grants-based org list, header-optional)
- PR 6: object grants + clients/notes cutover
- PR 7: teardown

## Summary by Sourcery

Establish the schema and validation foundations for grant-based authorization without changing runtime permission evaluation.

New Features:
- Add organization-independent roles and scoped grants supporting user, organization-delegated, organization-scoped, and future object-scoped authorization.

Bug Fixes:
- Prevent invalid authorization data through database invariants covering principal and scope exclusivity, self-delegation, duplicate grants, and global-versus-scoped role misuse.

Enhancements:
- Introduce reusable organization-path declarations for shelter-domain models and deploy-time permission system checks E001–E005.
- Record grant lifecycle changes through grant event history.

Documentation:
- Document the grant-based authorization architecture, organization scoping, migration roadmap, client write tiers, and guardian-equivalence design.

Tests:
- Add coverage for role and grant invariants, organization-path resolution, and permission system checks.


docs/adr/0001-grant-based-authorization.md
```diff
@@ -0,0 +1,418 @@
...
+class Room(OrgScoped, BaseModel):         org_via = ("shelter",)
+class Reservation(OrgScoped, BaseModel):  org_via = ("bed", "room")
+class ShelterPhoto(OrgScoped, BaseModel): org_via = ("shelter",)
+# outreach (future cutovers):
+class Note(OrgScoped, BaseModel):         org_via = ("organization",)
+class Referral(OrgScoped, BaseModel):     org_via = ("organization", "shelter")
+class ClientProfile(OrgScoped, BaseModel): org_via = None   # platform-shared by decision
+```
+
+**Reference-data note:** most of GSO's permissions are for *global reference data*
```
mikefeldberg, line 250: This note seems a bit misleading. It makes it sound like the "global role" exists to access non-org scoped data (city, spa, etc.). That's _kinda_ true in that the global shelter operator is the only role that will be able to modify them, but any authenticated user can view them.

apps/betterangels-backend/common/permissions/checks.py
```diff
@@ -0,0 +1,173 @@
...
+    try:
+        Role = apps.get_model("accounts", "Role")
+
+        errors: list[Error] = []
+        for role in Role.objects.prefetch_related("permissions__content_type"):
+            for permission in role.permissions.all():
+                model = permission.content_type.model_class()
+                if model is None or model._meta.abstract:
+                    continue
+                if not issubclass(model, OrgScoped):
```
mikefeldberg, line 160: fixed in PR 2

~not sure if this gets fixed in a subsequent PR but, as is, it'll will blow up on global shelter operator since it's a Role with perms for a bunch of non `OrgScoped` models. Should either move all those models to `OrgScoped` or only check this for `is_global=False`~

mikefeldberg (commented):
I shortened these

## 🔴 High

### 1. E005 will fail on real role data — the check contradicts the ADR's own design

kept for reference, but I see that this gets fixed in PR 2

### 2. `org_paths()` returns `()` for `org_via=None`, but the ADR's `visible()` checks `paths is None`

kept for reference, but I see that this gets fixed in PR 7

---

## 🟠 Medium

### 3. E004 doesn't cover two of the three `org_paths()` failure modes

`org_paths()` raises `TypeError` for three conditions:
- (a) multi-valued hop — ✅ covered by E004 (`_org_via_errors_for_model`)
- (b) hop to a model that doesn't declare `OrgScoped`
- (c) `org_via = ()` with no single-valued `organization` FK

E004 (`check_org_via_hops_are_single_valued`) only checks (a). Conditions (b) and (c) surface only at **runtime**, when the scope filter first calls `org_paths()` — despite the ADR §2.7 and the `org_paths()` docstring both claiming E004 surfaces "the same condition as a deploy-time error." Your own `test_a_hop_to_an_unscoped_model_raises` pins the runtime raise but has no E004 counterpart. All three checks are DB-free, so E004 should validate all of them.

### 4. `grant_has_exactly_one_scope` lets a stray `scope_object_id` through, creating a uniqueness bypass

The org-scope branch requires only `scope_org__isnull=False AND scope_object_type__isnull=True` — it doesn't require `scope_object_id__isnull=True`. So this row is valid:

```
scope_org=OrgA, scope_object_type=NULL, scope_object_id=42
```

### 5. Phase-1 backfill must satisfy the checks at every `migrate` checkpoint

E001/E002/E003/E005 are registered on `Tags.models`, which runs on **every** management command — including `migrate`. So the data-only backfill (GSO memberships → global `Role`; Shelter Operator memberships → `Grant` rows) must not pass through any intermediate state that violates E001/E002 (e.g. a scoped `Role` temporarily in `user.groups`), or `migrate` fails mid-deploy. State this explicitly in the ADR §4 / PR description. (The `DatabaseError` guard itself is correct.)

### 6. `Grant` indexing — verify when the predicate lands (phase 2)

The unique constraints give indexes led by `(principal_user, …)` and `(principal_org, …)`, which serve `scopes()`'s direct-grant and org-principal queries reasonably well. The delegation arm (`principal_org__in=Subquery(at)`) and `scope_org`-led lookups aren't obviously covered. Not a blocker for a schema-only PR, but put an `EXPLAIN` check on the phase-2 agenda.

---

## 🟡 Minor / polish

- **E001 N+1** — one extra query per matching user; fine for a deploy-time check, but a `prefetch_related` would tidy it.
- **Missing tests worth adding:**
  - duplicate *object*-scoped grants (the `nulls_distinct=False` behavior on the object columns) and the org-vs-object row *not* colliding — the exact rationale in the code comment;
  - the `scope_object_id`-stray row from #4;
  - `org_via = None` (from #2) and `org_via = ()` with a missing `organization` FK (from #3).
- **Forward references that don't exist yet:** the `Role` docstring and ADR mention `RoleDef` and `accounts.seed.sync_roles` — neither is in this diff. Fine as forward-refs, but the ADR's phase-1 scope ("provisioning") means they ship in this PR or the promise dangles.
- **`_org_paths` annotation** (`tuple[str, ...] | None`) is slightly misleading: `None` is only the "uncomputed" sentinel, and the computed value for platform-shared is `()` — same confusion as #2; fixing the return type resolves it.
- **ADR nits:** §2.8 table has a broken line-wrap; the `visible()` object-arm line `Q(pk__in=qs.values("pk")) | ...` is a tautology placeholder and should be marked as such.

docs/adr/0001-grant-based-authorization.md
vecchp: Fixed in 1a2d615a on this PR. The note now says reference lookups are open to any authenticated user; what lives on the global role is the write/admin surface for global data — which no org-scoped role can carry (the models don't declare OrgScoped, so E005 would fire) and for which an org-context write would be meaningless.

apps/betterangels-backend/common/permissions/checks.py
vecchp: Confirmed resolved: E005 now filters is_global=False (scoped roles only) in PR #2410 (perm/grant-roles), matching your note. At #2409 the all-roles check is vacuous in practice — no Role rows exist until #2410 provisions them — so it is stage-accurate here and is superseded when #2410 merges.

vecchp: Confirmed resolved: E005 now filters is_global=False (scoped roles only) in PR #2410 (perm/grant-roles), matching your note. At #2409 the all-roles check is vacuous in practice — no Role rows exist until #2410 provisions them — so it is stage-accurate here and is superseded when #2410 merges.

mikefeldberg (approved):

PR #2410, opened 2026-09-01
feat(permissions): Role provisioning + grant backfill (ADR 0001 phase 1) (DEV-2553)

vecchp:
## Summary

Second PR of the grant-based authorization redesign, per **ADR 0001** §4 phase 1. **Stacks on #2409** (`perm/grant-redesign`). Provisioning + backfill — `Role`/`Grant` are still not read at runtime (that's PR 3).

## What's in this PR

- **`RoleDef`** (`common/permissions/config.py`) — code-owned role definition with an `is_global` flag; shelter `RoleDef`s derived from the legacy templates in `shelters/groups.py` so permission lists keep one source.
- **`sync_roles`** — idempotent `post_migrate` provisioning of `Role` rows (get_or_create + reconcile permissions + `is_global`), wired into the `accounts` post_migrate hook and exposed as `manage.py sync_roles`.
- **Backfill** — `backfill_shelter_grants` converts legacy *Shelter Operator* `PermissionGroup` memberships into `Grant(user, role, scope)` rows; `backfill_global_role_members` moves GSO members onto the global `Role`. **Only shelter roles convert** — every other domain's `PermissionGroups` are untouched.
- **`OrgScoped` declarations** for every model the scoped Shelter Operator role grants: `Video`, `MediaLink`, `Schedule`, `ShelterAvailability`, `ContactInfo` (`org_via = ("shelter",)`) and `ClientProfile` (platform-shared, `org_via = None`).
- **E005 now audits scoped roles only** — a global role's permissions are never org-confined, so their models need no declaration until a scoped role holds them (ADR updated).
- **Bug fix (found by the new tests):** the PR-1 `Grant` uniqueness constraints were not partial — two *different* users holding the same role at the same scope collided through the org-principal index. Migration `0009` makes `unique_user_grant` / `unique_org_grant` partial (`condition=Q(principal_user__isnull=False)` / `Q(principal_org__isnull=False)`), plus a regression test.

## Tests (TDD)

- `accounts/tests/test_roles.py` — sync provisions scoped + global roles, idempotent, reverts permission/`is_global` drift, backfill creates one grant per member, idempotent, converts only the shelter role, moves GSO members, and E005 is quiet with seeded roles.
- New: two-users-same-scope regression; E005 global-role exemption; `org_paths` for the newly declared models.

40 tests in the grant suite pass; `ruff` + `mypy` clean; `manage.py check` → 0 issues; `manage.py sync_roles` verified end-to-end on the dev DB.

## Follow-ups (per ADR 0001 §4)

- PR 3: `scopes()`/`visible()`/`can()` predicate + shelter cutover (global + user arms) — closes SDB-218
- PR 4: org→org delegation + admin inline
- PR 5: frontend reachability
- PR 6: object grants + clients/notes cutover
- PR 7: teardown

## Summary by Sourcery

Provision grant-based roles and bootstrap shelter authorization data while preserving legacy permissions for domains not yet migrated.

New Features:
- Add code-owned role definitions with idempotent provisioning and a management command for synchronizing roles and backfilling grants.
- Backfill scoped Shelter Operator memberships into organization-scoped grants and move Global Shelter Operator members to the global role.
- Declare organization scoping for shelter-related models and platform-shared client profiles.

Bug Fixes:
- Correct grant uniqueness constraints so different users can hold the same role at the same organization scope.
- Prevent provisioning permissions that resolve to nonexistent model content types.

Enhancements:
- Restrict permission audits to scoped roles and prevent scoped roles from being assigned through Django group administration.
- Keep global role permissions available through the Django admin while hiding code-owned role records from editable group administration.

Documentation:
- Update the authorization ADR with scoped-role auditing, provisioning structure, uniqueness behavior, and phase-one transition caveats.

Tests:
- Add coverage for role synchronization, drift reconciliation, grant backfill idempotency and scope, global-role migration, organization paths, audit exemptions, phantom permissions, and grant uniqueness.


mikefeldberg (commented):
## 🟠 Medium

### 1. `_resolve_permissions` breaks for custom perms. It derives ContentType from the codename via `c.rsplit("_", 1)[-1]`.

- `view_private_shelter` → `"shelter"` ✓
- `change_shelter_is_reviewed` → `"reviewed"` ✗ — creates `ContentType(shelters, "reviewed")`, which has no model class

This was pre-existing in `seed.py`, but `sync_roles` now provisions the **full GSO list** through it on every migrate, so a phantom `(shelters, "reviewed")` + `change_shelter_is_reviewed` Permission row will be created in production. E005 tolerates it (`model_class()` → `None` → skipped), but any future `content_type.model_class()` usage or the object arm will silently mis-handle it. 

Worth a guard: verify every RoleDef permission's codename suffix resolves to a real model (or resolve the model from the app's actual models instead of the codename).

## 🟡 Low / notes

- **E005 now exempts global roles entirely** — that's the documented rule ("fires the moment a scoped Role holds one"), and it's the right call for phase 1. But it means nothing guards a global role granting perms on an *undeclared org-bearing* model until a scoped role or the object arm touches it. Fine as a deliberate deferral; just be aware the declarations now ride on phase-2's `visible()`/object-arm work, not on a check.
- **Seeded `Role` rows are `auth.Group` rows.** They'll appear in the Django auth admin's Group changelist alongside `PermissionGroup`s. Anyone adding a user directly to the scoped "Shelter Operator" Role group there would silently make it global — E001 catches it only at `check` time. Consider hiding/guarding the auth admin for Role rows.
- **`backfill_*` hard-`get` the Role** (`Role.objects.get(name=...)`) — they raise `DoesNotExist` if called without `sync_roles()` first. The hook and command both order it correctly, but a `get_or_create` (or explicit precondition) would make the functions self-sufficient.
- **Backfill only matches `template__name`** — label-only legacy groups (hand-made in admin) are intentionally skipped, which `test_backfill_converts_only_the_shelter_role` pins. Just confirming that's the intent (it means any org whose Shelter Operator group is label-based won't convert — fine for now, since seeded groups are template-based).
- **`backfill_*` on every `migrate`** acts as a continuous reconciler for shelter memberships (new orgs' legacy groups get backfilled on the next migrate) — that's arguably a feature during transition, but it means the "backfill" name undersells a permanent sync. Worth a comment so a future reader doesn't try to delete it as a one-time migration.

vecchp:
Thanks @mikefeldberg — all addressed in `f0c4da1a` on `perm/grant-roles`.

**🟠 Medium — `_resolve_permissions` custom-perm heuristic**

Agreed, and confirmed against the current lists: only `view_private_shelter` flows through the resolver today and its last token happens to be the real model, so no phantom row is actually created in production *yet*. The guard is now in place so the latent case can't slip in silently: `sync_roles` calls a new `_raise_on_phantom_role_permissions`, which refuses any RoleDef permission whose resolved ContentType has no model class (e.g. `change_shelter_is_reviewed` → `(shelters, "reviewed")`). It runs inside `sync_roles`' transaction, so the phantom rows roll back with the error, and `manage.py`/`migrate` fails loudly instead of provisioning them. Two tests added: one asserts every provisioned permission binds a real model, one asserts a phantom RoleDef raises.

**🟡 Low/notes**

1. **E005 global exemption** — agreed it's the right phase-1 call and it's deliberate; left as-is (the moment a scoped role holds one of those perms, E005 re-arms). No change.
2. **Seeded `Role` rows in the auth Group surface** — addressed. Scoped `Role` rows (`is_global=False`) are now excluded from the auth `Group` admin **and** the `User` admin groups picker, so the bare "Shelter Operator" Role can't be picked next to the same-named org-scoped `PermissionGroup`s (E001 stays the deploy-time backstop). Global `Role` rows remain selectable — the Django admin is the sanctioned global-grant surface (ADR 0001 §3).
3. **`backfill_*` hard-`get` the Role** — acknowledged; left as a documented ordering contract for now (both call sites run only after `sync_roles`), since `get_or_create` would silently mask a mis-ordering. Happy to revisit.
4. **Backfill matches `template__name` only** — confirmed intent (pinned by `test_backfill_converts_only_the_shelter_role`). The cutover implication (label-based groups won't convert) is now called out in ADR §4 phase 1.
5. **Backfill-on-every-`migrate` as a continuous reconciler** — fair; the ADR §4 phase-1 row now documents that the phase-1 backfill is **add-only and migrate-time only** (a removed membership leaves a stale `Grant`; a brand-new org gets none until the next migrate) and that phase 2's dual-write must write `Grant`s synchronously on membership change and make the `reconcile` command *revoke* stale rows, not just backfill.

Also folded in from the adversarial pass: a `list()` copy in `RoleDef.from_template` (frozen dataclass was aliasing the legacy template's permission list) and stale `accounts.seed.sync_roles` → `accounts.services.sync_roles` docstring cross-refs.

Verification: mypy + ruff clean, `manage.py check` 0 issues, `makemigrations --check` clean, 110 tests pass (grant + admin suites) on an isolated DB.

mikefeldberg (approved):

PR #2411, opened 2026-09-01
feat(permissions): grant authorization selectors + write services (ADR 0001 §2.4/§2.9) (DEV-2553)

vecchp:
## Summary

Third PR of the grant-based authorization redesign. **Stacks on #2410** (`perm/grant-roles`). Implements the predicate as **selectors** and the write side as **services**, per `docs/styleguides/python.md` (HackSoft service/selector pattern — pinned in ADR 0001 §2.9).

## What's in this PR

- **`common/permissions/selectors.py`** (read side, pull-only):
  - `scopes(user, perm)` — `ALL` sentinel or the org ids where the user holds the perm via scoped `Grant`s; memoized per request on the user instance.
  - `visible(qs, user, perm, *, in_org=None)` — the permission-aware queryset. Global tier hoists (unconfined); platform-shared models (`org_via=None`) show all rows to any holder; org-scoped models filter by `org_paths()`; models not yet declared `OrgScoped` fail closed. `in_org` confines **only finite scopes** — a global holder is never org-confined by a stale header (ADR §2.4).
  - `can(user, perm, *, org)` / `can_obj(user, perm, obj)` / `can_anywhere(user, perm)` — read-only authorization checks for services/API.
  - The global tier is read explicitly (superuser, global Role in `user.groups`, `user_permissions`) — **not** `has_perm` — until the legacy `PermissionGroup` rows that pollute it are gone (ADR §2.4).
- **`accounts/services.py`** (write side, push-only): `grant_create` (`full_clean` before save, per the styleguide), `grant_delete`, `role_assign` (global tier only — refuses scoped roles, matching check E001), `role_remove`.

**Not in this PR:** the shelter-domain cutover (flipping `shelter_queryset`/mutations to `visible()`, the `createShelter` target-org convention, and the assign-service dual-write) — it lands next so the transition stays consistent (backfill from #2410 keeps Grants in sync via `post_migrate`).

## Tests (17 new, 57 in the grant suite)

- `common/tests/test_permissions_selectors.py` — org-scoped vs global visibility on real shelters, `in_org` confinement, superuser, no-grant → nothing, platform-shared `ClientProfile` reads, fail-closed for undeclared models, `_scope_cache` memoization, `can`/`can_obj`/`can_anywhere`.
- `accounts/tests/test_grant_services.py` — `grant_create` creates + validates before save, `grant_delete` revokes, `role_assign` global vs scoped, `role_remove`.

All 57 pass; `ruff` + `mypy` clean.

## Follow-ups (per ADR 0001 §4)

- **PR 4**: shelter cutover — `shelter_queryset`/mutation surface → `visible()`/`can()`, `createShelter` explicit org input, assign-service dual-write (closes SDB-218)
- PR 5: org→org delegation + admin inline
- PR 6: frontend reachability
- PR 7: object grants + clients/notes
- PR 8: teardown

## Summary by Sourcery

Introduce the grant-based authorization read selectors and write services needed to centralize permission checks and role management.

New Features:
- Add authorization selectors for resolving global or organization-scoped permissions, filtering visible querysets, and checking organization, object, and platform-wide access.
- Add grant and global-role services for creating and revoking scoped grants and assigning or removing global roles.

Enhancements:
- Fail closed for models without organization-scope declarations and preserve global authority when applying optional organization filters.

Tests:
- Add coverage for selector visibility, scope memoization, authorization checks, platform-shared models, and grant service behavior.

### DIFF
--- a/apps/betterangels-backend/accounts/services.py
+++ b/apps/betterangels-backend/accounts/services.py
@@ -20,10 +20,12 @@ from organizations.models import Organization, OrganizationOwner, OrganizationUs
 from .emails import base_url_for
 from .groups import ORG_ADMIN
 from .models import (
+    Grant,
     OrganizationProfile,
     OrgTypeChoices,
     PermissionGroup,
     PermissionGroupTemplate,
+    Role,
 )
 from .models import User as UserModel
 from .role_manager import OrgRoleManager
@@ -609,3 +611,41 @@ def backfill_global_role_members() -> None:
     for group in groups.prefetch_related("user_set"):
         for user in group.user_set.all():
             user.groups.add(role)
+
+
+# ── Grant write services (ADR 0001 §2.10) ────────────────────────────────
+
+
+def grant_create(*, user: UserModel, role: Role, scope_org: Organization) -> Grant:
+    """Grant *user* the scoped *role* at *scope_org* (ADR 0001 §2.2).
+
+    Validates via ``full_clean`` (which checks the model constraints since
+    Django 4.1) before saving, per the repo styleguide.
+    """
+    from accounts.models import Grant
+
+    grant = Grant(principal_user=user, role=role, scope_org=scope_org)
+    grant.full_clean()
+    grant.save()
+    return grant
+
+
+def grant_delete(*, grant: Grant) -> None:
+    """Revoke a scoped grant — the audit trail is pghistory's, not a flag."""
+    grant.delete()
+
+
+def role_assign(*, user: UserModel, role: Role) -> None:
+    """Add *user* to a *global* Role's group — the global tier (ADR 0001 §2.1).
+
+    Scoped roles are never placed in ``user.groups`` (checks ``permissions.E001``);
+    they are granted through a ``Grant`` instead.
+    """
+    if not role.is_global:
+        raise ValidationError(f"Role {role.name!r} is scoped; grant it via grant_create, not user.groups.")
+    user.groups.add(role)
+
+
+def role_remove(*, user: UserModel, role: Role) -> None:
+    """Remove *user* from a global Role's group."""
+    user.groups.remove(role)
--- a/apps/betterangels-backend/accounts/services.py
+++ b/apps/betterangels-backend/accounts/services.py
@@ -622,8 +622,6 @@ def grant_create(*, user: UserModel, role: Role, scope_org: Organization) -> Gra
     Validates via ``full_clean`` (which checks the model constraints since
     Django 4.1) before saving, per the repo styleguide.
     """
-    from accounts.models import Grant
-
     grant = Grant(principal_user=user, role=role, scope_org=scope_org)
     grant.full_clean()
     grant.save()
--- a/apps/betterangels-backend/accounts/services.py
+++ b/apps/betterangels-backend/accounts/services.py
@@ -634,10 +634,33 @@ def grant_delete(*, grant: Grant) -> None:
 
 
 def role_assign(*, user: UserModel, role: Role) -> None:
-    """Add *user* to a *global* Role's group — the global tier (ADR 0001 §2.1).
+    """Grant *user* a *global* Role by adding it to ``user.groups``.
 
-    Scoped roles are never placed in ``user.groups`` (checks ``permissions.E001``);
-    they are granted through a ``Grant`` instead.
+    .. warning::
+       **GLOBAL AUTHORITY — read before calling.** This is the global tier
+       (ADR 0001 §2.1): once *role* is in ``user.groups``, *user* can exercise
+       every permission *role* carries **across every organization**, with no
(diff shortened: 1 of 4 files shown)


mikefeldberg (commented):
## 🔴 High

### 1. The delegation arm is missing from `scopes()` — org→org grants are inert

fixed in PR 5, but maybe add a comment that it's intentionally missing here

## 🟠 Medium

### 2. `grant_create` doesn't refuse global roles — the write surface is asymmetric

`role_assign` has the guard:

```python
if not role.is_global:
    raise ValidationError(...)
```

But `grant_create` has no mirror: passing a global role creates a `Grant` referencing it, which E002 only catches later at `manage.py check` time. The ADR §2.6 says explicitly "Global roles (`is_global=True`) are never granted through a `Grant`," and `grant_create` is *the* service that's supposed to enforce the write surface. Add the symmetric check (`if role.is_global: raise ValidationError(...)`) plus a test — the pair `role_assign`-refuses-scoped / `grant_create`-refuses-global makes the two-tier invariant airtight at the API boundary, not just at check time.

### 3. Committing `--reuse-db` to `settings.json`

I assume this was unintentional

## 🟡 Low / nits

- **`grant_create` re-imports `Grant` inside the function** while it's already imported at the top of `services.py` (the diff adds it to the top-level import). Redundant — drop the inner import.
- **`role_remove` lacks the symmetric guard.** `role_assign` refuses scoped roles, but `role_remove(user, scoped_role)` is a silent no-op. Minor consistency; consider guarding or documenting.
- **`can(..., org: Any)`** — could be typed `Organization` for clarity (minor).
- **`scopes` memoization on `user.__dict__["_scope_cache"]`** — correct per the ADR's per-request intent, but note it's stale if a grant is created/revoked for the same user instance mid-request. Fine for now; just be aware if a mutation ever grants-and-then-queries in one request.
- **`test_grant_create_validates_before_saving`** — good test, and it confirms `full_clean()` catches the check constraint (which is the right service-layer behavior). Nice.
- **`test_unscoped_model_fails_closed`** — this pins that even a *global* role holder sees nothing on an undeclared model via `visible()`. That's the correct fail-closed default, and worth keeping — it's the guardrail that makes the E005 deferral safe.

vecchp:
Thanks @mikefeldberg — reviewed against the synced branch; most items addressed in `f64f27e4` on `perm/grant-predicate`, the rest are intentional carries.

**🔴 High — delegation arm missing from `scopes()`**
Agreed and intentional for this PR (org→org delegation is its own PR in the stack). Now documented on `scopes()` itself so no reader mistakes it for an oversight: *"Direct user grants only at this stage — org→org delegation is deliberately absent here and lands with the delegation PR (ADR 0001 §2.4)."*

**🟠 Medium — `grant_create` doesn't refuse global Roles**
Agreed — the write surface should mirror `role_assign`. This is already carried by the grant-hygiene PR (#2418), which adds the symmetric `if role.is_global: raise` refusal to `grant_create`/`grant_delegate` plus a test (mirroring E002). I kept it out of #2411 deliberately to avoid backporting a fix that would then conflict/drop when #2418 re-parents onto this branch. Flagged in the review notes so it isn't lost.

**🟠 Medium — `--reuse-db` committed to `settings.json`**
Not present on the current tree — `--reuse-db` exists only in the backend `pyproject.toml` as intended pytest `addopts`; no `settings.json`/`.vscode` config in the branch. Non-issue here.

**🟡 Low/nits**
- **Redundant inner `Grant` import in `grant_create`** — dropped (`f64f27e4`); it's top-level-imported.
- **`role_remove` lacks the symmetric guard** — acknowledged. `role_remove(user, scoped_role)` is a harmless no-op because E001 means scoped roles are never in `user.groups`; the invariant is enforced on the E001/`role_assign` side. Leaving as-is unless you'd prefer a defensive guard.
- **`can(…, org: Any)`** — cosmetic; leaving `Any` for now since callers pass an `Organization` instance and tightening buys little here.
- **`scopes` memoization staleness** — acknowledged; per-request intent per ADR, fine as documented.
- 👍 Thanks — `test_grant_create_validates_before_saving` and `test_unscoped_model_fails_closed` (the latter pins the fail-closed guardrail that makes the E005 deferral safe) were both kept.

**Additional adversarial fixes folded into `f64f27e4`:**
- `scopes()` now filters `scope_org__isnull=False` — an object grant (NULL `scope_org`) can never register as an org scope or as "holds the perm somewhere" for a platform-shared model.
- `can_obj()` carries a C1 warning: on platform-shared models (`org_via = None`) `visible()`'s read rule leaks in, so no platform-shared single-row *write* may route through it until the object-arm/fail-closed fix lands.
- New test: a narrower global Role (VIEW only) composes per-permission — `ALL` for the held perm, nothing for an unheld one (ADR §2.2).

Verification: 127 tests pass on an isolated DB, mypy + ruff + format clean.

mikefeldberg (approved):

PR #2412, opened 2026-09-01
feat(perm): cut over the shelter domain to grant-based authorization (DEV-2555)

vecchp:
## Summary

Closes **SDB-218** — flips the shelter domain from `permissioned_queryset` / `HasOrgPerm` to the grant-based selectors from ADR 0001 (§2.4, §2.6). This is **PR 4** of the grant-redesign stack, stacked on #2411.

## What changed

**Read side (fail-closed, header optional)**
- `shelter_queryset` / `room_queryset` / `bed_queryset` / `reservation_queryset` now wrap `visible()` instead of `permissioned_queryset`; `organization_id` is optional (`str | None`).
- Output-type `get_queryset`s use `active_org` (ADR §2.6): no header ⇒ view confined to grant scopes; global holders are never org-confined by a stale header (§2.4).
- Removed `HasOrgPerm` from every shelter query field (`operatorShelter(s)`, `shelter`, `bed(s)`, `room(s)`, `reservation(s)`). Unauthorized reads now return **no rows** (list) / **not-found** (single object) instead of a 403.

**Write side (authority via `can()`)**
- `create_shelter` accepts an explicit `organization_id` input (global holders can target an org without the header), validates the org exists, then `can(user, ADD, org=target)` → `OperationInfo(kind=PERMISSION)` on failure.
- `room_create` / `bed_create` / `reservation_create` now check `can(user, <Model>.ADD, org)` in the service (defense-in-depth beyond the parent VIEW lookup).
- Delete paths thread the exact permission (`DELETE`) through `visible()`; a user with VIEW but not DELETE fails closed.
- Removed `HasOrgPerm` from all 16 shelter mutations.

**Transition dual-write (ADR §4 phase 2)**
- `OrgRoleManager.add_roles` / `remove_roles` / `clear_roles` / `replace_roles` now mirror scoped-role memberships into `Grant` rows (authoritative for shelters) alongside the legacy `PermissionGroup` membership (authoritative for everything else). Only role-backed templates (Shelter Operator) produce Grants; `CASEWORKER` etc. keep legacy-only.

**Tests**
- Retargeted shelter permission setup from legacy `PermissionGroup`/`user_permissions` mutations to Roles + Grants.
- Asserted fail-closed read semantics (empty / not-found, not 403) and `OperationInfo(kind=PERMISSION)` for unauthorized creates.
- Bumped query-count budgets for the grant subquery (one scoped-subquery per org-path hop; memoized per request).
- New coverage: `accounts/tests/test_role_manager_grants.py` (dual-write) and `shelters/tests/test_grant_cutover.py` (GSO cross-org reads, target-org creates, DELETE gating).

## Validation
- `shelters/` + `accounts/` + `common/` suites: **941 passed**
- `ruff format --check` + `ruff check`: clean
- `mypy --strict` (worktree PYTHONPATH): clean
- `manage.py check`: 0 issues

## Known cost (documented)
The grant subquery adds ~3 queries per distinct permission per request (global-role check, direct-perm check, grant scope), memoized per user/perm in `_scope_cache`. Query-count tests were updated to the new budget; a follow-up can combine the two global-tier checks into one query.

## Stack
1. #2409 — models + checks (base)
2. #2410 — provisioning + backfill (base)
3. #2411 — selectors + write services (this PR's base)
4. **#… — this PR (shelter cutover)**
5. org→org delegation
6. frontend reachability
7. clients/notes + object grants cutover
8. teardown

## Summary by Sourcery

Cut over the shelter domain to grant-based authorization while preserving legacy role memberships during the transition.

New Features:
- Enable shelter reads and writes through grant-based authorization, including optional organization headers and explicit target organizations for shelter creation.

Bug Fixes:
- Make unauthorized shelter reads fail closed with empty results or not-found responses instead of authorization errors.
- Ensure create, clone, update, and delete operations enforce the specific required grant, including separate create and delete authority checks.

Enhancements:
- Replace legacy shelter permission checks with grant-based visibility and service-layer authorization.
- Dual-write scoped role memberships to Grant records while retaining legacy permission-group memberships for the transition.
- Document and test authorization coverage, global-role cross-organization access, target-organization creation, and fail-closed behavior.

Documentation:
- Clarify active-organization behavior for shelter entity services in the grant authorization ADR.

Tests:
- Retarget shelter authorization tests to Roles and Grants and add coverage for grant dual-writing and the shelter authorization cutover.


apps/betterangels-backend/shelters/types/outputs.py
```diff
@@ -283,7 +283,9 @@ class OperatorShelterType(ShelterTypeMixin):
     @classmethod
     def get_queryset(cls, queryset: QuerySet, info: Info) -> QuerySet[models.Shelter]:
         user = cast(User, get_current_user(info))
-        org_id = get_current_organization(info)
+        # Header optional (ADR 0001 §2.6): absent ⇒ unconfined to the user's
+        # grant scopes; global holders are never confined by a stale header.
+        org_id = active_org(info)
         return shelter_queryset(queryset, user=user, organization_id=org_id, perms=[models.Shelter.perms.VIEW])
```
mikefeldberg: don't need `perms` now that `shelter_queryset` sets `VIEW` as default. same for bed, room, res

apps/betterangels-backend/shelters/schema.py
```diff
@@ -49,51 +58,86 @@
...
+from strawberry import ID, Some, UNSET
 from strawberry.types import Info
 from strawberry_django.auth.utils import get_current_user
 from strawberry_django.pagination import OffsetPaginated
 
 
+P = TypeVar("P")
+
+
+def _build_payload(
```
mikefeldberg: seems like this should live in `common`. I know we're only using this in shelter atm, but I assume we'll adopt the dataclass model everywhere

mikefeldberg (approved):

PR #2414, opened 2026-09-02
feat(perm): frontend reachability — grants-based org list + global permissions (DEV-2557)

vecchp:
## Summary

**Frontend reachability** (ADR 0001 phase 3, finding F24) — the shared contract the frontend gates on. This is PR 6 of the grant-authorization stack: deltas 1–2 of the ADR §5.2 refinement (§7 item 7), stacked on #2413.

### The shipped contract

- **`currentUser.permissions`** — the GLOBAL permission list, **bounded server-side to the product-modeled catalog** (the registry the FE `PermissionEnum` is generated from): superuser → every modeled permission; otherwise `user_permissions` ∪ permissions from global Roles in `user.groups`, also bounded to the modeled set. Admin-internal DB permissions (`auth.*`, `accounts.view_user`, …) never ship to the client. Scoped (grant) permissions are per-org and are **not** in this list.
- **`currentUser.organizationsOrganization`** — the FINITE org-scoped list: membership ∪ direct user grants ∪ delegated orgs (ADR §2.4). Deliberately **never** expanded to every org for a global holder — a global user's cross-org reach is unscoped reads (`visible()` never confines a global holder) + `currentUser.permissions`, not an enumeration of the platform. There is **no "All" provider mode** (rejected in review; see ADR §5.2 / §7 item 7).
- **Per-org `permissions`** — now EFFECTIVE: the org-scoped union (legacy group perms, grant role perms, delegated org→org grants with the ceiling) **plus** the global tier folded in server-side **only where the backend enforces it at an org** (grant-only/dual domains — finding H2). An org entry is the complete "what can I do fully here" answer, and the FE org-scoped gate is a **single membership test on the active org's entry** — no client-side union with the global list (finding H2).

### Backend

- `accounts/types.py` — `CurrentUserType.permissions` (global list); `organizationsOrganization` is the finite grants-based list; per-org `permissions` is the effective report.
- `accounts/selectors.py` — `organization_permissions()` (scoped-only base) + `organization_effective_permissions()` (domain-aware global fold, memoized per request).
- `common/permissions/selectors.py` — `global_permissions()` (modeled-bound, via new `modeled_permission_strings()`), `switchable_orgs()`; memo invalidation extended to the new memos; the dead `reachable_orgs()`/`global_holder()` all-orgs expansion removed.
- `common/permissions/domain.py` — single-source domain-state matrix (`LEGACY_INERT_APPS` / `DUAL_APPS` / `GLOBAL_TIER_ORG_APPS`), shared by the report and the equivalence tests.
- `schema.graphql` regenerated (also carries PR 4's shelter-mutation changes that were never written to the checked-in file).

### Frontend (`libs/ba-platform`)

- `CurrentOrgUser` query + `CurrentUser` type carry the global `permissions` list (transported for future global-tier/non-org gating; no consumer in this PR).
- `useActiveOrgState` gates org-scoped surfaces on the active org's **effective** entry only (H2); org `permissions` are raw backend strings filtered through `isPermission`, so an unmodeled backend permission can never satisfy a gate or leak to consumers.
- `currentUserProvider` / `createUserProvider` map the global list and org lists through `isPermission` at the boundary.
- Generated types hand-synced (no codegen in the dev container) and kept in sync with `schema.graphql`.
- New tests: finite org list + effective entries for global/superuser holders, org-scoped gate equivalence (entry ≡ `can()`), `isPermission` boundary, provider mapping.

### Validation

- Backend: `common/` + `accounts/` + `shelters/` suites green locally; ruff + mypy clean (CI typecheck green after the memo-annotation fix in this update).
- Frontend: `ba-platform` vitest suite + TS hand-verified (no `node_modules` in the dev container); full CI test run follows on push.

## Stack

1. #2409 — models + checks
2. #2410 — provisioning + backfill
3. #2411 — selectors + write services
4. #2412 — shelter cutover
5. #2413 — org→org delegation
6. **this PR — frontend reachability (deltas 1–2 of §5.2)**
7. object grants + clients/notes cutover
8. teardown

Next delta (ADR §7 item 7): org-as-variable transport + header retirement; delta-3 operator-read decoupling.


vecchp:
### Design question: should per-org `permissions` be *effective* (global folded in)?

This lands the tiers 1–2 surface cleanly — I especially like that the org list is already "every org you can act in" (`reachable_orgs`, including all orgs for global holders) and that both channels are `isPermission`-sanitized before they can gate anything.

One design question before this hardens into the long-term FE contract: **today the wire carries two channels plus a composition rule the FE must implement.**

- `currentUser.permissions` = global tier (superuser → all; global Roles + `user_permissions`)
- `organizationsOrganization[].permissions` = org-scoped only (grants + delegations + legacy per-org roles); global deliberately excluded
- FE gate: `activeOrg.permissions.has(p) || globalPermissions.has(p)` — plus docstrings explaining why the org list "excludes global"

The practical consequence: for a global holder (e.g. GSO), every org's `permissions` is empty unless they also hold scoped grants there, and the whole org-context gate depends on the FE re-deriving the union on every check. That's a small but real authority-computation surface on the client — and it's the thing ADR §5.2's "FE gates on capabilities, not raw grants" (F24) argues against. The resolved capability per context should arrive server-side.

**Proposal to consider:** make `organizationsOrganization[].permissions` **effective** = global ∪ org-scoped(this org). Then:

- the org-context gate collapses to one membership test on the active org — no union rule, no "global excluded" special case;
- a GSO simply sees their perms listed in every reachable org, which is exactly what the backend `can(user, perm, org)` returns;
- the org-scoped-only report ("what does *this org* grant, not counting global") stops riding the gate path and moves to the §5.3 member/role surface (org admin view = members + grants), which is where that question gets asked anyway.

Tradeoffs, stated honestly:

- **Payload:** global perms get duplicated across reachable orgs — dozens of short strings, negligible.
- **Lose the org-scoped-only read on this field:** but the only consumer today is this resolver plus its tests — nothing else reads `organization_permissions` (I verified).
- **Change surface is small:** one resolver + tests; both inputs (`_org_permissions` report and `global_permissions(user)`) are already memoized per request.

Separately, to preempt a direction I don't think we want: this is *not* an argument for sending roles or an "assume-role" model to the FE. Roles stay server-side policy; the client should receive resolved capabilities per context.

**Questions:**
1. Is there a consumer you know of that needs the org-scoped-only semantics on this field (needs to distinguish "granted by this org" from "effective here")? If so, where?
2. Is there a real org-less (tier-1) surface today that gates on `currentUser.permissions` alone? If not, that list currently exists mainly to feed the union.
3. If we go effective — land it here while the contract is young, or defer until the §5.3 org-admin milestone?

Happy to make the change if you agree, or drop it if you see a consumer I'm missing.

vecchp:
### Follow-up: the shape this converges on (sketch)

If we go effective-per-org, here's the concrete end-state contract — three capability contexts, each server-resolved, so the FE never computes authority (no unions, no "excluded global" rules, no role→permission derivation).

**Wire contract**

```graphql
# Tier 1 — org-less context
type CurrentUserType {
  permissions: [String!]!                 # global tier ONLY — shrinks to org-less features
  organizationsOrganization: [CurrentUserOrganizationType!]
}

# Tier 2 — org context (effective)
type CurrentUserOrganizationType {
  id: ID
  name: String
  permissions: [String!]!                 # EFFECTIVE = global ∪ org-scoped(this org)
  isMember: Boolean!                      # display metadata ONLY — "joined" vs "reachable via global"
  roleNames: [String!]!                   # display ONLY — never a gate input
}

# Tier 3 — record context (per whitelisted model; ships with the clients cutover)
type ClientProfileType {
  canChange: Boolean!                     # can_obj(user, perms.CHANGE, self) — batched Exists
  canDelete: Boolean!                     # can_obj(user, perms.DELETE, self)
}
# generic can(permission) + currentUser.objectGrants stay rejected (§5.2)
```

**Payloads**

Alice (org-scoped only):
```json
{
  "permissions": [],
  "organizations": [
    { "id": "A", "permissions": ["organizations.add_org_member", "teams.change_team"], "isMember": true,  "roleNames": ["Org Admin"] },
    { "id": "B", "permissions": ["shelters.view_shelter"],                             "isMember": true,  "roleNames": ["Shelter Operator"] }
  ]
}
```

Sam (GSO): global folded into every reachable org; `isMember` still lets the switcher say *why*:
```json
{
  "permissions": ["shelters.add_shelter", "..."],
  "organizations": [
    { "id": "A", "permissions": ["shelters.add_shelter", "...", "shelters.view_shelter"], "isMember": false, "roleNames": [] },
    { "id": "B", "permissions": ["shelters.add_shelter", "...", "shelters.view_shelter"], "isMember": true,  "roleNames": ["Shelter Operator"] }
  ]
}
```

**FE usage — one rule per context**

```tsx
// Tier 2 — org-scoped actions (the gate)
const { can } = useActiveOrg();
{can(PermissionEnum.ADD_ORG_MEMBER) && <AddMemberButton />}

// Tier 1 — org-less features only
const { permissions } = useUser();
{permissions?.includes(PermissionEnum.ACCESS_HMIS) && <HmisLink />}

// Tier 3 — the row already knows; no hook, no list lookup
{row.canChange && <EditButton onClick={...} />}
```

**Delta vs this PR**

| Piece | Status |
|---|---|
| Tier-1 `currentUser.permissions` | ✅ already here (unchanged) |
| Org list = every reachable org (`reachable_orgs`) | ✅ already here |
| Tier-2 org list **effective** | 🔶 the one flip (resolver + tests + FE drops the union) |
| `isMember` / `roleNames` display metadata | ➕ new, display-only, optional |
| Tier-3 `canChange`/`canDelete` per row | ➕ later — ships with the clients cutover |
| Org-admin "what does my org grant" | ➕ moves to the §5.3 member/role surface |

**Forward-compat guards**
1. `can_obj` must become fail-closed before tier-3 surfacing ships (it currently returns `True` for every row to any perm-holder) — otherwise `canChange: true` on rows the backend refuses = the button-lies bug.
2. Never re-attach org-scoped-only reporting to `organizationsOrganization[].permissions` — when §5.3 needs "who can do what in my org," it queries members + grants.

vecchp:
### Full design: one surface for every persona, capabilities per scope

Following up on the two comments above — this is the complete design that threads them together, with the adversarial findings folded in (resource-derived writes, header demoted to a browse convenience, cache-safe org, tenant lifecycle).

**Thesis.** Authority is resolved server-side at the resource's tenant. The FE receives resolved capability lists per *scope* (org or platform) and never computes authority. One GraphQL surface serves the scoped operator and the GSO because org is a target/argument on the operation — never a role branch.

#### 1. Backend authority model (the core — unchanged)

- Scoped roles → `Grant` rows (`role × scope_org`); global roles → `user.groups` (global tier).
- Predicate is the single source of truth: `can` / `can_anywhere` / `can_obj` / `visible(in_org=?)` / `scopes()`.
- Global tier = `scopes()==ALL` is a **deliberate, accepted blast radius** (covers future orgs; not per-org revocable) — demands hardened accounts; documented, not accidental.

#### 2. Wire contract — every scope has ONE effective list

```graphql
type CurrentUserType {
  permissions: [String!]!              # the PLATFORM scope's effective list (global tier)
  organizationsOrganization: [CurrentUserOrganizationType!]
}
type CurrentUserOrganizationType {
  id: ID
  name: String
  isMember: Boolean!                   # display only — "joined" vs "reachable via global"
  roleNames: [String!]!                # display only — never a gate input
  permissions: [String!]!              # EFFECTIVE = global ∪ org-scoped(this org)
}
```

A scope's list is the resolved answer to *"what can I do here?"* — the FE never unions, never excludes global, never derives perms from roles. A GSO sees their global perms listed in every reachable org (`isMember` distinguishes joined vs reachable); an org-scoped operator sees only their orgs' effective lists.

#### 3. Queries — same surface; org is an optional *variable*

```graphql
query ShelterList($orgId: ID) { shelters(orgId: $orgId) { id name } }
```

- `orgId="X"` → confined to X (`visible(in_org=X)`); `orgId=null` → unconfined: scoped users get the union of orgs where their grants live, global holders get everything.
- **Org as a GraphQL variable** (not only a header) makes it part of the Apollo cache key — fixes cross-org cache collisions. The `X-Organization-ID` header stays only as a transport default for `orgId`.

#### 4. Mutations — same surface; the tenant comes from the resource

**Principle:** org is resolved from the row being touched (or given explicitly for top-level creates) — never from the header, never by role branch. One write semantic everywhere.

```python
def bed_create(*, user, room_id, data):
    room = room_get(user, room_id, permission=Room.perms.VIEW)   # fail-closed parent load
    org  = room.shelter.organization                              # tenant = the resource
    require_can(user, Bed.perms.ADD, org=org)                     # one check, both personas
    return Bed.objects.create(room=room, shelter=room.shelter, **data)

def shelter_create(*, user, organization_id, data):               # top-level create: explicit org input
    require_can(user, Shelter.perms.ADD, org=organization_id)
    ...
```

- No header → no per-request org plumbing, no FE auto-switch / "context follows record" machinery.
- Cache-safe, auditable from the operation alone, curl/testable.
- Scoped user touching a foreign row: parent load 404s or `can()` fails — fail-closed. GSO: `can()` passes at every resolved org — reach, not a second code path.
- Platform-shared models keep their RFC 0002 rule (`can_anywhere` / `can_obj`) as the *named exception*, not a parallel semantic.

#### 5. Frontend — active *scope*, single `can()`

```ts
type Scope = { kind: 'org'; org: Org } | { kind: 'platform' };    // platform list = user.permissions
// can(p) = activeScope.permissions.has(p)  ← the ONLY rule; no unions
```

- **Scoped operator:** `scopes=[org X]`; switcher hidden; queries send `orgId=X`.
- **GSO:** `scopes=[Platform, org A, org Y, …]`, default last-used; Platform mode sends `orgId:null` ("All shelters" is a real, visually distinct place). Opening a shelter never mutates global state — the mutation just names the row; the backend resolves the org.
- Switching scope changes the cache key (org is a variable).

#### 6. End-to-end

- **Scoped operator (org X):** lists her org's shelters → `createBed(roomId)` → backend resolves org X → `can` ✓. A crafted `createBed` for an org-Y room 404s.
- **GSO:** browses *all* shelters (`orgId:null`) → adds a bed in shelter Y with the *same* `createBed(roomId)` → backend resolves org Y → `can` ✓ (global tier). Creates a shelter with `createShelter(organizationId:…)`.
- **Multi-org scoped user:** resource-derived writes are correct in every scope — `can()` at the row's org decides, matching the permission model.

#### 7. Tenant lifecycle (fixes the orphan escape hatch)

- **Org deletion never `SET_NULL`-promotes data to the global tier** — decide per model: block / reassign / archive.
- **Shelter create requires an org** (GSO's create form includes an org picker); existing orphans get a one-time assignment/backfill. No second write path for orphans.

#### 8. Open decisions

1. Effective-per-org wire flip (comment above) — this design's §2.
2. Platform scope as a real switcher entry + default behavior.
3. Shelter create requires org + orphan backfill (ADR §7 item 2).
4. Org-deletion semantics (block/reassign/archive).
5. GSO blast-radius acceptance (MFA, no per-org revocation).
6. Shelter-subset scoping is inexpressible today (E003) — park unless product asks.

#### 9. Migration sequence

1. Decide tenant lifecycle (3–4) — cheapest, gates the rest.
2. #2414: effective-per-org + FE single-`can()` (decision 1).
3. Shelter services → resource-derived writes (§4); document the single write semantic.
4. Demote the header: `orgId` as query variable; fix cache identity.
5. Platform scope in the shelter-operator FE (decision 2).
6. Tier-3 `can*` per-record fields ride the clients cutover, unchanged.

vecchp:
### Design reset: record-centric reach (supersedes the org-context direction above)

Stepping back from the earlier comments in this thread: when I lay out the real personas side by side, the org-context model stops holding up.

| Persona | What they do | Reach |
|---|---|---|
| Org shelter operator | shelter/bed/room CRUD | their org |
| GSO | shelter/bed/room CRUD | any/all orgs |
| Org note editor (caseworker) | note/task work on clients | their org's notes + shared clients |
| Global note editor | note/task work | every org + shared clients |

The operations are **identical across all four — only reach differs**. And org itself is a soft administrative grouping: `Shelter.organization` is nullable, clients have **no org at all**, note/client reads are deliberately cross-org. So an "active org + header + switcher + platform mode + per-org permission lists" is a state machine expressing something that was never org-shaped.

**The reset: org is a property of data, not a place the user stands.**

- **Reads:** the backend returns exactly the user's *reach* (`visible()` over their full authority — org grants + global tier + per-record grants). No context, no header. A caseworker sees their org's notes + shared clients; a GSO sees everything. "Global" is just *more rows in the same list*, never a special mode.
- **Buttons:** every list row carries a small, batched capability vocabulary — `canChange`/`canDelete`/`canManage`. Org-uniform rows are annotated in bulk (`row.org ∈ orgs-where-user-holds-CHANGE`); per-record rows (caseworker notes, shared clients) use the object-grant `Exists`. No per-org permission lists shipped to the FE at all.
- **Creates:** org is a **required input** for row-less org-owned creates (`createShelter(organizationId: …)`); the picker is populated by one small query — "orgs where I can ADD" — which is also the button's gate, so the button and the picker can't disagree.
- **Row-scoped mutations:** org comes from the resource (`updateShelter(id)` → `shelter.organization` → `can()`; `updateNote(id)` → note org + per-record grant → `can_obj()`). No org parameter, no header, one check serves every persona.
- **Org-admin screens** (members/teams/reports): org is a **page parameter** (`orgId` variable), and the screen's booleans (`canManageMembers`, …) are resolved server-side for that one org.

```graphql
type CurrentUserType {
  permissions: [String!]!        # global tier ONLY — org-less actions (nav, HMIS, superuser)
  organizations: [UserOrg!]!     # metadata for filters/pickers — NO permission lists
}
type UserOrg { id: ID name: String! isMember: Boolean! roleNames: [String!]! }

query Workspace($orgFilter: ID) {
  shelters(orgFilter: $orgFilter) {          # orgFilter is a FILTER (variable → cache-safe)
    id name organization { id name }
    canChange canDelete canManage
    rooms { id canChange beds { id canChange } }
  }
}
```

**What this deletes:** the `X-Organization-ID` header (and the Apollo cache collision it causes — org is never ambient, so it's always in the cache key or the row), the active-org provider/switcher-as-authority, "platform mode," `organizationsOrganization[].permissions`, the FE union rule, and the strict-vs-union dual write semantic. One rule remains: **authorize against the resource.**

**Costs, stated honestly:** GSO lists are large (pagination + org-as-filter handle it); batched capability fields need a small whitelisted vocabulary per model; and the migration is the real price — every current header/org-context consumer (teams, notes, reports, member management) moves domain by domain.

**Open decisions before building:**
1. Note-create owner-org when the creator spans orgs (or is global) — the one create where org choice is load-bearing (decides edit reach + reporting). Picker vs. creator's home org?
2. "Home org" as a remembered client-side preference for creates/filters (UX only, never authority).
3. Shelter create requires `organizationId` (no orphan minting) + org-deletion semantics (never `SET_NULL`-promote to the global tier).
4. Per-row vocabulary stays small and whitelisted (`canChange`/`canDelete`/`canManage`) — no generic `can(permission)` field.

This supersedes the effective-per-org / platform-scope direction I posted earlier in this thread — that was still org-centric. Happy to pressure-test or adjust before this becomes a real proposal.

vecchp:
### Permission surface — where we landed + required steps

I've consolidated the whole discussion (earlier comments removed) into one artifact:
**ADR 0001 §5.2 / §7 item 7** in this PR (`docs/adr/0001-grant-based-authorization.md`). It supersedes all intermediate
sketches.

**Where we landed — refinements to this PR, not a redesign.** The grant-authorization
core (roles/grants/global tier/predicate/delegation/per-org report) is right and
mostly already here. The proposal narrows to three surgical deltas:

1. **Effective per-org lists** — `organizationsOrganization[].permissions` folds the
   global tier in server-side, so an org's entry is the complete "what can I do fully
   here" answer (a GSO who's also a member/delegated sees global ∪ their grants; a
   delegated org shows its ceiling). One membership test in the FE; no client union.
2. **Finite org list + "All" mode** — the switcher is member ∪ delegated orgs only
   (no all-orgs enumeration for global holders); global users default to an All mode
   gated by `currentUser.permissions` and drill into org views from there.
3. **Org as a query variable / route param** — org becomes part of the cache key for
   reads instead of living only in the `X-Organization-ID` header (fixes the Apollo
   cache collision). Mutations keep the active org as their target; backend `can()`
   is always the authority.

Non-admin users are effectively one org at a time, so their `can(p)` degenerates to
the active org's list ("given the org I'm in, can I add a shelter?"). Switching orgs
updates data + UI from the new entry. Delegation is already handled by the report
(ceiling applied). Object-level (per-record) permissions and impersonation are
**deferred but reserved** (per-row `can*` flags as the extension point; the org report
is already principal-parameterized).

**Required steps forward**
1. In this PR (small): effective per-org lists (resolver + tests); finite org list +
   FE All mode.
2. Per domain (mechanical): org as query variable / route param for reads; retire the
   header as the only org signal.
3. FE apps: migrate admin + shelter-operator (+ mobile) onto the effective lists and
   All/org modes; drop the client union.
4. Later: object-level surfacing, impersonation, org-deletion semantics (never
   `SET_NULL`-promote data to the global tier).

Details, actor shapes, the wire contract, deltas, and invariants are in the doc —
happy to discuss any of it.

docs/adr/0001-grant-based-authorization.md
```diff
@@ -819,9 +819,12 @@ The FE gates features on **capabilities, not raw grants** (finding F24). Three t
...
 1. **Global** — `currentUser.permissions`: the global tier (global Role perms +
    `user_permissions`; superuser → every permission). Ships with phase 3 (PR #2414).
-2. **Per-org** — `currentUser.organizationsOrganization[].permissions`: the union of
-   legacy group perms and Grant role perms (including delegated org→org grants) at each
-   reachable org. Ships with phase 3 (PR #2414).
+2. **Per-org** — `currentUser.organizationsOrganization[].permissions`: the
+   ORG-SCOPED union per reachable org (legacy group perms, Grant role perms, and
+   delegated org→org grants with the ceiling applied) — the global tier is reported
+   once in `currentUser.permissions`, and the FE gate composes the union. Ships with
+   phase 3 (PR #2414). Whether the per-org lists should instead be *effective*
```
vecchp: **H1 — this ADR bullet contradicts the code the PR ships.** It still describes per-org `permissions` as the ORG-SCOPED union with the FE composing `global || org`, and marks *effective* as "open" — but the implementation (`organization_effective_permissions` + `CurrentUserOrganizationType.permissions`) ships **effective** entries (global folded in server-side), and §7.7 below still tags the whole shape "[open — under review in PR #2414]". Later cutover PRs (#2427–2429, mobile/admin) read this ADR as ground truth, so it can't be internally contradictory at merge time. Suggest updating this tier-2 bullet *and* §7.7 to the decided contract (effective entries + finite switchable org list) in the same commit that implements it.

apps/betterangels-backend/accounts/selectors.py
```diff
@@ -147,3 +147,134 @@ def role_names_by_organization(*, user_id: int) -> dict[str, list[str]]:
...
+        return cached
+
+    org_ids = set(switchable_orgs(user).values_list("pk", flat=True))
+    global_list = global_permissions(user)  # memoized inside the selector
+    if user.is_superuser:
+        result = {org_id: global_list for org_id in org_ids}
+    else:
+        scoped = organization_permissions(user, org_ids=org_ids)
+        result = {
+            org_id: sorted(set(global_list) | set(scoped.get(org_id, []))) for org_id in org_ids
```
vecchp: **H2 — the effective fold is not domain-aware; "global holder × legacy-only domain" is an untested over-gate.** Folding the whole `global_permissions(user)` into every org entry assumes global authority is enforceable at every org. That holds for grant-only domains (shelters — `scopes()==ALL` ⇒ `can()` true anywhere) but not legacy-only domains (member management, reports, teams), whose enforcement is `permissioned_queryset`/`@hasOrgPerm`: I checked `common/permissions/utils.py` — it consults **only org `PermissionGroup`s**, never `user_permissions`/global roles. Concretely: a member of org X holding `reports.view_reports` as a direct `user_permission` would get it advertised in X's entry and in `currentUser.permissions` → FE shows the control → the legacy gate at X denies. The equivalence suites never exercise this cross product (`CurrentUserReportCanEquivalenceTestCase` compares to `can()` using grant-only shelter perms only; `CurrentUserLegacyDomainReportEquivalenceTestCase` uses only org-scoped holders). Suggest making the fold domain-aware, or adding a "global holder × legacy-only-domain perm" equivalence test against `permissioned_queryset`.

apps/betterangels-backend/accounts/selectors.py
```diff
@@ -147,3 +147,134 @@ def role_names_by_organization(*, user_id: int) -> dict[str, list[str]]:
...
+        }
+    user.__dict__["_org_effective_permissions"] = result
```
vecchp, line 292: **M2 — the new memo caches have no invalidation hook.** Unlike `_scope_cache` (which grant write services pop via `invalidate_scope_cache`), `_org_effective_permissions` here — and `_global_permissions`/`_global_holder` in `common/permissions/selectors.py` — are never invalidated. A request that grants/revokes/role-assigns on the same user instance and then resolves the report mid-request serves a stale entry (under-gates the same response; the org *list* is derived live, so list-vs-perms can disagree mid-request). Cheap fix: pop these keys in `invalidate_scope_cache`, or document the request-scope/no-mid-request-mutation invariant.

apps/betterangels-backend/common/permissions/selectors.py
```diff
@@ -60,6 +61,134 @@ def _roles_carrying_perm(perm: str) -> "QuerySet":
...
+    if global_holder(user):
+        return cast("QuerySet[Organization]", Organization.objects.all())
+    return _finite_org_scope(user)
+
+
+def switchable_orgs(user: "User") -> "QuerySet[Organization]":
+    """The FE org list / switcher (ADR 0001 §5.2 refinement, §7 item 7).
+
+    Member ∪ direct-grant ∪ delegated orgs — the orgs the user can switch to in
+    the UI.  Deliberately NEVER expanded to every org for a global holder: a
```
vecchp, line 134: **M1 — the finite org list ships, but the "All mode" that makes it usable for global holders does not (in this PR).** `switchable_orgs` is deliberately never expanded for a global holder, and the FE side only adds the global list to `hasPermission` — there is no "All" org entry, no org-less flow, and no "enter org by id" resolver added here. So a superuser/GSO with no org membership gets an empty switcher and no path into any org, even though `can()`/`visible()` give them global reach (and `visible()` is never org-confined for `ALL`). ADR §7.7 lists "effective lists + finite org list + All mode in #2414" as the required step. Either re-scope that step to the follow-on org-variable transport, or implement at least the drill-in-by-id path here so the GSO cutover isn't blocked.

apps/betterangels-backend/common/permissions/selectors.py
```diff
@@ -60,6 +61,134 @@ def _roles_carrying_perm(perm: str) -> "QuerySet":
...
+    effective per-org entry in the same request share one lookup.
+    """
+    from django.contrib.auth.models import Permission
+
+    cached: Optional[list[str]] = user.__dict__.get("_global_permissions")
+    if cached is not None:
+        return cached
+
+    if user.is_superuser:
+        perms = Permission.objects.all().values_list("content_type__app_label", "codename")
```
vecchp: **M4 — superuser "every permission" is the whole `Permission` table, then repeated per org.** `Permission.objects.all()` spans every content type in the project (including django-admin/logentry and apps this client doesn't use), and `organization_effective_permissions` copies that list into **every** switchable org entry *plus* `currentUser.permissions`. `isPermission` trims it client-side, but the wire payload multiplies per org, and non-ba-platform clients (mobile) that don't filter get the raw list. Consider trimming server-side to modeled/enforceable perms, or at least documenting the cost.

libs/ba-platform/src/lib/react/providers/activeOrg/useActiveOrgState.ts
```diff
@@ -8,11 +9,19 @@ import {
...
+ * filters them to the modeled ``PermissionEnum`` set before exposing
+ * ``activeOrg``/``organizations`` (see ``useActiveOrgState``), so an unknown
+ * backend permission can never satisfy a check or leak to consumers that read
+ * ``permissions`` directly.
+ */
 export interface Org {
   id: string;
   name: string;
-  permissions: readonly PermissionEnum[];
+  permissions: readonly string[];
```
vecchp, line 24: **M3 — type + semantic change whose consumers are outside this PR.** `Org.permissions` goes `PermissionEnum[] → string[]`, and the field semantics change: org list membership-only → member ∪ grant ∪ delegated, per-org perms legacy-only → effective global∪scoped. Consumers not touched here include the expo HMIS app's own `UserProvider`/generated types (still typed to the old shape — only the shared `schema.ts` string was regenerated) and the shelter-web/admin org switchers that now receive non-member grant/delegated orgs. This PR was only TS-hand-verified (no node_modules). Please get a full-repo FE typecheck/test signal across consuming apps before merge, and enumerate the intended consumers of the expanded list.

libs/ba-platform/permissions/src/index.ts
```diff
@@ -1 +1,29 @@
...
+
+// The generated module exports only the ``as const`` permission objects (types
+// are erased at runtime), so every runtime export is an object of permission
+// strings.  This is the runtime mirror of the ``PermissionEnum`` union.
+const KNOWN_PERMISSIONS: ReadonlySet<string> = new Set(
+  (
+    Object.values(PermissionGroups) as unknown as ReadonlyArray<
+      Readonly<Record<string, string>>
+    >
+  ).flatMap((group) => Object.values(group)),
```
vecchp: Nit — `KNOWN_PERMISSIONS` is built by flattening **every runtime export** of the generated module. That's safe while codegen emits only `as const` permission objects, but any future non-permission export (string enum, constant, array) silently widens the gateable set and defeats the F9 boundary. A shape assertion (each value matches `^[a-z_]+\.[a-z0-9_]+$`) or a test pinning the set would harden it.

vecchp (commented):
## Adversarial review — frontend reachability (PR 6)

Reviewed at head `f5d52246` vs `main`. I read the head-state files, cross-checked the enforcement predicates (`scopes`/`can`/`visible` and the legacy `permissioned_queryset`/`@hasOrgPerm`), the `Grant`/`Role`/`PermissionGroup` models, the generated permission module behind the FE boundary, and every non-generated FE file. I could not execute the suites from here (the checked-out worktree is on a descendant branch), so runtime claims rest on the PR's own validation plus code reading.

Overall: this is careful, well-tested work — the report is never an enforcement input (no escalation surface), the scoped report's arms are meticulously matched to `scopes()`, query counts are bounded/flat, and the FE boundary filtering (F9) is a genuinely good idea. The findings below are contract, drift, and completeness defects rather than data-exposure holes. Inline comments cover the specifics; this is the summary.

### 🔴 High

**H1 — ADR vs code contradiction on the current contract.** The code ships *effective* per-org entries and a *finite switchable* org list, but the ADR tier-2 bullet (see inline) still describes ORG-SCOPED entries with the FE composing the union and marks "effective" as open, and §7.7 still tags the whole tier-1/tier-2 shape "[open — under review in PR #2414]". Later cutovers (#2427–2429, mobile/admin) read this ADR as ground truth — reconcile the contract section + flip §7.7 to "decided" before merge.

**H2 — the effective fold is not domain-aware (global × legacy-only-domain over-gate).** Folding the entire global list into every org entry assumes global authority is enforceable at every org. True for grant-only domains (shelters), false for legacy-only domains (member management/reports/teams), which are enforced by `permissioned_queryset` — it consults only org `PermissionGroup`s, never `user_permissions`/global roles. The equivalence suites never test the "global holder × legacy-only-domain" cross product; fixtures deliberately pick domains where global == enforceable.

### 🟠 Medium

**M1 — "All mode" is documented as shipping here but isn't implemented.** Finite switchable list + no "All" entry / org-less flow / "enter by id" resolver means a superuser/GSO with no org membership gets an empty switcher and no path into any org, despite `can()`/`visible()` granting global reach. Re-scope the ADR step to the follow-on PR or implement the entry path here.

**M2 — new memo caches lack an invalidation hook.** `_org_effective_permissions` / `_global_permissions` / `_global_holder` are never invalidated (unlike `_scope_cache`), so grant/role writes on the same user instance mid-request serve stale reports.

**M3 — behavior-change blast radius beyond this PR.** Org list (membership → member ∪ grant ∪ delegated) and per-org `permissions` (legacy → effective global∪scoped) change meaning for consumers not touched here (expo HMIS app's own `UserProvider`, shelter-web/admin switchers); `Org.permissions` type changes `PermissionEnum[] → string[]`. Needs a full-repo FE typecheck (only hand-verified here).

**M4 — superuser payload.** `Permission.objects.all()` (every content type) is repeated into every switchable org entry + `currentUser.permissions`; FE trims it but the wire payload multiplies per org and non-ba-platform clients get the raw list.

### 🟡 Nits
- `isPermission` flattens every runtime export of the generated module — widen-safe only while codegen emits solely permission objects (see inline).
- Doc drift: "five queries" vs "four-query scoped report + two-query global probe".
- `organization_permissions(org_ids=None)` expands `reachable_orgs()` = every org for a global holder — latent perf foot-gun for future callers.

### Bottom line
Semantics and tests are the strongest part of this stack. Blockers before merge: **H1** and **H2**; strongly consider **M1/M3/M2**. Happy to draft the ADR edits, the missing equivalence test, or the invalidation patch.

vecchp:
## Resolution of adversarial-review findings

Follow-up on the review (all findings landed on the follow-on branch `perm/org-variable-transport`, which sits directly on this PR's head `f5d52246`):

- **H1 — ADR vs code contradiction**: resolved in `7849afcba` — ADR §5.2/§7 item 7 marked **decided** (effective per-org lists + finite member∪delegated org list, shipped in #2414/#2440); the "All" provider mode was explicitly rejected and documented.
- **H2 — domain-unaware effective fold / FE union over-gate**: fixed in `00c64d0f` + `b05c7b6d` —
  - Backend: `organization_effective_permissions` now folds the global tier **domain-aware** (`GLOBAL_TIER_ORG_APPS` = grant-only ∪ dual in `common/permissions/domain.py`). Legacy-only domains (member mgmt / reports / teams, enforced by `HasOrgPerm` → org `PermissionGroup` rows) only reach org entries via the legacy arm — no more superuser/user_permission over-claim at orgs where the user has no group. Superuser short-circuit removed so org-group perms still surface. New equivalence tests pin the global-tier × legacy-only cross product.
  - FE: `hasPermission` gates on the **active org's effective entry only**; the global-list union (and `globalPermissions` plumbing) removed — global-tier gating is the user provider's job via `currentUser.permissions`.
- **M1 — All-mode gap**: resolved by decision in `7849afcba` (no provider "All" mode; global users' cross-org surfaces are unscoped views — delta 3).
- **M2 — memo invalidation**: fixed in `13698b5d` — `invalidate_scope_cache` now also drops `_global_permissions` / `_global_holder` / `_org_effective_permissions`.
- **M4 — superuser payload**: mitigated by H2 (org entries no longer repeat the full global list per org); `currentUser.permissions` remains one full list by design, trimmed client-side by `isPermission`.
- **Nits** (doc-count drift, `organization_permissions` all-orgs default foot-gun, `isPermission` flatten fragility): fixed in `13698b5d`.
- **M3 — consumer blast radius**: outstanding as an audit item — expo HMIS generated types are stale w.r.t. `currentUser.permissions` (regenerate when that app next touches the user query) and the cross-app FE typecheck should run in CI before merge.

Validation: backend 56 tests (reachability + queries + delegation) and ba-platform 72 FE tests pass; ruff/format clean.

_Note: these commits are on `perm/org-variable-transport`, not on this PR's branch — the code will reach `main` when that follow-on merges. Happy to cherry-pick onto `perm/grant-reachability` if you'd rather the fixes ride in this PR._

vecchp:
Update: the review fixes (H2 backend, H2 frontend, M2/nits) have now been **added to this PR's branch** as well — `perm/grant-reachability` is at `810a2a4d` (cherry-picked from the follow-on `perm/org-variable-transport`, content-identical and re-validated: backend 56 tests + ba-platform 72 tests green). H1/M1's ADR decision remains on the follow-on (`7849afcba`), since it also carries delta-3 content.

mikefeldberg:
## Issues

### F1 [Med] — Grant arms are not domain-filtered → report over-claims `clients.*` for scoped holders
`organization_permissions` filters only the **legacy** arm by `LEGACY_INERT_APPS`. The **direct-grant and delegated arms union every permission of the user's grants with no domain check**. Only `shelters` is grant-authoritative today, yet the scoped Shelter Operator role carries `clients.view_clientprofile` (`shelters/groups.py:62`), and clients enforcement is still `HasPerm`/`HasRetvalPerm` + guardian — i.e. `has_perm`, which **Grant rows never feed**. After the shelter teardown a scoped Shelter Operator has no group, so the backend will deny `clients.view_clientprofile`, while this report claims it at every org where they hold the role — precisely the "report ≠ enforcement" over-claim the PR's own docstring promises never to make. The equivalence tests only cover shelters (`test_reported_shelter_perms_are_enforceable`) and ORG_ADMIN legacy — no test for a grant-role perm whose app isn't grant-only.

Fix: restrict the grant arms to apps in `GLOBAL_TIER_ORG_APPS` (mirroring the legacy arm), and add an equivalence test for a scoped-role perm on a non-grant-only domain (`clients.view_clientprofile` is the live case).

### F2 [Med] — The domain taxonomy collapses two enforcement classes into "legacy-only," so the global fold is wrong for `has_perm`-gated domains
`GLOBAL_TIER_ORG_APPS = LEGACY_INERT_APPS ∪ DUAL_APPS = {shelters}` asserts the global tier is org-enforceable *only* for grant-only/dual domains. That's accurate for `HasOrgPerm` domains (member mgmt, reports, teams — verified: `extensions.py` goes through `permissioned_queryset`, no superuser bypass). But `clients`/`notes`/`tasks`/`referrals` are gated by `HasPerm`/`HasRetvalPerm` → `user.has_perm`, which **does** honor the global tier (superuser, global roles, `user_permissions`). Consequences:
- A superuser who can genuinely read any `ClientProfile` gets no `clients.*` in any org entry → **under-claim** (the mirror-image bug the fold was built to avoid, just in the safe direction).
- The same domain's perms *are* claimable through the legacy arm when the user happens to be in an org group — so the same permission's presence in an entry depends on which arm supplied it, not on whether it's enforceable.

`test_superuser_org_entry_is_group_scoped_for_legacy_only_domains` only pins `reports`/`teams`/member perms — it would fail (or the code is wrong) for `clients`/`notes`. The module needs three enforcement classes (`HasOrgPerm`-group, `has_perm`-global, grant), or the fold logic needs a per-app statement of "which tier does enforcement honor at an org," rather than the binary grant-only/legacy split.

### F3 [Med, sequencing] — FE `hasPermission`/org-list semantics flip is a breaking contract change
`hasPermission` stops consulting the global list and the org list stops expanding to every org for global holders. Both are correct end-states, but they invert what the previously-merged intermediate shipped, so every consumer app (admin, shelter-operator, mobile) must land "All" mode + org-by-id entry + global-list gating in the **same** release or GSO/global holders lose org switching and org-scoped gates. The ADR lists app migration as a *later required step* — landing the lib flip first creates a window of broken gating. Also: a persisted `activeOrgId` pointing at an org that dropped out of the finite list leaves `activeOrg` undefined and silently disables all org-scoped gates — default-to-All/first-org handling isn't in the diff.

### F4 [Low] — Reachability and report duplicate the delegation predicate and can drift
`_finite_org_scope`'s delegated arm (role-level "shares ≥1 permission," membership via `grants__scope_org__users`) and `organization_permissions`' delegated arm (permission-level `perm ∈ held_in_org[B]`, membership via `member_org_ids`) are *equivalent today* (I walked the cases: reachable ⇒ ≥1 shared perm ⇒ that perm is reported), but they're two hand-written formulations of the same C-1 rule in two files. That duplication is exactly where a future edit (e.g. changing "acts at B" to include delegated-into-B holders, or adding role mapping) desynchronizes the org list from the per-org report. Extract one shared predicate.

### F5 [Low] — Object grants are scanned by the direct-grant arm
`organization_permissions` iterates `Grant.objects.filter(principal_user=user)` with no `scope_org__isnull=False`, relying on the `None`-keyed rows being dropped. The `switchable`/`reachable` arms filter correctly. Harmless now, but wasteful once object grants exist and inconsistent with the sibling arms — add the filter for symmetry.

### F6 [Low] — New memo caches widen the same-request staleness surface
`_global_permissions`/`_global_holder`/`_org_effective_permissions` join `_scope_cache` on `user.__dict__`, and only `invalidate_scope_cache` clears them. That helper must now be called by **every** authority write in a request that then re-reads authority — including legacy role add/remove via `OrgRoleManager` (group membership changes don't clear `_global_permissions`/`_global_holder`). Audit callers; if any mutation grants/revokes and then renders a member/org view on the same user instance, the memoized `ALL`/global list is stale.

### F7 [Low, residual] — `currentUser.permissions` still advertises legacy-only perms
The org entries correctly suppress `reports.view_reports` etc. for a groupless superuser, but the **global** list (tier 1) still contains them (asserted in the tests). Any FE that gates a `HasOrgPerm`-domain control on the global list reproduces the H2 bug one level up. The FE discipline ("legacy-domain UI gates only on org entries") is now load-bearing and undocumented outside comments — worth a written FE contract in the RFC/merge guide.

mikefeldberg (approved):

Issue #2417, opened 2026-09-02
Grant system follow-ups from adversarial review of #2409–#2416


vecchp:
Adversarial review of the grant-authorization stack (#2409–#2416) surfaced the medium/low items below. The high-severity one (admin role editor silently revoking grant-held roles) is fixed in #2416.

## Medium

1. **`grant_create`/`grant_delegate` don't refuse global roles at write time.** Only `grant_obj` and `role_assign` guard; a global-role Grant can be written via service or `GrantAdmin` (no `formfield_for_foreignkey` filter) and is caught only by deploy-time check `permissions.E002`. Fix: `if role.is_global: raise ValidationError(...)` in both services + restrict `GrantAdmin.role` choices to `is_global=False`.

2. **ADR §2.5 vs code drift on `OBJECT_ARM_ENABLED`.** ADR says the object arm stays `False` until the clients cutover ("turned on with its first consumer, not before"); PR 7 shipped it `True` and whitelisted `ClientProfile` while §5.1 (client ownership model) is still open (§7.6). Reconcile: either flip the flag back to `False` or update the ADR.

3. **Header-optional convention applied to only `create_shelter`.** 17 shelter mutations still hard-require `X-Organization-ID` via `get_current_organization`; a global-tier holder cannot create/update rooms, beds, reservations without a header — inconsistent with ADR §2.1 rule 3 ("Authority never requires the header") and §2.6's "explicit `organization_id` input". Extend the optional-header pattern or document as deferred (frontend phase 3).

## Low

4. **Dead code:** `shelters/selectors/operator.py::user_shelter_list` has no callers after the PR-4 cutover (only re-exported). Remove.

5. **`GrantAdmin` writes directly** (plain ModelAdmin) rather than through `grant_create`/`grant_delegate` — no `full_clean()`; DB constraints cover the exactly-one rules, but service-level guards are bypassed. Route through services or add admin guards.

6. **`remove_roles`/`member_roles_replace` don't clean stale legacy memberships for role-backed templates** — mostly theoretical post-deploy (deploy reconcile deletes stale Shelter Operator groups), but worth a comment or cleanup.

## Design question (confirm, not a bug)

Direct `Grant` to a **non-member** confers full org authority without membership (delegation deliberately requires membership; direct grants don't). Confirm "grants are the sole authority" is intended and document it.

vecchp:
**FE capability surfacing (from user queue-up) — already done, plus one gap to design:**

[...]

PR #2416, opened 2026-09-02
feat(perm): teardown — Shelter Operator is grant-only (PR 8)

vecchp:
## Summary

Final PR of the grant-based authorization stack (ADR 0001 §4 **phase 5 — teardown**). Role-backed templates — starting with **Shelter Operator** — stop dual-writing the legacy `PermissionGroup` membership. The `Grant` row is now the **only** scoped-authorization input for the migrated shelter domain.

**Stack:** #2409 → #2410 → #2411 → #2412 → #2413 → #2414 → #2415 → **this PR**

### Source
- **`OrgRoleManager`** (`accounts/role_manager.py`): `add_roles`/`remove_roles` for a role-backed template write/delete **only** a `Grant` — no more `user.groups` dual-write. Non-role-backed templates (CASEWORKER, ORG_ADMIN, …) keep the legacy group path until their domain cutover.
- **`reconcile_org_groups`** (`accounts/services.py`):
  - Skips *creating* legacy rows for role-backed templates.
  - Deletes **stale** pre-teardown Shelter Operator rows.
  - Revokes `Grant`s for role-backed templates the org no longer offers — the grant-world equivalent of the legacy derived-group deletion, so dropping the shelter org type actually removes the role.
- **Admin role-loss confirmation** (`accounts/admin.py`): `_template_holders` counts holders across **both** authorities (Grant ∪ legacy group), so the "Revoke roles in {org}?" prompt still fires for role-backed roles — and confirming revokes the Grants.
- **`annotate_permission_templates`** (`accounts/annotations.py` + `types.py`): the member `permissionTemplates` list unions Grant role names (deduped in the resolver) — a Shelter Operator grant now shows up in the GraphQL field.
- **`member_add`**: the existing-member check recognizes a held Grant as "already holds the role" (no redundant re-grant).

### Tests
- `test_role_manager_grants.py` rewritten from "dual-write" to "grant-only" semantics.
- `test_admin.py`, `test_services.py`, `test_roles.py`, `test_queries.py`, `test_mutations.py` updated to assert Grant-based behavior (helpers union both authorities).
- `clients/tests/test_mutations.py`: delete query count 19 → 20 — a gap from PR 7 (object-grant orphan cleanup on `ClientProfile` delete).

**Validation:** full backend suite green — `1646 passed, 31 skipped`; `ruff format`/`ruff check` and `mypy` (strict) clean.

## Deferred (documented in ADR 0001 §5)
[...]

PR #2418, opened 2026-09-02
feat(perm): grant hygiene — write-time guards, admin role filter, ADR reconciliation (PR 9)

vecchp:
## Summary

Follow-ups from the adversarial review of the grant stack (#2409–#2416), tracked in #2417. Stacked on the teardown (#2416).

### Write-time global-role guards (medium finding 1)
- `grant_create` / `grant_delegate` now **refuse a global Role** with `ValidationError` at write time (mirrors `permissions.E002`), matching `grant_obj` and `role_assign`. Previously a global-role Grant could be written and only flagged by `manage.py check`.
- `GrantAdmin`, `GrantInline`, `DelegatedGrantInline` restrict the `role` choices to scoped Roles, so staff can't write a global-role Grant through the admin.
- The old test that **pinned** the permissive behavior (`test_grant_delegate_rejects_a_global_role`) is inverted to assert the write-time refusal; added `test_grant_create_refuses_a_global_role` and three admin role-filter tests. Both guard tests fail against the old services.

### Dead code (low finding 4)
- Removed `user_shelter_list` from `shelters/selectors/operator.py` + its re-export — no callers since the shelter cutover.

### ADR reconciliation (medium finding 2 + §7.6)
- ADR 0001 §2.5 reconciled with code: `OBJECT_ARM_ENABLED` ships **on** but is a whitelist-gated no-op for every model except `ClientProfile`, which no consumer reads through `visible()` yet. Revises the earlier finding-F9 "keep it `False`" rule.
- New **RFC 0002** — `docs/adr/0002-client-writes-ownership.md`: the §7.6 client-ownership decision request, recommending `created_by_org` (per ADR rule 4: ownership is a data property, not grant rows at creation). §7.6 now links to it.

## Validation
- Full backend suite: **1652 passed, 31 skipped, 0 failed** · ruff format/check clean · mypy (strict) clean.

## Deferred (already tracked in #2417)
- Header-optional mutation convention (only `create_shelter` follows §2.6) — for the cutover wave.
- Tier-3 per-record FE surfacing (`canChange`/`canDelete`) — ships with the clients cutover (ADR §5.2).

## Summary by Sourcery

Harden grant creation and delegation against invalid global-role assignments while aligning authorization documentation with the current object-grant rollout and client ownership decision.

Bug Fixes:
- Reject global roles when creating or delegating grants instead of allowing invalid grant records to be written.
- Prevent the admin grant forms from offering global roles as selectable grant targets.

Enhancements:
- Remove the unused user_shelter_list selector and its export.
- Reconcile ADR 0001 with the enabled, whitelist-gated object-grant arm and document the client-write ownership decision in RFC 0002.

Documentation:
- Update the grant authorization ADR and add the client-write ownership RFC reference and decision recommendation.

Tests:
- Add service and admin coverage for rejecting and filtering global roles in grants.

### DIFF
--- a/apps/betterangels-backend/accounts/admin.py
+++ b/apps/betterangels-backend/accounts/admin.py
@@ -340,6 +340,17 @@ class MemberInviteAdminMixin:
         }
 
 
+def _scoped_role_queryset() -> Any:
+    """Grant forms must not offer global Roles.
+
+    A global Role is held in ``user.groups`` (the global tier), never in a
+    Grant — ``permissions.E002`` makes a grant referencing one a deploy-time
+    error, so the admin refuses the choice up front instead of writing a row
+    that only a check will flag.
+    """
+    return Role.objects.filter(is_global=False)
+
+
 @admin.register(Role)
 class RoleAdmin(admin.ModelAdmin):
     """Code-owned roles (ADR 0001 §2.2) — read-only in the admin."""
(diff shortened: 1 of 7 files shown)


vecchp:
Closing: stale base (perm/grant-teardown chain pre-dates the main redesign); write-time guard content is superseded by the rebuilt #2415 and work already on main. Re-create off main when needed.

PR #2420, opened 2026-09-02
perf(perm): memoize scopes() decision + reconcile role lookup (PR 11)

vecchp:
## Summary

Query-optimization pass over the grant predicate and provisioning, from a performance-focused review.

### 1. `scopes()` now memoizes the **whole** decision, global tier included
`scopes()` cached the lazy org-scope union, but ran the global-tier checks (superuser / global-Role / `user_permissions` — **2 EXISTS**) **before** the cache lookup — so every `visible()` call on a page re-paid them. The whole decision — the `ALL` sentinel included — is now memoized per `(user, perm)`.

Measured on a real shelter-operator scenario:
- 3 `visible()` calls on the same perm: **10 → 6 queries**
- 3 `scopes()` calls: **6 → 2 queries**

### 2. `reconcile_org_groups` loads the scoped-Role name set once
It issued a `Role.objects.filter(is_global=False).exists()` per template per org; the role-backed name set is now loaded in a single query per reconcile (migrate-time path).

### 3. Regression tests
- `ScopesMemoizationTestCase`: repeated `scopes()` calls execute **0** extra queries; the cached `ALL` sentinel too.
- `test_room_query_with_beds_status` query count 12 → 10 — it had pinned the redundant per-call global EXISTS.

## Validation
Full `accounts`/`common`/`shelters` suite: **978 passed, 0 failed** · ruff format/check clean · mypy (strict) clean.

## Notes (reviewed, not changed)
- `can()` rebuilds the memoized scope union for a one-org membership test — kept, because the memoized union is reused and a targeted rewrite would duplicate delegation logic and risk drift.
- Per-field subquery overhead in `visible()` (~2–3 subqueries per permission) is the documented fail-closed tradeoff; the object arm is whitelist-gated so non-whitelisted models pay only an always-false predicate.

## Summary by Sourcery

Reduce permission query overhead by memoizing complete scope decisions and batching role lookups during organization reconciliation.

Enhancements:
- Memoize complete permission scope decisions, including global permission checks, to avoid repeated queries within a request.
- Optimize organization group reconciliation by loading scoped role names in a single query per reconciliation.

Tests:
- Add regression coverage confirming repeated scoped permission checks, including cached global access, execute no additional queries.
- Update room query expectations to reflect reduced permission-check query overhead.

### DIFF
--- a/apps/betterangels-backend/accounts/services.py
+++ b/apps/betterangels-backend/accounts/services.py
@@ -321,6 +321,15 @@ def reconcile_org_groups(org: Organization) -> None:
     org_types = OrganizationProfile.objects.values_list("org_types", flat=True).filter(organization=org).first()
 
     if org_types is not None:
+        # One query for the whole reconcile, not one per template: the set of
+        # scoped Role names (role-backed templates are Grant-authoritative).
+        role_backed_names = set(Role.objects.filter(is_global=False).values_list("name", flat=True))
+        unscoped = {template_config.name for template_config in REGISTRY.unscoped}
+        # ``expected`` excludes role-backed templates, so ``derived - expected``
+        # deletes any stale legacy row a pre-teardown org still carries — the
+        # Grant is authoritative now, and the row is dead weight.
+        derived = {name for name in REGISTRY.template_names() if name not in unscoped}
+
         expected: set[str] = set()
         expected_role_backed: set[str] = set()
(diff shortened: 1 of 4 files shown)


vecchp:
Closing: scopes()-memoization content already landed on main (selectors._scope_cache); rest of the PR sits on the stale pre-redesign chain.

PR #2421, opened 2026-09-02
fix(perm): can_obj fails closed for per-record writes on platform-shared models (C1)

vecchp:
## Summary

Implements the full-audit finding **C1** (recorded in #2417): per-record **writes** on a platform-shared model must fail closed. This is the last unblocked, decision-independent fix before the clients cutover.

## The bug (reproduced)

A user holding `clients.change_clientprofile` via an **org-scoped Grant** got:
- `visible(ClientProfile)` → **all rows** (correct — the platform-shared *read* rule)
- `can_obj(user, CHANGE, <unrelated client>)` → **True for every client**

Root cause: `can_obj()` delegated to `visible()`, whose platform-shared branch ("perm held anywhere ⇒ all rows") is the read rule and leaked into the per-record **write** check. Once the clients cutover wired an org-scoped role carrying client perms, any holder could CHANGE/DELETE every client.

## The fix

`can_obj()` on a platform-shared model (`org_via = None`) now checks the **object arm** (or the global tier) and **fails closed** otherwise; org-scoped models are unchanged. Documented convention (ADR §2.4 + RFC 0002): write services on platform-shared models must load records via `can_obj` — never the read-side `visible()`.

## Tests (fail against the old code)
- org-grant holder `can_obj(CHANGE/DELETE)` on an unrelated profile → **False**
- an object grant restores per-record authority for the holder
- the global tier can `can_obj` any platform-shared row
- reads remain platform-shared via `visible()` (unchanged)

## Validation
Full `accounts`/`common`/`shelters` suite: **982 passed, 0 failed** · ruff format/check clean · mypy (strict) clean.

## Notes
- C2 (E005/E006 guard on scoped-role client-write perms) is **resolved by this fix** rather than a new check: once `can_obj` fails closed, scoped roles carrying client-write perms are enforceable via object grants, so a permanent E006 would conflict with the cutover end-state. Documented in the ADR.

## Summary by Sourcery

Fail closed for per-record authorization on platform-shared models so organization-scoped permissions cannot mutate unrelated records.

Bug Fixes:
- Prevent per-record writes on platform-shared models from being authorized solely because a user holds the permission through an organization grant.
- Allow platform-shared record access through explicit object grants or the global permission tier while preserving existing read behavior.

Enhancements:
- Document the separation between platform-shared read visibility and per-record write authorization, including the requirement for write services to use can_obj.

Documentation:
- Update the authorization ADRs with the platform-shared per-record write policy and its relationship to the client-writes ownership decision.

Tests:
- Add coverage for organization-grant holders, object grants, global-tier access, and unchanged platform-shared reads.


vecchp:
Folded into #2415 (perm/grant-object-arm, commit 10e24b49): can_obj fails closed on platform-shared per-record writes alongside the object-grant arm that made them possible; the RFC 0002 precondition note is folded into #2418. Closing.

PR #2422, opened 2026-09-02
fix(perm): view_private_shelter is global-tier only (PR 13)

vecchp:
## Summary

Audit cleanup: the **scoped** Shelter Operator role carried `view_private_shelter`, but its only consumer — the public `.approved()` directory (`shelter_list`) — reads it via legacy `has_perm`, which **Grant rows don't feed**. A grant-based shelter operator could never exercise it there (their own org's private shelters come through the org-scoped `visible()` operator path instead), so the scoped role declared an authority it can never exercise.

In the old model, the SO legacy group leaked `view_private_shelter` **globally** through `has_perm` — the exact org-bypass bug class this migration fixes. The new behavior (scoped operators don't see *other orgs'* private shelters in the public directory) is the intended correction; this PR only stops declaring the dead perm on the scoped role.

## Changes

- Remove `Shelter.perms.VIEW_PRIVATE` from the `SHELTER_OPERATOR` template (kept on `GLOBAL_SHELTER_OPERATOR`); `sync_roles` drops it from the SO Role.
- Test: `view_private_shelter` is **absent** from the scoped SO Role and **present** on the GSO Role.

## Validation
Full `accounts`/`common`/`shelters` suite: **983 passed, 0 failed** · ruff format/check clean · mypy (strict) clean.

## Summary by Sourcery

Limit private-shelter directory access to the global Shelter Operator tier and align role permissions with the scoped access model.

Bug Fixes:
- Restrict `view_private_shelter` to the global Shelter Operator role so scoped operators no longer declare an unusable global permission.

Tests:
- Add coverage verifying that the private-shelter permission is absent from the scoped role and retained by the global role.

### DIFF
--- a/apps/betterangels-backend/shelters/groups.py
+++ b/apps/betterangels-backend/shelters/groups.py
@@ -52,7 +52,11 @@ SHELTER_OPERATOR = TemplateConfig(
         Reservation.perms.CHANGE,
         Reservation.perms.DELETE,
         Reservation.perms.VIEW,
-        Shelter.perms.VIEW_PRIVATE,
+        # view_private_shelter deliberately NOT on the scoped role: private
+        # shelters in the public directory are a global-tier gate (the only
+        # consumer reads has_perm, which Grants don't feed); a shelter operator
+        # sees their org's private shelters through the org-scoped visible()
+        # path instead (ADR 0001 §2.4).
         ClientProfile.perms.VIEW,
     ],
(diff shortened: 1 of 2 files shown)


vecchp:
Folded into #2410 (perm/grant-roles, commit a7a85f21): view_private_shelter removed from the scoped Shelter Operator role at the point sync_roles provisions it. Test lives in test_roles.py. Closing.

PR #2423, opened 2026-09-02
fix(perm): cloning a bed/room requires ADD (PR 14)

vecchp:
## Summary

Audit finding (audit #4): `bed_clone`/`room_clone` **created rows** while only resolving the source with **VIEW** — so a VIEW-only user (e.g. a delegated viewer) could clone (= create) beds and rooms they could never otherwise create. This was inconsistent with the migration's own create convention (ADR 0001 §2.6: creates check `can(ADD)`).

## Changes

- `bed_clone` / `room_clone` now resolve the source with VIEW **and then require** `can(user, ADD, org)` — matching `bed_create`/`room_create`.
- Tests (fail against the old code):
  - a VIEW-only viewer (scoped Role carrying `view_bed`/`view_room` only) gets `PermissionDenied` on clone, and no clone row is created
  - existing clone tests still pass (SO holders have ADD)
- Clone mutation query counts bumped by the added `can(ADD)` check (bed 34→37, room 32→35).

## Validation
Full `accounts`/`common`/`shelters` suite: **985 passed, 0 failed** · ruff format/check clean · mypy (strict) clean.

## Summary by Sourcery

Enforce create-level authorization for bed and room cloning.

Bug Fixes:
- Require ADD permission before cloning beds or rooms, preventing VIEW-only users from creating clone records.

Tests:
- Add coverage confirming VIEW-only users are denied cloning and no duplicate records are created, while updating mutation query-count expectations.

### DIFF
--- a/apps/betterangels-backend/shelters/services/bed.py
+++ b/apps/betterangels-backend/shelters/services/bed.py
@@ -123,10 +123,14 @@ def bed_delete(*, user: "User", organization_id: str, bed_ids: list[int]) -> lis
 def bed_clone(*, user: "User", organization_id: str, bed_id: str) -> Bed:
     """Clone an existing bed, including all M2M relationships.
 
-    Scopes to *organization_id* where *user* is a member.
+    Scopes to *organization_id* where *user* is a member.  Cloning creates a
+    new bed, so it follows the create convention (ADR 0001 §2.6): the source
+    is resolved with view authority and create authority is checked with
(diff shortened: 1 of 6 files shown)


vecchp:
Folded into #2412 (perm/grant-cutover, commit e373e86b-equivalent): the clone-requires-ADD fix is now part of the shelter-domain cutover where the clone mutations landed. Closing.

PR #2428, opened 2026-09-02
feat(perm): reports reads authorize via grant OR legacy (§5.3 slice 2)

vecchp:
## What & why

Second slice of ADR §5.3 (org-admin role-backed milestone), grouped by endpoint
(**reports** reads only). `view_reports` rides the **legacy `ORG_ADMIN` template**
— no scoped `Role` row exists, so no holder has a `Grant` and the pure-Grant
predicate returns nothing. A grants-only flip would strip access from every org
admin. This slice converts both report consumers to the transitional dual-read
(introduced in #2427):

- **GraphQL** `reportSummary` — `HasOrgPerm` → `HasOrgPermOrGrant` (slice 1 extension).
- **DRF** `ExportInteractionDataApi` — `HasReportAccess` now resolves the org via
  `report_org_for_user()`: the legacy `get_user_permitted_org` arm first (current
  behavior, one query) **or** the grant `can()` arm (dormant until provisioning).
  `can()` never implies existence (ADR §2.6 finding F7) — the org is re-fetched by pk.

The predicate stays pure-grant; both transitional arms are deleted in the §5.3
provisioning PR. `schema.graphql` regenerated (`@hasOrgPermOrGrant` on
`reportSummary`). FE types need regen in a node-enabled checkout.

## TDD
New grant-arm tests failed before the change, pass after:
- DRF: scoped-Grant holder (no PermissionGroup) can export; Grant at org A does
  not authorize exporting org B.
- GraphQL: scoped-Grant holder can read `reportSummary` (aggregates real notes);
  Grant at org A does not authorize a summary at org B.

## Changes
- `reports/schema.py` — swap extension on `reportSummary`
- `reports/permissions.py` — `report_org_for_user()` dual-read + `HasReportAccess`
- `reports/tests/test_views.py` — `grant_view_reports_via_role` helper + 4 tests
- `schema.graphql` — regenerated

## Verification
- reports suite: 68 passed
- ruff format/check clean; mypy strict clean

## Summary by Sourcery

Enable transitional grant-or-legacy authorization for report reads while closing organization permission-isolation vulnerabilities.

New Features:
- Authorize report exports and GraphQL summaries through scoped grants while retaining legacy permission-group access during the transition.

Bug Fixes:
- Prevent permissions held by one organization group from incorrectly authorizing members of a different group.
- Prevent report access granted for one organization from authorizing access to another organization.

Tests:
- Add coverage for grant-based report access and cross-organization isolation across DRF and GraphQL.
[...]

PR #2426, opened 2026-09-02
docs(perm): correct ADR §4.1 readiness matrix + add §5.3 org-admin milestone

vecchp:
## What & why

The ADR §4.1 readiness matrix mislabeled four domains "Mechanical". Investigation
(code-grounded) showed two distinct blockers:

1. **Tasks / Referrals** write guardian rows at creation (`assign_object_permissions`) —
   the same over-permission pattern as notes §5. Promoting CHANGE/DELETE onto the org
   role would let every role-holder mutate every row in the org's scope, including
   cross-org rows the org should only view. → reclassified **Not mechanical — §5-equivalence**.
2. **Teams / Reports** have clean domain paths but their perms live on the legacy
   `ORG_ADMIN`/`ORG_SUPERUSER` templates (`accounts/groups.py`), which have **no scoped
   `Role` row** → no holder has a Grant → the pure-Grant predicate returns nothing.
   Role-backing is all-or-nothing per template (`reconcile_org_groups` retires the
   legacy group the moment a `Role` row exists), so teams/reports/member-management
   must cut over **atomically** in a new **§5.3 milestone**, not as standalone PRs.

Adds §5.3 documenting the milestone + the transitional **dual-read** split-stack
(consumer slices green-alone, provisioning PR retires the legacy arms). Merge-guide
ordering updated.

## Review focus
- Matrix classification vs. the actual authority paths (grep `accounts/groups.py`,
  `reconcile_org_groups`, `common/permissions/selectors.py`).
- §5.3 atomicity claim: a scoped `Role` row named "Organization Admin" makes reconcile
  stop creating + delete the legacy group; every `ORG_ADMIN` perm consumer breaks in
  the same deploy unless converted.

Docs only — no code.

## Summary by Sourcery

Correct the grant-migration roadmap and document the atomic role-backed transition for org-admin permissions across teams, reports, and member management.

Enhancements:
- Correct the ADR’s org-scoping examples and delegation semantics, including ownership declarations for org-owning models and role-keyed delegation inheritance.
- Reclassify remaining domain migrations by their actual authority paths and document the guardian-row and legacy-template blockers.
- Define an atomic org-admin role-backed milestone covering teams, reports, and member management, including transitional dual-read sequencing.
- Update the migration merge guide to reflect the new milestone ordering.

Documentation:
- Expand the authorization ADR with the corrected §4.1 readiness matrix and new §5.3 org-admin migration plan.

PR #2430, opened 2026-09-02
fix(perm): single-join org-perm check + role-keyed delegation (audit)

vecchp:
## What & why

Two correctness fixes from the external audit of the grant stack
(`~/.claude/plans/can-you-do-a-eager-aurora.md`), both verified by reproduction:

1. **`/reports/export/` org-history leak** — `get_user_permitted_org` chained two
   `.filter()` calls over the multi-valued `permission_groups` relation, so the
   membership and permission conditions could be satisfied by **different**
   groups at the same org. Every org is provisioned with every template, so a
   Caseworker (member of the CASEWORKER group) passed the `view_reports` gate via
   the org's ORG_ADMIN group, and `note_list_for_org` (org + date only) returned
   the org's whole note history as CSV. Fix: both conditions now stay in a
   **single** `.filter()` call, mirroring the already-correct
   `_org_perm_exists_across_fields`. This is the legacy arm my §5.3 slices
   (#2428/#2429) adopted, so one fix covers all three.

2. **Delegation amplification** — `scopes()`'s delegation arm keyed "acts at B"
   on holding *any* grant at B, so a member of B holding only a weak role
   inherited everything B delegated at C — **more authority at C than at B**
   (a VIEW-only holder at B could delete C's shelters). "Acts at B" is now
   **role-keyed** (`grants__role__in=Subquery(roles)`): inheritance requires a
   grant at B whose role carries the permission being checked. Not reachable
   while only one scoped `Role` exists, but §5.3 (role-backing `ORG_ADMIN`)
   creates the second — fixed while cheap. ADR §3/§2.4 updated to match.

## TDD
3 new tests failed before the fix, pass after:
- `accounts/tests/test_org_perm_check.py` — membership in one group no longer
  confers another group's permission (helper-level, control test included);
- `reports/tests/test_views.py` — DRF end-to-end: a CASEWORKER at a provisioned
  org gets 403 on `/reports/export/`;
- `common/tests/test_delegation.py` — a weak-grant member of B inherits VIEW at C
  but not CHANGE/DELETE (the old code granted all three).

## Changes
- `accounts/permissions.py` — single-join `get_user_permitted_org`
- `common/permissions/selectors.py` — role-keyed `acting_at` in `scopes()`
- `docs/adr/0001-grant-based-authorization.md` — §3/§2.4 delegation text
- tests above

## Verification
- accounts/common/shelters/teams/reports: 1128 passed
- ruff format/check clean; mypy strict clean. No schema change.

## Summary by Sourcery

Fix organization permission bypasses and delegation amplification in the grant-based authorization system.

Bug Fixes:
- Prevent organization permission checks from combining membership and permission conditions across different permission groups.
- Prevent weak roles from gaining stronger delegated permissions at downstream organizations.

Enhancements:
- Align the authorization ADR with role-keyed delegation behavior.

Documentation:
- Update authorization documentation to describe role-keyed delegation and its remaining role-translation limitation.

Tests:
- Add regression coverage for single-join organization permission checks, report export authorization, and delegation amplification.


vecchp:
Content folded into origin PRs: get_user_permitted_org single-join fix landed in #2428 (reports, its first adopter), C-1 role-keyed delegation landed in #2413 (grant-delegation), ADR note in #2426. Closing.

PR #2438, opened 2026-09-04
DEV-2478 Run ESlint with Wildfire App

nparmar1:
[DEV-2478](https://betterangels.atlassian.net/browse/DEV-2478)

This story focuses on configuring the Wildfire app for eslint and addressing all the lint erros.

## Summary by Sourcery

Configure ESLint for the Wildfire app and bring its codebase into compliance while strengthening related type safety and runtime handling.

Bug Fixes:
- Harden Wildfire data parsing and submission behavior by validating external content and ensuring retry attempts complete reliably.
- Prevent unsafe rendering of subscription errors and clean up event listeners when the translation menu closes.

Enhancements:
- Enable the Wildfire app to use the repository ESLint configuration and resolve lint issues across its components and utilities.
- Improve survey, print-state, modal, and resource handling with stricter types and more reliable React effect dependencies.

Build:
- Configure the Wildfire app's Vite server to access the workspace root and include it in Nx project inputs.

[DEV-2478]: https://betterangels.atlassian.net/browse/DEV-2478?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/wildfires/src/app/shared/components/register/MailChimpSubscribe.tsx
```diff
@@ -60,10 +60,11 @@ const SignupForm = ({ status, message, subscribe }: ISubscribeFormProps) => {
         <div className="flex flex-row flex-wrap w-full mb-2">
           <div className="flex flex-row flex-wrap w-full mb-2">
             <div className="flex flex-1 flex-col mb-2">
-              <label className="mb-1 font-bold">
+              <label htmlFor="firstName" className="mb-1 font-bold">
                 First Name<span className="text-[#FF0000]">*</span>
               </label>
               <input
+                id="firstName"
```
tglaz: this isn't blocking, but can you check if it'd work just as well without adding the `id` attribute?
the HTML `id` attribute needs to be unique, and `firstName` is kind of generic - probably ok in this app, but not best practice.

you may be able to avoid adding htmlFor + id attributes by wrapping the input with the label so you'd have something like below. 
if this renders as expected then I'd change it, otherwise your implementation is ok. I wouldn't spend too much time on this app.

```
              <label className="mb-1 font-bold">
                Last Name<span className="text-[#FF0000]">*</span>
                <input
                  style={styles.input}
                  onChange={handleChange}
                  type="text"
                  name="lastName"
                  value={formData.lastName}
                  required={true}
                  placeholder="Last Name"
                />
              </label>
```

apps/wildfires/src/app/shared/components/survey/utils/validateConfig.ts
```diff
@@ -67,7 +67,7 @@ function validateNextFormIds(forms: TSurveyForm[]): string[] {
   return invalidFormIds;
 }
 
-function getAllRules(forms: TSurveyForm[]): TConditionRule[] {
+function _getAllRules(forms: TSurveyForm[]): TConditionRule[] {
```
tglaz: we should probably avoid this pattern for unused functions. I'd prefer just deleting it if unused.

tglaz (commented):
looks good - but few minor comments.

While you're doing this, would you mind updating the `apps/wildfires/vite.config.mts` to add a fix for font loading? 
This issue already already existed, but font files get a 403, and just needs a small update to what we do in the vite.config.mts files for other apps:

```
--- a/apps/wildfires/vite.config.mts
+++ b/apps/wildfires/vite.config.mts
@@ -14,6 +14,7 @@ import {
 
 const SERVER_PORT = 8200;
 const SERVER_PORT_PREVIEW = 8201;
+const WORKSPACE_ROOT = path.resolve(__dirname, '../..');
 
 export default defineConfig(({ mode }) => {
   const isDev = mode === 'development';
@@ -30,6 +31,7 @@ export default defineConfig(({ mode }) => {
     server: {
       port: SERVER_PORT,
       host: 'localhost',
+      fs: { allow: [WORKSPACE_ROOT] },
     },

```

tglaz:
[not blocking] 
also, we left `no-explicit-any` as a warning for now, but will change it to an `error` soon, so if you want to update all of these warnings (or any other ones) then feel free to do so. but not really in scope so up to you.

```
 warning  Unexpected any. Specify a different type  @typescript-eslint/no-explicit-any
 ```

**Comment by nparmar1** [comment]
Hi, @tglaz, I made the following updates addressing your feedback:

1. Wildfires font loading — Updated `vite.config.mts` with `server.fs.allow` for the workspace root, fixing the local font-loading 403 issue.

2. Mailchimp form labels — Wrapped each input inside its label and removed the generic `id/htmlFor` attributes. Verified label focus, layout, and form validation still work correctly.

3. Unused code — Removed the unused function from` validateConfig.ts` rather than suppressing the lint error.

4. Lint cleanup — Removed the remaining explicit `any` usages and addressed the remaining hook dependency, event-listener cleanup, `dangerouslySetInnerHTML`, and unused-variable warnings. `yarn nx lint wildfires` now passes with 0 errors and 0 warnings.

Please have a look when you have the chance.

File: apps/wildfires/src/app/pages/result/index.tsx [hunk 3]
```diff
@@ -23,31 +28,37 @@ function findQuestionById(id: string) {
...
     for (const tag of optionTags) {
       answerTags.push(tag);
     }
   }
+
   return answerTags;
 }
 
-function getTags(answers: any[]): string[] {
+function getTags(answers: TSurveyResults['answers']): string[] {
```
tglaz, line 51: isn't TSurveyResults['answers'] just TAnswer?

apps/wildfires/src/app/shared/components/bestPractices/BestPracticesCard.tsx
```diff
@@ -17,12 +17,14 @@ export default function BestPracticesCard(props: IBestPracticesCardProps) {
...
-      setExpand(true);
+      setExpand((currentExpand) => {
+        wasExpandedBeforePrintRef.current = currentExpand;
+        return true;
+      });
     } else {
       setExpand(wasExpandedBeforePrintRef.current);
     }
-  }, [isPrinting]); // ✅ Removed expand from dependencies to prevent race conditions
+  }, [isPrinting]);
```
tglaz, line 27: optionally can look below (line 39) and incorporate the `tailwindcss(suggestCanonicalClasses)` suggestions.

```
The class `min-w-[50px]` can be written as `min-w-12.5`
```

apps/wildfires/src/app/shared/components/navBar/shared/GoogleTranslateBtn.tsx
```diff
@@ -59,43 +59,57 @@ export function GoogleTranslateBtn(props: IProps) {
   }, []);
 
   useEffect(() => {
-    if (isOpen && buttonRef.current && dropdownRef.current) {
-      const buttonRect = buttonRef.current.getBoundingClientRect();
-      const dropdownWidth = dropdownRef.current.offsetWidth;
+    if (!isOpen || !buttonRef.current || !dropdownRef.current) {
+      return;
```
tglaz, line 63: 👍 
early returns are great

apps/wildfires/src/app/shared/components/survey/Survey.tsx
```diff
@@ -74,7 +74,7 @@ export function Survey(props: IProps) {
 
   useEffect(() => {
     onChange && onChange(answers);
-  }, [answers]);
+  }, [answers, onChange]);
```
tglaz, line 77: I want to highlight this change as this adding dependencies into this array is not always desired - and sometimes can cause a big issues as could trigger re-renders or recalculations at wrong time, or just way too often. Please review the comments below and decide on best approach:

This is a common issue to look out for with the `useEffect deps`.

## Adding `onChange` to the deps changes the effect's semantics

**1. Why it was deliberately keyed to `answers` only**

This effect is the "answers changed" notification: it should fire when the *data* changes, and `answers` (a `useState` array in `SurveyProvider`) gets a new array identity exactly when an answer is added/updated. `onChange` is just the callback that delivers that data — it's not a trigger. The previous entry `` `onChange` `` was a constant (template literal with no interpolation), so the deps were effectively `[answers]`, which kept the effect from re-running merely because the parent re-rendered with a new function identity.

**2. What adding it does**

Every caller passes an inline, un-memoized callback — `FiresSurvey.tsx` defines `onChange` in the component body — so its identity changes on **every parent render**. With `[answers, onChange]` the effect now means "run on every parent render *and* whenever answers change", so `onChange(answers)` fires repeatedly with an unchanged payload.

That's invisible today only because the current handler is a no-op. Once it does real work this becomes:
- duplicated side effects (analytics, network, logging), and
- a possible infinite loop: if the handler writes state (e.g. `storeSurveyResults(...)`), the parent re-renders → new `onChange` identity → effect runs again → writes again…

Worth noting the previous constant form isn't lint-clean under our current react-hooks rules either (it reports "missing dependency" *and* "complex expression"), so replacing it was understandable — but this replacement trades a lint warning for a behavior change.

**3. Options**
- **Keep `[answers]` and suppress the rule on that line** — smallest possible diff, and semantically valid (the effect re-runs from the latest render, so the current `onChange` is used whenever it fires; the only difference is a changed callback isn't called for data that already exists). Caveats: it may take two rules to silence here (`react-hooks/exhaustive-deps` and `@eslint-react/exhaustive-deps`), the suppression also hides future genuinely-missing deps on that line, and the intent only lives in a comment. Acceptable fallback — but since we're on React 19.2, `useEffectEvent` gives the same behavior without opting out of the check, so should consider it. Otherwise should include a comment explaining why it's omitted so it's not added in the future:
  ```tsx
  useEffect(() => {
  // onChange is intentionally not a trigger: this should fire only when
  // `answers` change, and the latest callback is read from the current render.
  // eslint-disable-next-line react-hooks/exhaustive-deps, @eslint-react/exhaustive-deps
  }, [answers]);
  ```
- **Keep `[answers]` and read the latest callback via `useEffectEvent`** (React 19.2 is already pinned in the repo; this pattern lints clean here):
  ```tsx
  const emitChange = useEffectEvent((results: TAnswer[]) => {
    onChange?.(results);
  });

  useEffect(() => {
    emitChange(answers); // Effect Events are intentionally left out of deps
  }, [answers]);
  ```
  (the older-React equivalent is the "latest ref" pattern)
- **Keep `onChange` in the deps, but stabilize it at the call sites**: wrap the handlers in `useCallback` (what `exhaustive-deps` suggests) so the identity only changes when it genuinely should.
- **Keep as-is intentionally**: acceptable only if re-firing on every render is harmless for callers (idempotent callback that never sets parent state). If that's the intent, please add a short comment saying so, otherwise the next reader will "fix" it back.

apps/wildfires/src/app/shared/components/surveyResources/shared/ResourceCallout.tsx
```diff
@@ -40,12 +40,14 @@ export function ResourceCallout(props: IProps) {
...
-      setShow(true);
+      setShow((currentShow) => {
+        wasExpandedBeforePrintRef.current = currentShow;
+        return true;
+      });
     } else {
       setShow(wasExpandedBeforePrintRef.current);
     }
-  }, [isPrinting]); // ✅ Removed show from dependencies to prevent race conditions
+  }, [isPrinting]);
```
tglaz, line 50: **Non-blocking: the print expand/collapse effect in `ResourceCallout` — fine to merge, worth a follow-up refactor**

Not blocking for this PR, because it works today for a reason that isn't obvious from the code:

- Nothing in the app currently calls `setPrinting` — the results page just mounts a print copy of the tree with `initialPrinting={true}` and a screen copy with the default `false`. So `isPrinting` is effectively static per instance, the effect runs once per mount, and both branches collapse to "set the initial expanded state". No user-visible bug.
- The `[isPrinting]` deps are complete; no `exhaustive-deps` concern.

Why it's still worth tracking as a follow-up:

1. The state updater writes `wasExpandedBeforePrintRef.current` — updaters must stay pure (React double-invokes them in StrictMode and can rebase/discard them under concurrent rendering). It's idempotent today, so nothing breaks, but it's fragile if that block ever grows.
2. The `else` branch unconditionally resets `show` to the ref's initial `false` on mount. It only no-ops because `useState(false)` coincidentally matches — change the default (or initialize the ref differently) and the callout silently collapses on mount.
3. The whole save/restore half is dead code while `isPrinting` cannot transition.
4. If dynamic printing is ever wired up (calling `setPrinting` around `window.print()`), the forced-open becomes a one-shot rather than an invariant: a toggle during printing hides the content mid-print, and that toggle is then silently overwritten when printing ends. The control also stays clickable with no visible effect during that window.

Suggested follow-up (should also cover `BestPracticesCard.tsx`, which has the same pattern — ideally extract a small hook so it lives in one place):

```tsx
const [show, setShow] = useState(false);
const { isPrinting } = usePrint();

const isOpen = isPrinting || show; // forced open while capturing for print

const handleToggle = useCallback(() => {
  setShow((prev) => !prev);
}, []);
```

Then use `isOpen` for `aria-expanded`, the chevron rotation, and the content's `block`/`hidden` accordingly, and delete the ref + effect. This keeps the callout open for the entire print capture, removes the impure updater and the mount-time reset, and behaves the same whether `isPrinting` stays static or later becomes dynamic. (If toggling during printing shouldn't be possible at all, also disable the toggle while `isPrinting`.)

Unrelated drive-by: the hardcoded `id="resource-callout-content"` collides across every callout instance (there are two per resource, print + screen trees), so `aria-controls` resolves to the wrong node — `useId()` fixes that.

apps/wildfires/src/app/shared/components/surveyResources/SurveyResources.tsx
```diff
@@ -67,7 +67,7 @@ export function groupResources(resources: TResource[]): TCategoryResources[] {
   );
 
   const categoryResources: TCategoryResources[] = Object.entries(grouped).map(
-    ([slug, entries]) => ({
+    ([, entries]) => ({
```
tglaz, line 70: this works, but preferred format would be `([_slug, entries]) => ({`....

we added a rule somewhere for linter to ignore args prefixed with an underscore, which is a common convention. Leaving an empty space kind of hides what's available, though not a huge issue

apps/wildfires/src/app/shared/hooks/useSurveySubmission.tsx
```diff
@@ -42,53 +40,62 @@ const useSurveySubmission = (
...
-        setStatus('success');
-        console.log('Survey submitted successfully:', surveyID);
-      } catch (error) {
-        console.error(`Attempt ${retries + 1} failed:`, error);
-
-        if (retries < maxRetries) {
-          setTimeout(
-            () => attemptSubmission(survey, surveyID, retries + 1),
-            retryDelay,
+          const newStoredData: StoredSurveyData = {
```
tglaz, line 66: **Non-blocking, but worth fixing: a failed local write currently retries a successful POST**

The success bookkeeping here (`setStoredData` / `localStorage.setItem`) sits inside the same `try` whose `catch` retries the request. If the storage write throws — quota exceeded, or storage blocked/disabled in the browser — a **successful** submission is treated as a failure: the same payload is POSTed up to 3 more times, and the hook can end on `status: 'error'` even though the server accepted it (duplicate rows depending on whether the endpoint is idempotent on `id`).

Keeping the write in its own `try` means the retry path only ever covers the network call:

```suggestion
          const newStoredData: StoredSurveyData = {
            answers: survey.answers,
            surveyID,
          };

          setStoredData(newStoredData);

          try {
            localStorage.setItem(
              LOCAL_STORAGE_KEY,
              JSON.stringify(newStoredData),
            );
          } catch (storageError) {
            // POST already succeeded - don't retry it because we couldn't
            // persist locally (quota / blocked storage).
            console.error('Failed to persist survey submission:', storageError);
          }

          setStatus('success');
          console.log('Survey submitted successfully:', surveyID);
```

apps/wildfires/src/app/shared/hooks/useSurveySubmission.tsx
```diff
@@ -25,12 +25,10 @@ const useSurveySubmission = (
   maxRetries = 3,
   retryDelay = 1000,
 ): SubmissionStatus => {
-  const [storedData, setStoredDataState] = useState<StoredSurveyData | null>(
-    () => {
-      const data = localStorage.getItem(LOCAL_STORAGE_KEY);
-      return data ? JSON.parse(data) : null;
-    },
-  );
+  const [storedData, setStoredData] = useState<StoredSurveyData | null>(() => {
```
tglaz, line 28: **Non-blocking: an unguarded `JSON.parse` can crash the results page**

If `survey_submission` holds malformed or legacy JSON (or `localStorage` access itself throws in a locked-down browser), this throws during the initial render rather than degrading to "no stored submission" — the whole Result page fails to render. Folding the read into a `try` keeps a bad local value from taking the page down. Returning `null` is enough; clearing the bad key is optional and needs its own guard, since access can throw in the same cases.

```suggestion
  const [storedData, setStoredData] = useState<StoredSurveyData | null>(() => {
    try {
      const data = localStorage.getItem(LOCAL_STORAGE_KEY);
      return data ? JSON.parse(data) : null;
    } catch (error) {
      // Corrupted or legacy value, or storage unavailable - start fresh
      // instead of throwing during render.
      console.error('Failed to read stored survey submission:', error);
      return null;
    }
  });
```

could probably move the safe parsing into a `safeParse` util - in same file probably ok

tglaz (changes requested):

PR #2443, opened 2026-09-09
feat(perm): teams fully grant-based & header-free — ORG_ADMIN + CASEWORKER role-backed, org in filter/payload (DEV-2559)

vecchp:
## What & why

Cut the **teams** domain fully over to the grant-based authorization model on `main` (ADR 0001 §5.3, teams slice). The teams surface — **read and write** — now authorizes via `can()`/`require_can` over role-backed templates with backfilled Grants; org membership is no longer consulted for teams.

- **Mutations** (`createTeam`/`updateTeam`/`deleteTeam`): grant-only. `ORG_ADMIN`/`ORG_SUPERUSER` are role-backed with a scoped Role carrying `teams.*`.
- **Read** (`teams` query): grant-only. `CASEWORKER` is role-backed as the **RFC 0003 first step** — a scoped `Caseworker` Role carrying `teams.view_team` — so the workers who pick teams on notes/tasks read via grants.

Why not the §5.3 "all four consumers at once" milestone: on `main` the member-management codenames (`organizations.*`) and `reports.view_reports` resolve to no concrete model, so a scoped RoleDef carrying them fails `sync_roles`' phantom-ContentType guard. Teams was the slice that could ride Roles today; reports and member management stay legacy-only until their own slices.

## Changes

- **`accounts/groups.py`** — `ORG_ADMIN_ROLE`/`ORG_SUPERUSER_ROLE` scoped RoleDefs carrying `teams.*`, exported as `ORG_ADMIN_ROLES`.
- **`notes/groups.py`** — `Team.perms.VIEW` added to the `CASEWORKER` template + scoped **`CASEWORKER_ROLE`** RoleDef carrying `teams.view_team` (RFC 0003 first step; the rest of the caseworker bundle stays legacy until RFC 0003).
- **`accounts/services.py`** — `sync_roles()` provisions all three role sets; the three backfills are deduped into one `_backfill_role_grants()` helper with `backfill_shelter_grants`/`backfill_org_admin_grants`/`backfill_caseworker_grants` wrappers.
- **`accounts/apps.py`** — `backfill_org_admin_grants()` + `backfill_caseworker_grants()` run in `_seed_on_migrate` after `sync_roles()`.
- **`teams/schema.py`** —
  - mutations: `HasOrgPerm` → `require_can(…, teams.*)`; org resolved via a module-level helper that fails closed on an unknown org header (no `DoesNotExist`).
  - `teams` read: `require_can(user, teams.view_team, org)` — grant-only. Membership helper removed. Org resolution: `TeamFilter.organizationId` wins; the `X-Organization-ID` header is the deprecated fallback (`_resolve_teams_org` fails closed on an unknown org).
- **`teams/models.py`** — `Team` declares `OrgScoped` (`org_via = ()`); required by `permissions.E005` (no migration).
- **`common/permissions/domain.py`** — `teams` joins `LEGACY_INERT_APPS` (legacy rows not reported; global tier folds like shelters).
- **Tests** — `teams/tests/test_grant_authorization.py`: grant-only read + write contract (role-backed ORG_ADMIN & CASEWORKER read/manage; legacy-only admin denied on mutations; ADD-only cannot update/delete; member-without-grant denied on the read; direct-grant holder & superuser read; grant at org A does not authorize org B). Read-org tests: org filter works without the header, the filter wins over a stale header, and an unknown filter org is denied. Equivalence/report tests strengthened: `test_org_admin_member_report_matches_legacy_enforcement` now asserts `can()` for `teams.*` (was vacuously asserting the legacy predicate), and new `CurrentUserTeamsReportCanEquivalenceTestCase` pins entry ≡ `can()` for the teams domain. Role-manager dual-write tests updated (CASEWORKER is role-backed now). Admin/mutation query-count guards updated for grant-holding members. FE: TeamsPage tests assert the query is issued with `filters.organizationId` and that a view-only holder sees neither Add nor the per-row actions menu.
- **`schema.graphql`** — regenerated (`@hasOrgPerm` drops off the three mutations; `TeamFilter.organizationId` and `CreateTeamInput.organizationId: ID!` added).
- **`libs/react/betterangels-admin` Teams page + drawer** — passes `activeOrg.id` as `filters.organizationId` (read) and in the `createTeam` payload; gates Add/Edit/Delete on the team grants (see design notes). `ThreeDotMenu` takes `canEdit`/`canDelete` and conditionally renders each action.
- **`docs/adr/0001`** — §5.3 status note.
- **Regenerated clients** — expo shared-clients schema, `@monorepo/ba-platform` types (`TeamFilter.organizationId`, `CreateTeamInput.organizationId`).
- **Cleanups from adversarial review** — the read's org filter is `organization_id: auto` (strawberry-django resolves the FK column to `organizationId: ID`; the hand-written `filter_field` + `Q` are gone, schema byte-identical); `team_get`'s org confine is optional; single `_resolve_read_org` + `_org_or_deny` helpers replace the duplicated resolvers.

## Design notes / deliberate scope

- **Pure grant, no membership arm.** Workers read teams via the role-backed `Caseworker` grant (`teams.view_team`); admins via `ORG_ADMIN`. The FE gates on `currentUser.organizationsOrganization[].permissions` (grant-native since #2414).
- **Org filter, header deprecated (read) — mutations are header-free.** The read takes the org as a `TeamFilter.organizationId` (authoritative, so switching orgs re-runs the query scoped to the right org); the `X-Organization-ID` header is kept only as a backward-compatible fallback — the mobile clients (`libs/expo`) still rely on it — and is marked for later removal. The **mutations never read the header**: `createTeam` carries `organizationId` in the payload (required — no row exists to scope by yet) and `updateTeam`/`deleteTeam` resolve the row the payload names by id and authorize at `team.organization` (the org is an attribute of the row).
- **FE action gating.** The admin Teams page gates Add on `teams.add_team` and the per-row Edit/Delete actions on `teams.change_team`/`teams.delete_team`, so a view-only holder (e.g. a role-backed caseworker) sees a read-only directory — no Add button, no actions menu (previously the actions menu rendered for everyone with `View`).
- **Backfill at deploy** covers every existing admin/caseworker; `OrgRoleManager.add_roles` mirrors Grants for new ones (role-backed templates). Legacy groups are **kept** (dual) so member management/reports keep working off the legacy arm until their slices; `teams` legacy rows are inert (suppressed from reports).
- **Role bundles are partial** (`teams.*` / `teams.view_team` only) — a documented divergence from the §5.3 "one template, four consumers" plan, forced by the phantom-ContentType guard. Later slices extend the RoleDefs and `sync_roles` reconciles idempotently.
- **Known, pre-existing admin cost**: the Django-admin Organization page does ~6 queries/member for grant-holding members (present for shelter operators since #2412). The query-count guards were updated to pin that bounded budget rather than pretend grant-holding members cost nothing.
- Behavior flip: a legacy-only holder (PermissionGroup, no Grant) fails closed on teams — intended; backfill covers existing org admins/caseworkers.

## Review round 2 — adversarial fixes

- **Malformed/blank org ids deny, they don't crash.** `_org_or_deny` validates with the house `get_or_none` pk guard, so `organizationId: "not-an-id"` or `""` (read filter, header, or create payload) is a clean `PERMISSION` refusal instead of a Django `ValueError`. An absent filter keeps the header fallback; an explicitly empty one denies — it never silently falls back.
- **One refusal for missing vs. forbidden teams.** `updateTeam`/`deleteTeam` answer a missing team with the same message as a team the caller may not touch (`PERMISSION_DENIED_MESSAGE` in `common/permissions/utils.py`), closing an existence oracle.
- **The membership ⇔ Grant mirror is now enforced at the `User.groups` m2m edge** (`accounts.signals`). Every writer keeps the invariant — `OrgRoleManager`, the user-admin group picker, scripts, the shell — and the bespoke `UserAdmin.save_related` hook is gone. Reverse writes (`permission_group.user_set.add/remove/clear`) are covered. A cascading delete of a legacy `PermissionGroup` deliberately leaves the Grants (teardown retires legacy rows; the Grants are the successor authority) — the admin delete page now says so instead of implying the capability is revoked.
- **"Template ≈ Role" is guarded.** `sync_roles` provisions from one `_all_role_defs()` list, `ORG_SUPERUSER_ROLE` derives from `ORG_ADMIN_ROLE`, and a test asserts every scoped RoleDef's bundle is a subset of its template's.
- **Denied-path tests can no longer pass on a crash.** `_assert_denied` asserts the expected message; new tests cover non-numeric/blank filter, header and payload ids; an explicitly-null filter id still falls back to the header.
- **FE/docs polish:** the Teams query is skipped when the holder lacks `View` (no doomed request); the org guard moved into the create branch only (edits don't need an active org); stale `teams` references removed from the `useActiveOrgState` comment; ADR + overview record the mirror invariant and the actual landing route.
- **Cleanup:** removed dead shims (the test-side `PERMISSION_DENIED = PERMISSION_DENIED_MESSAGE` alias, the now-unused `organization` kwarg on the mirror helpers, a single-use `ORG_ADMIN_ROLE_PERMISSIONS` constant, a defensive optional branch in `_assert_denied`) and pruned duplicated/stale comments. Small perf wins: batched the admin delete-page role lookup (N→1), `select_related("organization")` in the backfills, an empty-`pk_set` short-circuit in the signal, and the message-only uniqueness query is skipped when a team edit doesn't change the name.

## Verification

- Backend: full suite **1835 passed, 31 skipped** (all apps, incl. hmis); ruff check + format clean; targeted mypy clean on all touched files
- `manage.py check` clean (incl. `permissions.E005`); `makemigrations --check` → no changes
- ruff check + format clean; targeted mypy clean on all touched files
- FE: eslint clean on changed files; TeamsPage/TeamFormDrawer tests extended (vitest + tsc run in CI)


mikefeldberg (commented):
# Adversarial review — teams grant-only cutover

I reviewed the diff against the working tree, the permission machinery (`selectors.py`, `utils.py`, `role_manager.py`, `seed.py`, `checks.py`), the FE consumers, and `docs/styleguides/{python,react}.md`. The architecture is sound and unusually well-documented — the filter-wins-over-header rule, fail-closed org resolution, row-derived org on update/delete (no existence oracle), and the E005↔`OrgScoped` interplay are all handled correctly. The real issues are below, roughly by severity.

## Findings worth fixing before merge

### 1. `test_admin.py` — an N+1 regression test was weakened into an acceptance test (med)
`OrganizationMemberInlineQueryCountTestCase` (test name: `test_rendering_the_members_inline_does_not_query_per_member`, class docstring: *"its cost must not grow per member"*) previously asserted `len(many) == len(few)`. The diff replaces that with `assertEqual(len(many) - len(few), 6 * added_members)` — i.e. it now **requires** 6 extra queries per added member.

Why this happens: the test adds members as `CASEWORKER`, which this PR makes role-backed, so `member_add` now mirrors a `Grant` per member. The superuser Organization page renders `GrantInline` (`accounts/admin.py` ~L500), one row per grant, and each row's `autocomplete_fields` renders FK labels with per-row queries (no `select_related`/prefetch on the inline queryset). Result: **the admin page now costs ~6 queries × members for superusers** — and the codebase itself notes *"production has an organization with 90 members"* (≈540 extra queries/page).

Problems with the change as written:
- The class docstring + test name now contradict the assertion.
- The `added_members = 6` variable no longer holds the number of members added (6) but the per-member multiplier — and the magic `6` is unexplained, so any unrelated query change breaks it with no decomposition to debug.
- Per `python.md` testing guidance, converting a strong invariant into a linear-budget pin is a smell.

Recommendation: fix the actual N+1 (prefetch/`select_related` on `GrantInline.get_queryset`, or `raw_id_fields`) and restore equality; at minimum fix the naming/docstring and add a test that decomposes the 6.

### 2. The two new backfills have zero direct test coverage (med)
`backfill_org_admin_grants()` and `backfill_caseworker_grants()` (plus the shared `_backfill_role_grants`) are the entire basis for existing org admins'/caseworkers' teams authority after deploy — but only `backfill_shelter_grants` has tests (`accounts/tests/test_roles.py` `BackfillTestCase`). Per `python.md` ("test services and selectors as the primary test surface"), these are service-level functions with no tests: a template-name mismatch or wrong role lookup would silently strip team management from every existing admin/caseworker in production with no failing test. Mirror `BackfillTestCase`: one grant per member, idempotent, converts only the intended template.

### 3. Direct `user.groups.add(PermissionGroup)` bypasses grant mirroring — silent capability loss for the new roles (med)
`_mirror_grant` only runs when membership changes go through `OrgRoleManager`. The Django User-admin group picker (`_groups_without_scoped_roles`, `accounts/admin.py` ~L81) excludes only scoped **Role** rows — it still offers role-backed **PermissionGroup** rows (e.g. `Acme Housing [3] · Organization Admin`). A superuser adding a user straight to that group now produces an admin who holds the legacy group but **no Grant**, so after this cutover they cannot manage teams. Pre-cutover the legacy group alone granted teams, so this is a functional regression for that path — consistent between FE report and backend (both silently agree), which makes it *worse*: nobody notices.

This predates the PR for `SHELTER_OPERATOR`, but this PR widens it to `ORG_ADMIN`/`ORG_SUPERUSER`/`CASEWORKER` — precisely the roles the new surface gates on. Recommendation: exclude role-backed-template groups from the picker (or mirror in `UserAdmin.save_m2m`), and/or add a deploy-check + test pinning group↔grant symmetry.

### 4. The header-deprecation plan is incomplete at merge time (med-low)
The ADR + `_resolve_read_org` docstring say the `X-Organization-ID` read fallback "will be stripped once none send it." But only betterangels-admin's `TeamsPage` passes `filters.organizationId`. The mobile hook `useOrgTeams` (`useOrgTeams`) — feeding NoteForm, TaskForm, FilterTeamsOptions, UserTeamPreferenceSelect — sends only `{ isActive }` and relies entirely on the header; it doesn't even accept an org argument. When the header is stripped, mobile team pickers break.

Also note the read cutover is a behavior change for members with **no role** at the active org (previously membership sufficed; now a `teams.view_team` grant is required). Fine for caseworker/admins (backfilled), but role-less members who could previously list teams can no longer — confirm that's intended and record the mobile dependency explicitly in the ADR so the "strip the header" step lists every client that must migrate first.

### 5. Product surface: caseworkers now see the admin Teams list (low, confirm intent)
CASEWORKER is now role-backed with `teams.view_team`, so caseworkers' per-org entries fold in `teams.view_team` → `hasPermission(TeamPermissions.View)` becomes true in betterangels-admin, where the Teams route is `'safe'`. The new FE tests pin the view-only rendering (no actions), so this looks deliberate — just confirm the product wants caseworkers to see the full team list/search in the *admin* app, not only team *pickers* in mobile.

## Minor / polish

- **Denied-message triplication** (`teams/schema.py`): create uses `require_can`'s standard message; update/delete use bespoke "…update/delete this team." for both not-found and not-permitted (deliberate fail-closed, good) — consider one consistent message to keep FE known-error matching simple.
- **No mutation test for an unknown/foreign `organizationId` on `createTeam`** — only the query-filter path is covered; `_org_or_deny` is shared, so one test would pin it.
- **`_resolve_read_org`**: `filter_org_id or get_current_organization(...)` means an explicitly-present empty `organizationId` (`""` is falsy) silently falls back to the header; treat a present-but-empty ID as invalid.
- **RoleDef duplication**: `ORG_ADMIN_ROLE` and `ORG_SUPERUSER_ROLE` are identical copies today (correct, since only `teams.*` can ride a Role), but consider deriving `ORG_SUPERUSER_ROLE` from `ORG_ADMIN_ROLE` so the drift risk is explicit when the role bundle grows.
- **RoleDef registry sprawl**: `sync_roles` hand-assembles `(*ROLES, *ORG_ADMIN_ROLES, CASEWORKER_ROLE)` and each backfill re-lists roles; a single code-owned registry (matching the `E001`–`E006` check style) would prevent a future role-backing from being forgotten in one of the lists.
- **FE styleguide** (`react.md`): `TeamsPage.tsx` uses a raw `useQuery` for a list where the guide prefers a co-located custom hook — pre-existing, but this PR touches these exact lines, so it's the natural moment to extract one (the mobile `useOrgTeams` already exists as a template).
- **`TeamFormDrawer`** still requires `activeOrg` even on the edit path, though the backend no longer needs it there — harmless today, but the guard is now broader than the constraint.

## What's genuinely good (worth keeping)

- **No authz holes found**: filter-wins-over-header, unknown-org fail-closed, no existence oracle on update/delete, authorization always at the row's own org, `can()`/delegation semantics hold for the new roles, and update can't move a team across orgs (`UpdateTeamInput` has no org field).
- `Team` → `OrgScoped` correctly satisfies `permissions.E005` for the new scoped roles; the phantom-ContentType guard rationale for limiting RoleDefs to `teams.*` is accurate.
- The equivalence/tripwire test additions (`CurrentUserTeamsReportCanEquivalenceTestCase`, grant read/write/denied suites) are exactly the right shape, and keeping template↔Role consistent (CASEWORKER template also gains `teams.view_team`) closes a real drift gap.
- ADR §5.3 updated to actual main state rather than leaving aspirational text — good doc hygiene.

**Top asks:** (1) fix or honestly re-scope the admin query test in Finding 1, (2) add backfill tests, (3) close the direct-group-add gap in Finding 3, and (4) put the mobile header dependency on the record before anyone strips the header.

mikefeldberg:
## Previous review items — status

| Item | Status |
|---|---|
| Admin N+1 / weakened test (top finding) | **Fixed.** `raw_id_fields` + `LoadedRowRawIdWidget`/`GrantRowForm` + `select_related`; test back to `delta == 0`. |
| Backfill tests for org-admin/caseworker | **Fixed.** `OrgAdminAndCaseworkerBackfillTestCase` + bonus `RoleDefTemplateConsistencyTestCase`. |
| Direct group-add bypasses mirroring | **Fixed**, via the m2m-edge signal (different from my recommendation — see F2). |
| Mobile header-deprecation checklist | **Addressed as documentation** (ADR §5.3 now lists every header-only consumer). Hook itself not migrated; no ticket ref. |
| `ORG_SUPERUSER_ROLE` derivation, blank/UNSET org handling, unknown-org tests, message consolidation, `_all_role_defs` | **Done** (message constant not reused in `extensions.py` — see F6). |
| Teams authz registry canary (my Phase 1) | **Not addressed** — no `teams/tests/test_authz_registry.py`; see F8. |

## New findings

### F1 (medium) — reconcile-driven group deletion no longer revokes the mirrored authority
Verified empirically with a throwaway probe (since deleted): an org with types `["outreach","shelter"]` changes to `["shelter"]`; `reconcile_org_groups` deletes the stale `Caseworker` `PermissionGroup`; the member's membership is gone but the mirrored `Caseworker` Grant **survives** (`PROBE grants after reconcile dropped the Caseworker group: ['Caseworker']`). The member keeps `teams.view_team` (`can()` is grant-driven) while their legacy-domain perms (notes/tasks/reports/member management) are revoked with the group — a partial, inconsistent revocation the next post-migrate applies silently.

This is the direct cost of the deliberate "group deletion must not revoke Grants" rule (needed so teardown can retire legacy rows). The signal can't distinguish teardown from config cleanup, but `reconcile_org_groups` can: it knows these are stale derived rows. Options: (a) unmirror explicitly in `reconcile_org_groups` before the stale delete (best — restores complete revocation for config changes while keeping teardown semantics), (b) accept and document the asymmetry in the ADR, (c) revisit the teardown asymmetry. Add a test pinning whichever you pick; today the only pinned behavior is `test_deleting_the_group_leaves_the_grant_for_teardown`, which is teardown-shaped.

### F2 (low-med, architecture) — signal-hosted invariant: document the exception precisely
- `python.md` says "Never put business logic in models, signals, or GraphQL resolvers." The mirror is now enforced by `mirror_group_membership_grants` (a signal), with logic centralized in `role_manager`. It's a conscious, ADR-documented trade-off — but the next contributor may "fix" it back into services and silently lose the admin/script writers. Worth an explicit line in the styleguide (or a strong comment on the receiver) naming it as the sanctioned exception and why.
- The docstring claim "every writer keeps the invariant — the manager, the user admin, scripts, the shell" overclaims: `through.objects.bulk_create`, raw SQL, `loaddata`, and `queryset.update()` bypass m2m signals. Soften to "every writer that goes through the m2m manager" so an authorization invariant isn't stated more broadly than it holds.
- Test-maintenance tax is real: modelling the pre-cutover "legacy-only" state now requires `Grant.objects...delete()` after the add (in `test_roles.py`, `test_grant_authorization.py`, and implicitly in `test_admin.py`). The helper exists in `test_roles.py` but is duplicated inline elsewhere — move it to `accounts/tests/helpers.py` so the transition concept has one home.

### F3 (low) — same-role direct Grant is destroyed with the membership
`unmirror_membership_grants` deletes by `(user, role, org)`. A direct Grant of the same role at the same org is the same row (unique constraint), so removing the role-backed membership also revokes the independently granted role. It's documented in the docstring, but there's no test pinning it and no warning on the admin surfaces that share the row. Cheap to add a test; consider a help-text note in `GrantAdmin`.

### F4 (low) — triple-literal permission message
`PERMISSION_DENIED_MESSAGE` is exported and used by `require_can`/teams, but `accounts/extensions.py` (L63, L97) still hard-codes the same string. Import the constant there too, or the "one string" claim isn't true.

### F5 (nit) — `set(scoped_roles_for_groups(objs))`
In `PermissionGroupAdmin.get_deleted_objects`, `set(dict)` silently means "keys of the mapping". `scoped_roles_for_groups(objs).keys()` (or a set-returning helper for membership tests) reads better.

### F6 (nit) — stale doc reference
`reconcile_org_groups`' docstring still credits `accounts.signals.delete_orphaned_group` for tearing down the `auth.Group`; that receiver no longer exists (teardown is structural via the MTI inheritance, per migration 0007). The diff touches this function area; worth correcting.

### F7 (nit) — `LoadedRowRawIdWidget` drops `using`
`GrantRowForm` reconstructs the widget without the original's `using` kwarg. Multi-DB isn't in play here, so this is fine today; a short comment or preserving `using` would make it robust.

### F8 — teams authz registry canary (repeat)
Still absent. `shelters/tests/test_authz_registry.py` pins every Query/Mutation field to its authorization route precisely because the cutover removed `@hasOrgPerm` and made protection behavioral. Teams did the same and is now in the same state, with no equivalent canary; adding one (4 fields: `teams`, `create_team`, `update_team`, `delete_team`) costs ~20 lines and matches the established pattern. If you're deliberately skipping it, record why in the ADR; otherwise it's the last outstanding item from my previous review.

mikefeldberg (approved):

vecchp:
Implemented — all in #2456 (stacked on #2455; merges right after the chain, #2443 untouched). Per finding:

- **F1** — took option (a): `reconcile_org_groups` unmirrors each stale derived row's members before the delete, so config cleanup now fully revokes; teardown deletes still leave Grants standing. Pinned by `test_reconcile_of_a_stale_group_unmirrors_the_grant` (org types outreach+shelter → shelter; group gone, mirror gone).
- **F2** — receiver docstring now says "every writer that goes through the m2m manager" and names the bypasses (`bulk_create`/raw SQL/`loaddata`/`queryset.update`), with ADR 0001 §4 phase 2 as the sanctioned exception. The helper duplication (iii) was already consolidated in #2451 — `common/tests/utils.py` (`make_permission_group`, `add_legacy_membership`, `make_legacy_only_holder`, `revoke_grants`).
- **F3** — test added: `test_removing_a_membership_revokes_a_same_row_direct_grant`. `GrantAdmin` now carries a fieldset description saying the unique constraint makes a direct grant and the membership mirror the same row, so removing the membership removes it.
- **F4** — moot: `accounts/extensions.py` was deleted by `590dc503` in this cycle; grep confirms no non-test hardcoded literal remains.
- **F5** — `.keys()`.
- **F6** — both `services.py` refs fixed, plus the same stale reference in `admin.py`'s `change_view` docstring found while doing this; all now describe the structural MTI cascade (migration 0007).
- **F7** — `using=getattr(widget, \"using\", None)` preserved.
- **F8** — `teams/tests/test_authz_registry.py` added, same shape as shelters (registry dict + exact set-equality over `__strawberry_definition__.fields`; 4 fields).
- Item 4 (mobile header hook) — resolved upstream: #2450/#2452 retired the `X-Organization-ID` header entirely, so there is no mobile consumer left to migrate.

One observation from doing F3 (no behavior change made — flagging for direction): the admin inline's row-deletion route now only strips the legacy membership — a mirrored Grant survives the row delete by design (the teardown asymmetry the delete-view note already calls out), while the org-type route fully revokes via F1. The `change_view` confirmation counts members on both routes, so on the row-delete route its \"revoked from N members\" can overstate authority loss for grant-holding members. Want the row-delete route treated as config cleanup too (unmirror there as well), or is holding the asymmetry right?

Validation: full backend suite **1890 passed / 31 skipped** on the branch; ruff clean.

PR #2427, opened 2026-09-02
feat(perm): teams mutations authorize via grant OR legacy (§5.3 slice 1)

vecchp:
## What & why

First slice of ADR §5.3 (org-admin role-backed milestone), grouped by endpoints
(**teams** operations only — reports and member management follow as separate slices;
the provisioning/backfill PR retires the transitional arm).

Teams write perms (`add_team`/`change_team`/`delete_team`) ride the **legacy
`ORG_ADMIN` template** — no scoped `Role` row exists, so no holder has a `Grant` and
the pure-Grant predicate returns nothing. A grants-only flip would strip authority
from every org admin at every org. This slice therefore converts the three team
mutations to a transitional **`HasOrgPermOrGrant`** extension:

- **legacy arm** — `permissioned_queryset` (exactly what `HasOrgPerm` checks today);
  run first so the common path's query count is unchanged (pinned 7/11/7 hold).
- **grant arm** — `can()` (the end-state predicate); dormant until the §5.3
  provisioning PR role-backs the template + backfills Grants.

The predicate stays pure-grant; only the consumer carries the transitional arm, and
it is deleted in the provisioning PR.

## TDD
New `teams/tests/test_grant_authorization.py` — grant-arm tests **failed before**
the extension (grant holder denied under legacy-only code), pass after:
- legacy `ORG_ADMIN` holder (no Grant) still manages teams;
- scoped-Grant holder (no legacy group) manages teams via the grant arm;
- update/delete thread CHANGE/DELETE (ADD-only holder denied);
- plain member denied; Grant at org A does not authorize org B.

## Changes
- `accounts/extensions.py` — `HasOrgPermOrGrant`
- `teams/schema.py` — swap the three mutations
- `teams/tests/test_grant_authorization.py` — new
- `schema.graphql` — regenerated (`@hasOrgPermOrGrant` on the three team mutations
  + new directive definition). FE types need regen in a node-enabled checkout.

## Verification
- teams suite: 65 passed; affected-apps regression (accounts/common/shelters/notes/
  tasks/reports/teams): 1289 passed
- ruff format/check clean; mypy strict clean

## Summary by Sourcery

Enable transitional grant-or-legacy authorization for team mutations without removing existing organization-admin access.

New Features:
- Allow team creation, updates, and deletions to be authorized by scoped grants alongside legacy organization permissions.

Enhancements:
- Preserve legacy organization-admin access during the transition to grant-based authorization while enforcing operation-specific permissions and organization scoping.

Tests:
[...]

PR #2444, opened 2026-09-09
feat(perm): reports is grant-only & header-free — role-backed ORG_ADMIN carries reports.view_reports

vecchp:
## What & why

The **reports** slice of the org-admin grant cutover (ADR 0001 §5.3) — the
second domain to ride the role-backed `ORG_ADMIN`/`ORG_SUPERUSER` grants that
landed for teams in #2443. Reports is grant-only **and header-free** (a full
cutover — it is a web feature, so no deprecated header fallback is needed):

- `reportSummary` (GraphQL) authorizes via `require_can` at the **payload org**
  (`organizationId` argument — no header is read).
- The DRF interaction-data export authorizes via `can()` at the `org_id` query
  param (already payload-based).
- Legacy-only holders fail closed; membership is no longer consulted. Member
  management (`organizations.*`) is now the **only** remaining legacy-only
  org-admin domain.

> **Rebase note.** #2443 is merged; this branch was rebased onto `main` — the superseded teams commits dropped, the reports commits (plus their phantom-row/typecheck fixups) replayed.

## The unlock: RoleDef permissions bind to real models

`accounts/seed._resolve_permissions` used to derive a RoleDef permission's
ContentType from the codename's last `_` token, so `reports.view_reports`
(bound to `ScheduledReport`) tripped `sync_roles`' phantom-ContentType guard.
It now binds each codename, in order, to:

1. an existing Permission row on a real model's ContentType;
2. the model that **declares** the codename in `Meta.permissions` — resolved
   from the app registry, so provisioning never depends on `create_permissions`
   having run for that app yet (post_migrate fires per-app);
3. the synthesized fallback (unchanged) for portal codenames no model declares
   (`organizations.*`), which keeps legacy template provisioning working.

## Changes

- **`accounts/seed.py`** — `_resolve_permissions` real-model binding (above) +
  `retire_superseded_phantom_permissions()`: idempotently deletes phantom
  Permission/ContentType rows superseded by a real twin (runs at post_migrate;
  keeps `organizations.*` phantoms, which still have no real twin).
- **`accounts/groups.py`** — `ORG_ADMIN_ROLE_PERMISSIONS` gains
  `reports.view_reports` (flows to `ORG_SUPERUSER` too). Existing backfilled
  org-admin Grants inherit it from the Role row on the next `sync_roles` — no
  re-backfill needed.
- **`reports/schema.py`** — `reportSummary(organizationId: ID!, …)`: the org is
  carried in the payload (header-free); `require_can(…, reports.view_reports)`
  at that org; unknown/non-numeric org ids fail closed (no `ValueError`).
- **`reports/permissions.py`** — `HasReportAccess` (DRF export): `can(…, org)`
  at the `org_id` param, failing closed on unknown/non-numeric org ids.
- **`reports/models.py`** — `ScheduledReport` declares `OrgScoped` (no
  migration; required by `permissions.E005` for the scoped Role).
- **`common/permissions/domain.py`** — `reports` joins `LEGACY_INERT_APPS`
  (legacy rows not reported; the global tier folds like teams/shelters).
- **`accounts/apps.py`** — post_migrate runs the phantom retirement after the
  backfills.
- **FE** (`betterangels-admin` Reports page) — `reportSummary` query and page
  pass `activeOrg.id` as `organizationId` (the only consumer of the reports
  surface).
- **Tests** — reports grant-auth contract (role-backed ORG_ADMIN reads +
  exports; legacy-only denied; direct-grant holder without membership; member
  denied; cross-org denied; superuser; plus headerless read, stale-header
  ignored, unknown/non-numeric org id fails closed); `test_views.py` updated to
  the payload org; reachability equivalence keyed on `LEGACY_INERT_APPS`;
  phantom-retirement test; roles phantom/E005 tests updated.
- **`schema.graphql`** + codegen regenerated (`reportSummary` gains the
  required `organizationId`).
- **`docs/adr/0001`** — §5.3 status notes reports joined, header-free.

## Adversarial-review findings addressed in this PR

- Header → payload full migration for `reportSummary` (no deprecated fallback).
- DRF `org_id` / schema `organizationId` fail closed on unknown **and
  non-numeric** ids (no `ValueError`).
- Residual phantom Permission/ContentType rows (superseded by real-model
  binding) are retired idempotently at post_migrate.
- Left for later (noted): Django `ScheduledReportAdmin` is still legacy-gated
  (`has_perm` never reads Grants — pre-existing design to revisit at teardown).

## Verification

- Full backend suite green at the branch head (1754 passed / 31 skipped).
- `ruff check` + `ruff format --check` clean on all touched files.
- Codegen re-run clean across all schema-consuming projects.
- FE typecheck/eslint run in CI.

## Next slice

Member management (`organizations.*`) — needs those portal codenames bound to a
real model before they can ride the Role; after that the ORG_ADMIN legacy group
can be retired wholesale.

## Summary by Sourcery

Move reports authorization fully to organization-scoped grants and carry the target organization explicitly in the request payload.

New Features:
- Authorize reports access through role-backed ORG_ADMIN/ORG_SUPERUSER grants for GraphQL summaries and DRF exports.
- Make the GraphQL report summary organization-targeted and header-free.

Bug Fixes:
- Fail closed for unknown or non-numeric organization identifiers instead of raising lookup or conversion errors.
- Retire superseded phantom report permissions while preserving existing user, role, group, and template grants.

Enhancements:
- Bind declared role permissions to their real model ContentTypes and classify reports as a grant-only permission domain.

Documentation:
- Document the reports grant-only cutover and its remaining member-management boundary in ADR 0001.

Tests:
- Add coverage for grant-based reports authorization, cross-organization isolation, legacy-only denial, header independence, invalid identifiers, and phantom-permission retirement.

Chores:
- Regenerate the GraphQL schema and reports client types for the required organizationId argument.


mikefeldberg:
## Findings

### Should address before merge
1. **The both-rows guard has no regression test** — `accounts/seed.py:252-259`; commit `e6af97dc2` added 18 lines and zero tests. Only the phantom-only-holder case is covered (`test_template_permissions.py:83-104`). The guard exists because a holder referencing both rows made `.update()` collide with the through-table unique constraint and **abort post_migrate** — a one-line refactor away from a redeploy-blocking bug. Add a parametrized case (holder with phantom+real) to the retire test.
2. **Hand-defined roles carrying `reports.view_reports` silently lose access** — backfills convert only code-owned roles (`backfill_org_admin_grants` → `ORG_ADMIN_ROLES`; shelters/caseworker likewise), and the m2m mirror only fires for templates with a same-named scoped `Role` (`role_manager.py:36-39`). A legacy admin-created group/template carrying reports gets **no Grant**, and `HasOrgPerm` is no longer consulted → revocation on deploy. Audit for these (or explicitly document the accepted revocation in the ADR — the migration matrix doesn't mention it).
3. **`retire` deletes broader than its stated purpose** — `accounts/seed.py:268-270` drops *every* model-less ContentType left with zero permissions, not just CTs of retired phantom rows. That sweeps unrelated stale CTs (e.g. apps no longer installed) and cascades into `django_admin_log`. Restrict to the CTs whose permissions were actually retired, or justify the sweep in the docstring.

### Worth considering
4. **N+1 in `real_by_key`** (`seed.py:245-247`): reads `p.content_type.app_label` per row without `select_related("content_type")` — one query per real Permission on **every post_migrate** (the other loops in this module use `select_related`).
5. **Duplicate through tables** (`seed.py:230-236`): `Role.permissions.through`, `PermissionGroup.permissions.through`, and `Group.permissions.through` are all `auth_group_permissions` (MTI subclasses of `Group`) — the loop re-points it 4×. Dedupe with `dict.fromkeys`. Also the docstring's "References are re-pointed, never silently dropped" overpromises: only the enumerated M2M throughs are re-pointed; any other FK to `Permission` still cascades.
6. **`id_lookup` determinism edge** (`seed.py:~113-121`): "prefer real" among multiple real rows is last-write-wins arbitrary, while `ct_by_key` keeps the first — if two real rows ever share `(app_label, codename)`, provisioning binds one CT and returns a different row's id. Prefer the row whose ContentType is `ct_by_key[(a, c)]`.
7. **Fail-closed org resolution is now triplicated** — `_org_or_deny` (`reports/schema.py:24-38`) and `HasReportAccess` (`reports/permissions.py:47-53`) repeat the same resolve-or-deny rule (teams/shelters have their own). `python.md` says the API layer should delegate to selectors; a shared helper is warranted now that this is the third copy.

### Test hygiene
8. `test_legacy_only_org_admin_is_denied` (`test_grant_authorization.py:164`) can pass vacuously — it never asserts the legacy group actually **carries** `reports.view_reports`. Assert `group.permissions.filter(...)` so the test proves the revocation it claims.
9. DRF export fail-closed paths are untested: unknown/non-numeric `org_id` → 403. The GraphQL equivalents are pinned (`test_grant_authorization.py:138,147`), as is a member-without-authority export; add the export-side edge cases. No test covers a `user_permissions`-holder enforcement either.
10. Leftovers: `test_views.py:393,425,466` still call `_set_active_org` (now no-ops); `test_stale_header_is_ignored` relies implicitly on the base fixture's header — assert it in-test so it can't silently degrade into a duplicate of the payload test.

### Comment/doc drift (stale after this change)
11. `accounts/groups.py:56` — "Same bundle as ORG_ADMIN while **teams is the only scoped permission**" (reports now rides too).
12. `accounts/services.py:550` — the guard's error hint "codename must end in the model name" is no longer the binding rule; name the real fix (declare it in `Meta.permissions`) or failure diagnosis misleads.
13. `common/permissions/domain.py` module docstring still says dual is "Empty until PRs #2427-2429" and cites shelters as the only grant-only example; the DUAL block comment was updated, not the prose.
14. `accounts/groups.py` "kept (dual write) **until reconcile retires them**" — `reconcile_org_groups` (`services.py:284+`) only deletes stale *derived* groups today; word it toward the teardown slice or the ADR.

vecchp:
All 14 points verified against `332d76f2` and addressed in `957cad7d` — thanks, this was a sharp review. Locally: full backend suite **1852 passed / 31 skipped**, `ruff check` + `format --check` clean — CI is re-running on the new head.

**Before merge**
1. The both-rows guard now has regression coverage: a user *and* a group each hold the phantom + real rows, and the test pins that the phantom reference is dropped (no through-table collision) and the real one survives.
2. Hand-defined roles: documented as an accepted revocation in ADR §5.3 with the audit query (non-`REGISTRY` templates + template-less groups). Running it against prod before this deploys; if it returns holders we re-grant via a role-backed template.
3. The orphan sweep is now limited to the ContentTypes this retire actually emptied — unrelated model-less CTs (and their `django_admin_log` FKs) are no longer touched.

**Worth considering**
4. `select_related("content_type")` on the real-row lookup.
5. Through list deduped with `dict.fromkeys` (`Group`/`Role`/`PermissionGroup` are the same `auth_group_permissions` table); the docstring no longer overpromises — it names the M2Ms that get re-pointed and notes that direct FKs to `Permission` (guardian object-permission tables) still cascade.
6. Binding now prefers the row on the provisioned ContentType, which makes the duplicate-codename edge deterministic.
7. Both fail-closed copies now use the house `get_or_none` guard (same as teams) instead of a hand-rolled try/except.

**Test hygiene**
8. The legacy-only test asserts the group carries `reports.view_reports` before the denial (no more vacuous pass).
9. DRF unknown / non-numeric `org_id → 403` pinned; `user_permissions`-holder enforcement added (GraphQL + DRF).
10. Dead `_set_active_org` calls dropped; the stale-header test pins the header it relies on.

**Doc drift**
11–14. Fixed (`groups.py` ×2, `services.py` error hint, `domain.py` docstring), plus the §4.1 matrix row for reports no longer says "no scoped `Role` row yet".

vecchp:
## Post-deploy audit — reports cutover (hand-defined roles + retire sanity)

As promised in the dispositions above, here is the exact runbook. All read-only; run on a console once this ships (`python manage.py shell`). If **#1** comes back empty, the accepted-revocation note is a proven no-op. Snippets are self-contained — copy one at a time.

### 1. Ground truth — who lost reports access

```python
from accounts.models import PermissionGroup
from common.permissions.selectors import can

PERM = {"permissions__content_type__app_label": "reports", "permissions__codename": "view_reports"}
affected = set()
for group in (
    PermissionGroup.objects.filter(**PERM).select_related("organization", "template").prefetch_related("user_set")
):
    org = group.organization
    for user in group.user_set.all():
        if user.is_active and not can(user, "reports.view_reports", org=org):
            affected.add((org.name, user.username, group.template.name if group.template else "<no template>"))

for row in sorted(affected):
    print(" | ".join(row))
print(f"{len(affected)} affected user(s)")
```

**Expected: `0 affected user(s)`.** Any rows are users who held reports through an admin-created (or template-less) group; re-grant them via a role-backed template, or accept the revocation knowingly. (Deactivated users are skipped, as shown.)

### 2. Custom roles still carrying the permission (the population behind #1)

```python
from accounts.models import PermissionGroupTemplate
from common.org_types import REGISTRY

PERM = {"permissions__content_type__app_label": "reports", "permissions__codename": "view_reports"}
for name in (
    PermissionGroupTemplate.objects.filter(**PERM)
    .exclude(name__in=REGISTRY.template_names())
    .values_list("name", flat=True)
    .distinct()
):
    print(name)
print("done")
```

**Expected: just `done`.**

### 3. Retire sanity (once, after the migrate that ships this PR)

```python
from django.contrib.auth.models import Permission
from django.contrib.contenttypes.models import ContentType

print(ContentType.objects.filter(app_label="reports", model="reports").exists())  # expect False
print(Permission.objects.filter(content_type__app_label="reports", codename="view_reports").count())  # expect 1
print(Permission.objects.filter(content_type__app_label="organizations", codename="add_org_member").exists())  # expect True (portal phantom kept)
```

Expected as commented: the synthesized `reports/reports` ContentType is gone, exactly one `view_reports` Permission remains (the `ScheduledReport` row), and the member-management portal phantoms survive.

Same shape will apply to the member-management cutover — it is the last domain that needs this check.

mikefeldberg (approved):

PR #2449, opened 2026-09-09
shelter contacts ui (SDB-277)

tglaz:
### Private Fields - Additional Contacts
SDB-277

#### View Mode

image

#### Edit Mode

image

#### With faked email error (submit invalid email to server)

image

## Summary by Sourcery

Add a feature-flagged BA Private shelter profile workflow for managing additional contacts with robust validation and indexed server-error handling.

New Features:
- Add a feature-flagged BA Private shelter profile page for viewing and editing additional contacts.
- Support adding, removing, and editing contact details including name, phone, email, title, and claimant status.

Bug Fixes:
- Improve GraphQL field-error handling for indexed additional-contact fields while preserving generic handling for unexpected fields.
- Replace the loose phone validation with US-specific NANP phone validation.

Enhancements:
- Extend shelter profile data loading and routing to support BA Private additional contacts.
- Expose inline validation and server-side errors for additional contact forms.

Tests:
- Add coverage for plain, nested, indexed, mixed, and unrecoverable GraphQL field-error scenarios.
- Expand US phone regex validation test cases.

Chores:
- Update test parameter formatting and related generated configuration changes.


libs/react/shelter-operator/src/lib/OperatorApp.tsx
```diff
@@ -113,6 +114,10 @@ export function OperatorApp() {
                 path={profileRouteConfig.children.media}
                 element={<ShelterMediaPage />}
               />
+              <Route
+                path={profileRouteConfig.children.baPrivate}
+                element={<BaPrivatePage />}
+              />
```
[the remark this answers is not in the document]
tglaz: this will get resolved with permissions

libs/react/shared/src/lib/static/regex.ts
```diff
@@ -5,8 +5,9 @@ export const Regex = {
   email: /^[a-zA-Z0-9._%+-]+@[a-zA-Z0-9.-]+\.[a-zA-Z]{2,63}$/,
   phoneNumber: /^[2-9]\d{2}[2-9]\d{6}$/,
   phoneNumberWithExtensionUS: /^[2-9]\d{2}[2-9]\d{6}(x\d+)?$/,
-  /** Accepts common phone formats: digits, spaces, dashes, parens, leading +, and extensions via x/ext/extension. */
-  phoneNumberLoose: /^\+?[\d(][\d\s\-().]*(\s*(x|ext|extension)\.?\s*\d+)?$/i,
+  /** US phone with loose formatting: optional +1/1 prefix, spaces/dashes/dots/parens, a valid 10-digit NANP number, and optional x/ext/extension. */
+  phoneNumberLooseUS:
+    /^\+?1?[\s.-]*(?:\([2-9]\d{2}\)|[2-9]\d{2})[\s.-]*[2-9]\d{2}[\s.-]*\d{4}(?:\s*(?:x|ext|extension)\.?\s*\d+)?$/i,
```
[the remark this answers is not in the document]
tglaz: this is intentionally a loose Regex to match server validation as close as possible - but server will do final validation. 99% of the time this will catch issues - can reference tests

libs/react/shelter-operator/src/lib/components/ShelterProfile/segments/BaPrivate/ShelterBaPrivate.tsx
```diff
@@ -0,0 +1,170 @@
...
+  if (!shelter) {
+    return null;
+  }
+
+  return (
+    <div className="px-6 flex-col flex-1 pb-48">
+      <Form className="flex-1">
+        <Form.Header
+          title="BA Private"
+          onEditClick={!isEditMode ? () => setEditMode(true) : undefined}
```
vecchp: Every other pencil-bearing segment gates this (`canEditShelter ? () => setEditMode(true) : undefined`). As written, a view-only operator can open the editor and only fails server-side on save. Suggest `const { canEditShelter } = useShelterPermissions();` and the same conditional.

libs/react/shelter-operator/src/lib/components/ShelterProfile/segments/BaPrivate/ShelterBaPrivate.tsx
```diff
@@ -0,0 +1,170 @@
...
+  type AdditionalContactsFormData,
+} from './formSchema';
+
+function toUpdateInput(
+  shelterId: string,
+  data: AdditionalContactsFormData,
+): UseUpdateShelterProfileInput {
+  return {
+    id: shelterId,
+    additionalContacts: data.additionalContacts.map((contact) => ({
```
vecchp, line 32: This is a full-replacement payload — any server row missing from `data.additionalContacts` is deleted (documented on `ShelterContactInfoInput`). Removing a row and saving is therefore destructive with no confirm/undo, while sibling delete flows use a modal (`DeleteScheduleModal`). At minimum, worth a confirmation when rows that came from the server are being dropped.

libs/react/shelter-operator/src/lib/components/ShelterProfile/segments/BaPrivate/ShelterBaPrivate.tsx
```diff
@@ -0,0 +1,170 @@
...
+
+  const { shelter } = useShelterOperatorProfile(shelterId);
+  const { updateShelter } = useUpdateShelterProfile();
+  const { showToast } = useToast();
+
+  const { control, handleSubmit, setError, reset } =
+    useForm<AdditionalContactsFormData>({
+      resolver: zodResolver(formSchema),
+      defaultValues: defaultFormValues,
+      mode: 'onBlur',
```
vecchp, line 62: `mode: 'onBlur'` is inert here: the row inputs never call RHF's `onBlur` (the row `Controller` wires only `value`/`onChange`), so validation effectively runs on submit only. Either wire `onBlur` through the row fields, or use `mode: 'onSubmit'` and say why — `ScheduleForm` documents that choice.

libs/react/shelter-operator/src/lib/components/ShelterProfile/segments/BaPrivate/Contacts/ShelterContactForm.tsx
```diff
@@ -0,0 +1,96 @@
...
+          label="Email"
+          dataType="email"
+          value={entry.contactEmail}
+          onChange={(e) => onChange({ contactEmail: e.target.value })}
+          isViewMode={isViewMode}
+          error={errors?.contactEmail?.message}
+        />
+
+        <Input
+          label="Title"
```
vecchp, line 82: `contactTitle` is the one field that can fail server-side with no visible message: `contact_title` is `max_length=255`, this input has no `error` prop, and `ContactFormErrors` has no key for it — the BE's `additionalContacts.<i>.contactTitle` is set but only surfaces as the toast. Deterministic repro: save a 256-char title. Add `error={errors?.contactTitle?.message}` (+ the key), and a `maxLength` would prevent the common case.

vecchp (commented):
## Review — shelter contacts UI (SDB-277)

Reviewed the FE flow end-to-end against the backend contract (#2442) and the two shared libs it touches. No blockers; specifics inline. Evidence for the conclusions, so the reasoning is auditable:

**Feature flag** — `ffShelterOperatorBaOnlyFields` matches BE `BA_ADMIN_ONLY_FIELDS_FLAG`. BE returns `[]` from `operatorShelter.additionalContacts` and rejects writes with `PermissionDenied` when off, so the FE flag is presentation-only — no escalation path.

**Error contract** — BE emits indexed camelCase paths (`additionalContacts.0.contactEmail`, pinned by `UpdateShelterAdditionalContactsErrorShapeTestCase`); the first-segment match in `filterRecoverableOperationMessages` is what makes them land; RHF matches the parent subscription for child `setError` calls (`shouldSubscribeByName`), and `_setFieldArray` shifts indexed errors on row removal, so indices stay correct.

**Post-save sync** — the mutation selects only `id` on success, so the refetch is the sync mechanism; Apollo's `refetch()` forces `network-only` (4.2.8), then `reset(toFormData(shelter))` re-seeds the form. Adds/removes show up without a manual reload.

**Data exposure / ids** — parsed `schema.graphql`: `additionalContacts` exists only on `OperatorShelterType` (not the public `ShelterType`). `ShelterContactInfoType.id` is the raw pk (no relay), so the full-replacement `id` round-trip and `Number(id)` sort are valid.

**CI** — head 74994fbc, all checks green (`nx affected -t test` + schema/type drift checks included). I could not run vitest locally (container's node_modules is missing the linux-arm64 rolldown binding), so test evidence is CI.

### Suggested before merge
- Title error rendering (inline) — deterministic repro, small fix.
- `canEditShelter` gate (inline) — every other pencil-bearing segment has it.

### Follow-ups
See inline: destructive replace without confirm · inert `mode: 'onBlur'` — flag-gated route · test placement in the shared utils.

### Tests worth adding
A unit test for `toFormData`/`toUpdateInput` (id round-trip, `''`→`null` coercion) and one RHF-level test that a nested server error renders per row — the manual fix in "update contactErrors handling" is exactly what a test should hold in place.

libs/react/shelter-operator/src/lib/components/ShelterProfile/segments/BaPrivate/Contacts/ShelterContacts.tsx
```diff
@@ -0,0 +1,99 @@
...
+            <Controller
+              key={field._rhfId}
+              name={`additionalContacts.${index}`}
+              control={control}
+              render={({ field: { value, onChange } }) => (
+                <ShelterContactForm
+                  entry={value}
+                  onChange={(patch) => onChange({ ...value, ...patch })}
+                  onRemove={() => onRemove(index)}
+                  errors={contactErrors}
```
vecchp, line 75: `useFormState({ name: 'additionalContacts' })` is the right seam (child `setError` paths match the parent subscription), but two recovered paths still have no render target: `additionalContacts.<i>.id` ("Invalid/Unknown additional contact id." — reachable from a stale tab or concurrent delete, since every row's `id` is resubmitted) and the object-level "Invalid additional contact." `ContactFormErrors` covers neither. `ExceptionRow` (Schedules) renders a row-level fallback for exactly this class — worth mirroring. (Indexed errors do shift correctly on `remove`, so stale row indexes aren't a concern.)

libs/react/shelter-operator/src/lib/OperatorApp.tsx
```diff
@@ -113,6 +114,10 @@ export function OperatorApp() {
                 path={profileRouteConfig.children.media}
                 element={<ShelterMediaPage />}
               />
+              <Route
+                path={profileRouteConfig.children.baPrivate}
```
vecchp, line 121: The route is registered unconditionally while the sidebar link is flag-gated, so a not-flagged user hitting the URL directly gets an empty contacts page plus an Edit affordance, and the save fails with a permission toast. The BE is the real gate, so this isn't a security gap — but consider gating here (redirect to `basic`) so the UI matches the flag.

libs/ba-platform/src/lib/apollo/graphql/response/utils/filterRecoverableOperationMessages.ts
```diff
@@ -16,11 +16,19 @@ type PartitionResult = {
...
  * - `field` is non-null, and
- * - `field` appears in `allowedFields`.
+ * - `field`'s first dotted segment matches an entry in `allowedFields`
+ *   (e.g. `additionalContacts.0.contactEmail` matches `additionalContacts`;
+ *   a plain `name` matches `name`).
  *
  * Everything else — other kinds, missing field, or field outside the filter —
  * is **unrecoverable** and must be handled by the caller (e.g. thrown).
  */
+function matchesAllowedField(field: string, allowedFields: string[]) {
```
vecchp: Right seam for the BE contract (`_contact_error_dict` → `additionalContacts.<i>.<field>`), and safe for existing callers — they all pass top-level names (`Object.keys(formSchema.shape)` / `['additionalContacts']`), so this is a strict superset of the old exact match. Two suggestions: (a) add a nested case to `getFieldErrorsOrThrow.test.ts`, which currently only covers top-level fields, so the intent is pinned at the public API; (b) recovered nested paths are only as good as the form's render coverage — see the Title/`.id` comments in this PR.

libs/react/shared/src/lib/static/regex.ts
```diff
@@ -5,8 +5,9 @@ export const Regex = {
   email: /^[a-zA-Z0-9._%+-]+@[a-zA-Z0-9.-]+\.[a-zA-Z]{2,63}$/,
   phoneNumber: /^[2-9]\d{2}[2-9]\d{6}$/,
   phoneNumberWithExtensionUS: /^[2-9]\d{2}[2-9]\d{6}(x\d+)?$/,
-  /** Accepts common phone formats: digits, spaces, dashes, parens, leading +, and extensions via x/ext/extension. */
-  phoneNumberLoose: /^\+?[\d(][\d\s\-().]*(\s*(x|ext|extension)\.?\s*\d+)?$/i,
+  /** US phone with loose formatting: optional +1/1 prefix, spaces/dashes/dots/parens, a valid 10-digit NANP number, and optional x/ext/extension. */
+  phoneNumberLooseUS:
```
vecchp, line 9: Aligned with the BE (`PHONE_NUMBER_REGEX = ^[2-9]\d{2}[2-9]\d{6}(x\d+)?$`) and it round-trips the serialized wire format (`national_number` digits, `...x<ext>`). Note the blast radius: `CreateShelterProfile` reuses `ShelterBasicInfoForm`, so shelter creation is now US-only too — likely intended, worth a line in the PR description.

libs/react/shelter-operator/src/lib/components/ShelterProfile/segments/BaPrivate/formSchema.ts
```diff
@@ -0,0 +1,35 @@
...
+
+export const defaultFormValues: AdditionalContactsFormData = {
+  additionalContacts: [],
+};
+
+export function toFormData(
+  shelter: ShelterProfileType,
+): AdditionalContactsFormData {
+  // Sort by the numeric id so the list renders in deterministic creation order.
+  const additionalContacts = [...(shelter.additionalContacts ?? [])].sort(
```
vecchp, line 30: `Number(a.id)` works today (the type's `id` is the raw pk, so numeric strings), but the comparator silently degrades to a NaN no-op if ids ever change shape. The BE already returns rows in pk order, so this could be dropped if you don't need the guarantee.

libs/react/shelter-operator/src/lib/components/ShelterProfile/segments/BaPrivate/Contacts/ShelterContactForm.tsx
```diff
@@ -0,0 +1,96 @@
...
+      <div className="min-h-[48px] mb-4 flex items-center justify-between w-full">
+        <Text variant="body-bold" className="pl-5">
+          Contact
+        </Text>
+
+        {!isViewMode && (
+          <Button
+            variant="trash"
+            onClick={onRemove}
+            aria-label="Remove contact"
```
vecchp: Multiple rows render an identical `aria-label="Remove contact"`; including the contact name or index keeps the control distinguishable for screen readers.

shiozakiyuko (approved):
It looks good to me. Approved.

libs/react/shelter-operator/src/lib/OperatorApp.tsx
tglaz: I just did a quick/dirty gating here.
i'll do a PR next to implement gating based on permissions and remove the flag - and will update all of this

libs/react/shelter-operator/src/lib/components/ShelterProfile/segments/BaPrivate/ShelterBaPrivate.tsx
tglaz: the `PUT` functionality is deliberate - thought about doing PATCH and perhaps doing updates against the model directly instead of using the updateShelter mutation, but figured that's an update for later (discussed with Mike also)

if we want some `confirmation` then I'd suggest moving that part into a separate ticket and then could fully flesh out the best UX treatment. and if we want this then I'd update to doing operations against individual records. that would clean up a lot of other complexities related to working against an array.
to drop a field a user would need to clear it in the form, which is already quite deliberate.

libs/react/shelter-operator/src/lib/components/ShelterProfile/segments/BaPrivate/Contacts/ShelterContactForm.tsx
tglaz: oh yeah - I totally forgot about maxLen that's why I left it. will update

libs/ba-platform/src/lib/apollo/graphql/response/utils/filterRecoverableOperationMessages.ts
tglaz: I updated `getFieldErrorsOrThrow` so we can explicitly pass recoverable fields for nested objects.
`bad ID` error isn't recoverable so a default message should be ok for now

vecchp (approved):

PR #2450, opened 2026-09-10
feat(perm): retire the X-Organization-ID header — mobile teams cutover (DEV-2566)

vecchp:
## Summary

Completes the **`X-Organization-ID` header retirement** (DEV-2566): mobile's teams pickers now pass the active org in the query variable, so the teams list read — the last `get_current_organization` consumer — takes its org from the payload, and the header plumbing is deleted backend-side.

Stacks on `feat/perm/core-hygiene` (#2448) — land after the teams→reports→members→teardown→write-tiers→hygiene stack.

## What changes

**Mobile (`libs/expo`)**
- `useOrgTeams` reads the active-org store reactively (`useSyncExternalStore` over `getActiveOrgId` / `subscribeActiveOrgId`) and sends it as `TeamFilter.organizationId`. The store is the same one the fetch interceptor read for the header, so the query can never run against a stale org, and it is available before the org list loads.
- The query is **skipped while no org is active** (none remembered, or the user has none) instead of issuing a request the backend would deny.
- Tests now pin the operation variables: every request names the active org; the remembered org wins over the first one; the org is available before the org list loads; the hook stays idle with no org.

**Backend — header retired**
- The teams read is **filter-only** (no fallback): a caller who names no org, an unknown org, or a blank/null/non-numeric id denies with the standard `You do not have access to this organization.` refusal. `_org_or_deny` maps `strawberry.UNSET` (an omitted optional filter field) onto the deny path; `_resolve_read_org` is gone.
- Deleted: `get_current_organization` (`common/permissions/utils.py`), `OrganizationMiddleware` (+ its `settings.py` entry and `common/middleware/__init__.py` export), and `common/tests/test_org_context.py`.
- Every vestigial `_set_active_org` / header setup is removed from the suites (teams, reports, shelters, notes, accounts, common), and the header-era tests are rewritten to the filter-only contract.

**Docs**
- ADR 0001 §2.3 / §2.6 / §5.3 / §7 and backend `docs/permissions.md` describe the retired header; the §7 "header retirement — remaining scope" note is updated to done.

## Behavior notes / review focus

- Mobile queries wait until an org is known (`skip`) — a change from "issue and get denied". If the user has no org at all the teams pickers simply stay empty with no error.
- No API/schema change: the schema export is **byte-identical**.
- FE may keep sending the (now ignored) `X-Organization-ID` header via its global `createOrgInterceptor`; dropping it is cosmetic cleanup, not correctness.

## Test plan

- Full backend suite: **1855 passed, 31 skipped**.
- `manage.py check` clean; `ruff` clean; schema export byte-identical.
- Mobile: ESLint clean. `vitest` cannot run in this container (rolldown native binding missing) — CI runs the mobile tests.

## Summary by Sourcery

Complete the organization-header retirement by moving mobile teams reads to payload-scoped organization variables and removing backend request-header support.

New Features:
- Pass the active organization ID in mobile teams queries and skip requests when no organization is active.

Bug Fixes:
- Prevent teams reads from relying on stale, missing, malformed, or unauthorized organization headers by requiring the organization in the query payload.

Enhancements:
- Retire backend organization-header plumbing and enforce payload-based organization resolution across teams authorization.
- Remove obsolete active-organization test fixtures and update authorization coverage for the header-free contract.

Documentation:
- Update the authorization ADR and permissions documentation to mark request organization-header retirement as complete.

Tests:
- Expand mobile and backend tests to verify operation organization variables, remembered-org behavior, no-org idle behavior, and fail-closed teams reads.

Chores:
- Delete the organization middleware, current-organization helper, and related context tests.


vecchp:
Rollout ordering note (compat audit before this lands):

- The **mobile half of this PR is extracted into #2458** and ships first. It's already compatible with today's backend — the teams read prefers the filter and falls back to the header — so old and new binaries both work while it deploys.
- **Please hold the backend half** (header + middleware removal) until #2458's build is released and adopted. Landing it beforehand breaks the teams read — and the TaskForm team picker — on released binaries with `You do not have access to this organization.`
- When rebasing onto #2458, **drop this PR's FE hunks** (`useOrgTeams.ts`, `useOrgTeams.test.tsx`) — they're that PR.
- #2452 (stop sending the header) must also land **after** the backend retirement here, not before — the interceptor is still the fallback for older binaries until then.

PR #2461, opened 2026-09-10
use perms instead of ffShelterOperatorBaOnlyFields (SDB-277)

tglaz:
### Private Fields - Additional Contacts - BE perms
SDB-277

## Summary by Sourcery

Replace feature-flag gating for BA-only additional contacts with global-tier ContactInfo permission checks.

New Features:
- Add a global-tier permission selector for distinguishing global authority from scoped grants.
- Control additional-contact visibility and updates through global ContactInfo permissions rather than a feature flag.

Bug Fixes:
- Prevent users with scoped grants, including scoped ContactInfo roles, from viewing or editing BA-only additional contacts.

Enhancements:
- Assign ContactInfo permissions to the Global Shelter Operator role while keeping them off the scoped Shelter Operator role.

Tests:
- Add coverage for global, scoped, granted-scoped, unauthorized, and superuser permission behavior across selectors, services, and GraphQL operations.

Chores:
- Remove the obsolete additional-contacts feature-flag constant and related flag-based checks.


vecchp (commented):
**Permissions review — global-tier gate is the right call; two deploy items**

Verified the mechanism against the grant model and walked every exposure path on this head:

**What's right**

- `holds_globally` is the correct instrument for SDB-277: the requirement is "a scoped grant must never satisfy this", and no scoped predicate can express a negation of its own tier. Worth pinning the near-miss in the docstring, because it's tempting: **`can_anywhere` admits scoped holders** (`ALL or s.exists()`) and would look equivalent while re-opening exactly the hole this closes — any scoped role carrying the ContactInfo perms would pass. Suggested line: *"sibling of `can_anywhere`, which admits scoped holders — use this only when scoped authority must never satisfy the gate."*
- Exposure audit on this head: `additional_contacts` is update-only (`shelters/types/inputs.py:206` — not on the create input), and the only two GraphQL paths — `updateShelter` (`shelters/schema.py:145`) and `OperatorShelterType.additional_contacts` (`shelters/types/outputs.py:289`) — are both gated. Admin is staff-side. No bypass found.
- Tests: the error-shape class logging in as GSO (so it exercises contact validation, not the gate) is the right adjustment.

**Before deploy**

1. **FE still reads the old flag** — `libs/react/shelter-operator/src/lib/components/AppSidebar/ShelterProfileLinks.tsx:21` gates on `FeatureFlags.SO_BA_ONLY_ADMIN_FIELDS`. With the BE no longer consulting it, the rollout interleaves: flag ON + scoped operator → FE shows the UI, BE returns `[]` and denies saves; flag OFF + GSO → FE hides UI the BE would allow. Suggest keeping the flag ON in all environments until the FE moves to permission-based gating (the BaPrivate contacts work, #2449, already carries that TODO) — or landing those together.
2. **Prod membership check** — access now keys on a global role carrying the ContactInfo perms (GSO). Worth confirming every current BA user who had access under the flag holds that global role, otherwise it's a silent access loss on deploy.

**Nit (non-blocking)**

- `can_globally` would match the `can`/`can_obj`/`can_anywhere` grammar; `holds_globally` is fine if the tier-statement name is deliberate.

LGTM otherwise — nice tightening.

shiozakiyuko (approved):

PR #2464, opened 2026-09-11
fix(perm): make Role provisioning resilient to auth.Group name collisions

vecchp:
## Incident

The 2026-09-11 deploy's `post_migrate` conversion aborted:

```
DETAIL: Key (name)=(Caseworker) already exists.
```

`Role` and `PermissionGroup` are MTI subclasses of `auth.Group`, so all three share the unique `auth_group.name`. `sync_roles()` used `Role.objects.get_or_create(name=...)`, which inserts the shared parent row first — a legacy **bare** `Group(name="Caseworker")` (no `Role`/`PermissionGroup` child, 0 members, 0 permissions) collided there. Because `sync_roles()` runs in one transaction, the rollback skipped every Role, every grant backfill (org-admin/caseworker/shelter) and the phantom retire — zero Grants, reports/teams denied to legacy `PermissionGroup` holders. Prod was remediated by hand (rename + re-run migrate; 152 Grants backfilled).

This is not a one-off row: the shared dev DB (`db:5432`) still holds a bare `Group(pk=1, name="Caseworker")`, so its next `manage.py migrate` fails exactly the same way, and any environment carrying such a leftover hits it too.

## Fix

`accounts/services.py` — `_provision_role()` resolves the row holding a `RoleDef`'s name **before** creating anything:

| name holder | policy |
| --- | --- |
| existing `Role` | reused (unchanged behavior) |
| **memberless** bare `Group` | **adopted** — `accounts_role` child written with the same pk (`Role(group_ptr=group, ...)`): the group becomes the role and keeps its pk/name/references (admin log, guardian perms) instead of being orphaned or duplicated |
| bare `Group` with members | **renamed** to `"<name> (legacy group <pk>)"` (deterministic; `-n` counter if taken) then a fresh `Role` — memberships stay on the renamed group, since a scoped `Role` in `user.groups` is a `permissions.E001` error and adoption is not an option |
| `PermissionGroup` | **never touched** — names are generated (`"<org> [<pk>] · <role>"`), so an exact match is hand-made data; raises an actionable error instead of a raw `IntegrityError` |

Idempotent: once a `Role` holds the name, later runs reuse it and touch nothing. No behavior change for already-converted DBs; the phantom-permission guard, atomic sync and reconcile behavior are untouched.

## Tests — `accounts/tests/test_roles.py::RoleNameCollisionTestCase`

- memberless bare group → adopted in place; Role carries the RoleDef's permissions; exactly one group owns the name
- re-run → idempotent
- adopted Role → caseworker grant backfills (org + member fixture)
- staffed bare group → renamed, membership intact, `permissions.E001` clean
- rename candidate already taken → `-1` variant converges
- `PermissionGroup` name collision → actionable `RuntimeError`, row untouched
- **acceptance**: `call_command("migrate")` (post_migrate) with the legacy bare group present completes, adopts it and backfills the grant

Validation: new suite 7/7; `accounts/tests` + teams/reports grant-authorization suites **369 passed**. Full backend suite running locally / CI.

## Docs

ADR 0001 §2.2 documents the shared `auth.Group` name namespace and the resolution policy; §4 phase 1 notes `sync_roles` is a single transaction (a failure there skips the whole conversion chain) and that name-holders are now resolved.

Fixes the root cause of the 2026-09-11 reports/teams outage (the data-side remediation is already live).

## Summary by Sourcery

Make Role provisioning safely resolve legacy auth.Group name collisions so migrations and authorization backfills complete successfully.

Bug Fixes:
- Make role provisioning resilient to collisions with legacy bare auth groups, preventing migration failures and allowing subsequent grant backfills to complete.

Enhancements:
- Adopt memberless legacy groups in place, rename staffed legacy groups while preserving memberships, and report actionable errors for PermissionGroup collisions.
- Document the shared auth.Group name namespace and deterministic collision-resolution policy.

Documentation:
- Document the role name collision policy and its transactional provisioning implications in ADR 0001.

Tests:
- Add regression and migration acceptance coverage for adoption, idempotency, staffed-group renaming, conflicting rename candidates, PermissionGroup collisions, and grant backfills.

### DIFF
--- a/apps/betterangels-backend/accounts/services.py
+++ b/apps/betterangels-backend/accounts/services.py
@@ -12,6 +12,7 @@ from typing import TYPE_CHECKING, Any
 
 from common.org_types import REGISTRY
 from common.permissions.config import RoleDef, TemplateConfig
+from django.contrib.auth.models import Group
 from django.core.exceptions import ValidationError
 from django.db import IntegrityError, transaction
 from organizations.backends import invitation_backend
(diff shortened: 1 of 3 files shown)


tglaz (approved):
lgtm
robot agrees


Some optional items given by ai, but none are blocking - just FYI for now
in priority order:

- Restore the two-line race tolerance in the no-Group branch — return Role.objects.get_or_create(name=role_def.name) — cheap insurance that preserves the old get_or_create behavior.
- Wording: the permissions.E001 justification for the staffed-group rename only applies to scoped roles (docstring + ADR §2.2); tighten the ADR phase-1 table sentence (bare-Group collisions can no longer abort the chain, a hand-made PermissionGroup match still does, by design).
- Tests: pin the global-RoleDef collision, the 150-char truncation, and rename idempotency; make the "Never touched" assertion in the PermissionGroup test meaningful (or drop it).
- Pre-existing, not this PR: the stale delete_orphaned_group references in reconcile_org_groups's docstring and accounts/admin.py.
- Idea for later: a permissions.E00x check flagging bare groups colliding with RoleDef names, so they surface in manage.py check instead of a deploy log.

PR #2468, opened 2026-09-16
gate ba-privagte fields in fe using perms not flag (SDB-277)

tglaz:
### Private Fields - Additional Contacts (FE)
SDB-277

### In this PR:
- viewing the `Ba Private` link in Shelter Profile is gated by the `Global Shelter Operator` role now instead of a feature flag
  - `superuser` would also see it, so should test with a user that is not `superuser` and toggle the group
  - we primary need to validate that:
    - with `Global Shelter Operator` role: can see `Ba Private` link and create/update/delete additional contacts
    - without `Global Shelter Operator` role: cannot see `Ba Private` link
- updates `additional contacts` form to use 3 columns
  - mostly to create space to make multiple contacts fit the screen better
    - easily reverted in future to 2 columns
- see screenshots below for reference

#### having this group allows viewing `Ba Private` link and adding/editing SO additional contacts
image

### SO view with `Ba Private` link visible and open
image

## Summary by Sourcery

Replace feature-flag access control for BA Private shelter fields with global permission-based gating and improve the additional contacts form layout.

New Features:
- Gate access to the BA Private shelter profile area and additional-contact editing using global ContactInfo permissions.

Bug Fixes:
- Prevent users without BA Private view permissions from accessing the route or seeing its sidebar link.
- Prevent users without contact-change permissions from being offered an edit action that the backend would reject.

Enhancements:
- Reorganize the additional contacts form into a three-column layout.
- Centralize BA Private capability checks in a reusable permissions hook and remove the corresponding feature flag.

Build:
- Adjust Shelter Operator type checking to support the updated project configuration.

Tests:
- Add coverage for BA Private navigation, sidebar visibility, permission mapping, and edit affordances.


shiozakiyuko (commented):
Hi @tglaz 

I just tested this PR **with** and **without** `Global Shelter Operator` role. 
When I don't have the role, it's completely empty even though I am under the Test BA org. I was expecting to see a list of shelters (since I'm under Test BA org), I can get into Basic Info, Details, etc under each shelter WITHOUT BA Private. 
Am I wrong? This is what I see without `Global Shelter Operator` role (screenshot).

Image

shiozakiyuko:
ok, I tested it out and it looks good now.
The account I tested with has these groups. Now I see the list of shelters but no BA Private.

image

shiozakiyuko (approved):

PR #2472, opened 2026-09-22
fix: document download (OUT-225)

mikefeldberg:
closes: OUT-225, [BACS-119](https://betterangels.atlassian.net/browse/BACS-119)

[BACS-119]: https://betterangels.atlassian.net/browse/BACS-119?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by Sourcery

Fix document caching, downloading, and form state so refreshed documents remain accessible without overwriting user edits.

Bug Fixes:
- Fix document downloads so re-downloading to an existing cache destination succeeds.
- Prevent document filename edits and note form values from being overwritten by background data refreshes.
- Refresh cached document URLs in the background to keep previews and downloads usable.

Enhancements:
- Use stable document-specific PDF viewer caching to improve preview behavior.

Tests:
- Add a regression test confirming that a typed document filename survives a background refetch.


libs/expo/betterangels/src/lib/screens/FileScreenComponent/index.tsx
```diff
@@ -51,6 +51,11 @@ export default function FileScreenComponent(props: TFileScreenComponent) {
   const [fileView, setFileView] = useState<TFileView | null>(null);
   const { data } = useQuery(ClientDocumentDocument, {
     variables: { id },
+    // Cached results can carry an expired signed `file.url`, which the
+    // thumbnail and PDF viewer cannot load. Refresh in the background, then
+    // serve cache so the refresh does not repeat on every re-render.
+    fetchPolicy: 'cache-and-network',
+    nextFetchPolicy: 'cache-first',
```
tglaz, line 58: **Possible input reset on the edit screen (introduced by `cache-and-network`)**

`filename` — the value of the edit input — is seeded from the query result by the effect that depends on `[data]`. With the default `cache-first` this effect effectively ran once per screen mount. With `cache-and-network`, every mount now also fires a background refetch, and when it resolves `data` is replaced, so the effect runs a second time and calls `setFilename(...)`, overwriting whatever the user has typed in the meantime.

It's not a rare timing coincidence: `file.url` is re-signed on every request, so the refetched result always differs from the cached one and always produces a new `data` object — the effect is guaranteed to re-run on every visit once the network response lands.

Scenario:

1. Client → Docs → tap a document → "Edit file name" (`/file/[id]?editing=true`).
2. Start typing a new name right away (simulate a slow link, e.g. Network Link Conditioner / throttled connection, to widen the window).
3. When the background refetch resolves, the input reverts to the original filename (any edits so far are lost).

Only affects the editing route — the read-only view just writes the same text back. Probably worth exploring whether the seeding effect should run once instead of on every `data` change.

tglaz (changes requested):

libs/expo/betterangels/src/lib/ui-components/DocumentModal.tsx
```diff
@@ -121,7 +121,9 @@ export default function DocumentModal({
 
     try {
       const cacheDest = new File(new Directory(Paths.cache), originalFilename);
-      const downloaded = await File.downloadFileAsync(url, cacheDest);
+      const downloaded = await File.downloadFileAsync(url, cacheDest, {
+        idempotent: true,
```
tglaz, line 125: possibly add comment about what this solves (not common knowledge)

tglaz (approved):

